PR #9, opened 2017-08-11
Add extra mobile parameters for Prebid Server to handle

nhedley:
This PR adds in device params and edits the user params to be OpenRTB compatible. It also updates the version for Prebid to be 0.0.2, which will be the version for release on Wednesday 8/16.

Unit tests included.

@anwzhang For Android, if we don't already pass strings for user gender, we need to edit the params to do so before release to be Open RTB compatible with the Prebid Server changes.

### DIFF
--- a/sdk/PrebidServerAdapter/PBServerAdapter.m
+++ b/sdk/PrebidServerAdapter/PBServerAdapter.m
@@ -158,7 +158,7 @@ static NSString *const kPrebidMobileVersion = @"0.0.2";
             gender = @"F";
             break;
         default:
-            gender = @"0";
+            gender = @"O";
             break;
     }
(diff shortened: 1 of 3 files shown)


sdk/PrebidServerAdapter/PBServerAdapter.m
```diff
@@ -148,19 +149,19 @@ - (NSDictionary *)user {
...
-            gender = 1;
+            gender = @"M";
             break;
         case PBTargetingParamsGenderFemale:
-            gender = 2;
+            gender = @"F";
             break;
         default:
-            gender = 0;
+            gender = @"0";
```
anwzhang: The spec seems to be character “O” not number "0"

PR #23, opened 2017-09-22
Update targeting

nhedley:
Check for clLocation before setting PBServerLocation

### DIFF
--- a/sdk/PrebidServerAdapter/PBServerAdapter.m
+++ b/sdk/PrebidServerAdapter/PBServerAdapter.m
@@ -13,11 +13,9 @@
  limitations under the License.
  */
 
-#import <arpa/inet.h>
 #import <AdSupport/AdSupport.h>
 #import <CoreTelephony/CTCarrier.h>
 #import <CoreTelephony/CTTelephonyNetworkInfo.h>
-#import <ifaddrs.h>
 #import <sys/utsname.h>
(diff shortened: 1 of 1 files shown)


sdk/PrebidServerAdapter/PBServerAdapter.m
```diff
@@ -257,7 +257,10 @@ - (NSDictionary *)device {
 
 - (NSDictionary *)geo {
     CLLocation *clLocation = [[PBTargetingParams sharedInstance] location];
-    PBServerLocation *location = [PBServerLocation getLocationWithLatitude:clLocation.coordinate.latitude longitude:clLocation.coordinate.longitude timestamp:clLocation.timestamp horizontalAccuracy:clLocation.horizontalAccuracy];
+    PBServerLocation *location;
+    if (clLocation) {
+        location = [PBServerLocation getLocationWithLatitude:clLocation.coordinate.latitude longitude:clLocation.coordinate.longitude timestamp:clLocation.timestamp horizontalAccuracy:clLocation.horizontalAccuracy];
```
anwzhang: What happens if clLocation does not exist, just don't initialize location?

nhedley: Yeah if the clLocation isn't set we don't send anything for location to Prebid Server.

PR #26, opened 2017-09-25
Bid expiry update alignment with Android

nhedley:
New expected behavior:
- When bid responses are saved, the expiry time on those bid responses is set to 4 minutes 30 seconds.
- Every 30 seconds poll will check all ad units for expired bids. 
- If there was a bid for that ad unit and that bid is expired, start a new auction in Prebid Server.
- On iOS, on MoPub or DFP autorefresh, send any cached bid to MoPub / DFP line items and start a new auction.
- On iOS, when setBidKeywords API is called set adUnit identifier to start a new auction on swizzled load ad method.
- On both platforms, a new auction is run every time the setBidKeywords API is called.

### DIFF
--- a/sdk/PrebidMobile/PBBidManager.m
+++ b/sdk/PrebidMobile/PBBidManager.m
@@ -114,7 +114,7 @@ static dispatch_once_t onceToken;
 }
 
 - (nullable NSDictionary<NSString *, NSString *> *)keywordsForWinningBidForAdUnit:(nonnull PBAdUnit *)adUnit {
-    NSArray *bids = [self getWinningBids:adUnit];
+    NSArray *bids = [self getBids:adUnit];
     [self startNewAuction:adUnit];
     if (bids) {
         PBLogDebug(@"Bid is available to create keywords");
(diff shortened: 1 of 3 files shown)


sdk/PrebidMobile/PBBidManager.m
```diff
@@ -236,27 +232,26 @@ - (void)checkForBidsExpired {
...
     }
-    return topBidResponse;
+    return nil;
 }
 
 - (BOOL)isBidReady:(NSString *)identifier {
-    if ([[_bidsMap allKeys] containsObject:identifier] && [_bidsMap objectForKey:identifier] != nil) {
+    if ([[_bidsMap allKeys] containsObject:identifier] &&
+        [_bidsMap objectForKey:identifier] != nil &&
+        [[_bidsMap objectForKey:identifier] count] > 0) {
```
anwzhang: This is getting one object from _bidsMap, why checking the count here?

nhedley: It's just an extra check to make sure the _bidsMap isn't empty. The same check is done on Android in this function, I wanted to make them consistent.

PR #29, opened 2017-09-26
Clean up changes for 0.1.0 release

nhedley:
Went through to make sure that code looked good. Found a couple of the changes below
- Fix removing duplicate ad units on registration to check against adunit identifier instead of adunit. add tests
- remove contains check on isBidReady, its redundant to the next check and takes O(n) time and risks iterating through unnecessarily
- Fix bids expired test

### DIFF
--- a/sdk/PrebidMobileTests/PrebidMobileCoreTests/PBBidManagerTests.m
+++ b/sdk/PrebidMobileTests/PrebidMobileCoreTests/PBBidManagerTests.m
@@ -110,6 +110,20 @@ NSString *const kBidManagerTestAdUnitId = @"TestAdUnitId";
     XCTAssertNil(returnedAdUnit);
 }
 
+- (void)testRegisterAdUnitWithSameIdentifier {
+    PBAdUnit *returnedUnit = nil;
+    PBAdUnit *bannerAdUnit1 = [[PBBannerAdUnit alloc] initWithAdUnitIdentifier:@"sameid" andConfigId:@"0b33e7ae-cf61-4003-8404-0711eea6e673"];
+    [bannerAdUnit1 addSize:CGSizeMake(320, 50)];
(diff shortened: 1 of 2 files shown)


sdk/PrebidMobileTests/PrebidMobileCoreTests/PBBidManagerTests.m
```diff
@@ -110,6 +111,20 @@ - (void)testRegisterBannerAdUnitNoSizeException {
...
+    PBAdUnit *returnedUnit = nil;
+    PBAdUnit *bannerAdUnit1 = [[PBBannerAdUnit alloc] initWithAdUnitIdentifier:@"sameid" andConfigId:@"0b33e7ae-cf61-4003-8404-0711eea6e673"];
+    [bannerAdUnit1 addSize:CGSizeMake(320, 50)];
+    PBAdUnit *bannerAdUnit2 = [[PBBannerAdUnit alloc] initWithAdUnitIdentifier:@"sameid" andConfigId:@"0b33e7ae-cf61-4003-8404-0711eea6e673"];
+    [bannerAdUnit2 addSize:CGSizeMake(320, 50)];
+
+    [[PBBidManager sharedInstance] registerAdUnits:@[bannerAdUnit1, bannerAdUnit2] withAccountId:self.accountId];
+    returnedUnit = [[PBBidManager sharedInstance] adUnitByIdentifier:[bannerAdUnit1 identifier]];
+
+    XCTAssertNotNil(returnedUnit);
```
anwzhang: Do you need to test that only one adunit with the identifier exist in the bid manager?

nhedley: Updated the test to check that the second ad unit is the one that is returned, the first ad unit gets removed.

PR #32, opened 2017-09-26
Batch requests to prebid server on bid expiry

nhedley:
This should reduce the amount of requests to Prebid Server if people are creating a lot of ad units.


sdk/PrebidMobile/PBBidManager.m
```diff
@@ -116,6 +116,7 @@ - (void)assertAdUnitRegistered:(NSString *)identifier {
 - (nullable NSDictionary<NSString *, NSString *> *)keywordsForWinningBidForAdUnit:(nonnull PBAdUnit *)adUnit {
     NSArray *bids = [self getBids:adUnit];
     [self startNewAuction:adUnit];
```
anwzhang: Remove startNewAuction?

sdk/PrebidMobile/PBBidManager.m
```diff
@@ -116,6 +116,7 @@ - (void)assertAdUnitRegistered:(NSString *)identifier {
...
+    [self requestBidsForAdUnits:@[adUnit]];
```
anwzhang: Is this the only place that startNewAuction being used?

nhedley: It's being used 2 times--here and in checkForBidsExpired.

anwzhang: It's probably clearer to change the name too, to something like resetAdUnit to suit for the actual function.

anwzhang:
@ppuviarasu this can be merged.

PR #41, opened 2017-10-02
Check for adUnit to resolve crash in 0.1.0

nhedley:
This fixes a crash where when the app was closed, there is an NSInvalidArgumentException in startNewAuction.


sdk/PrebidMobile/PBBidManager.m
```diff
@@ -206,9 +206,11 @@ - (void)requestBidsForAdUnits:(NSArray<PBAdUnit *> *)adUnits {
 }
 
 - (void)startNewAuction:(PBAdUnit *)adUnit {
-    [adUnit generateUUID];
-    [_bidsMap removeObjectForKey:adUnit.identifier];
-    [self requestBidsForAdUnits:@[adUnit]];
+    if (adUnit && adUnit.identifier) {
+        [adUnit generateUUID];
+        [_bidsMap removeObjectForKey:adUnit.identifier];
```
anwzhang: Help me understand the problem correctly: why would startNewAuction be called if deallocation happens ? and what does the deallocation do in details?

nhedley: Here is what I think is happening-- I can only reproduce sometimes when I close out of our demo app and it seems like requestParameters on DFP is getting called which calls startNewAuction. At that time the adUnit is nil so I'm assuming it got deallocated. Therefore adding the check here and in the swizzling will fix.

anwzhang: Looking at the crash log, the crash seems to be happening on removeObjectForKey method. If AdUnit was deallocated, wouldn't the exceptions happens when adUnit generateUUID is called?

nhedley: If the adUnit is nil, it doesn't actually throw the exception until removeObjectForKey. In objective C its valid to call a method on a nil object, its just treated as a noop (https://stackoverflow.com/questions/2696891/calling-a-method-on-an-uninitialized-object-null-pointer).

However, if you call removeObjectForKey on a nil object, it throws the `NSInvalidArgumentException` we're seeing in the crash. Its possible that something to remove the adUnit from DFP is happening in this call in the crash log `GADCategories_GADSDKCoreController_ActiveViewUnloadCache` and then that calls the requestParameters swizzled method when the adUnit no longer exists.

anwzhang: I see. Should it be happening that ad unit being removed? I don't remember we provide anything in the SDK that does that. Is it supposed to be deallocated when app goes into background? If it is, is there a way that we can have the prebid working when app comes back from background, since this solution simply stops making requests.

nhedley: There's nothing in our SDK that would cause the adUnit to be removed. But we can't control however DFP is handling their adViews on app closed. This isn't happening when the app is in the background, its happening when the app is actually closed out. Prebid works with the app in the background.

PR #48, opened 2017-11-20
Use local storage to cache ads instead of Prebid Cache

nhedley:
For pubs with DFP as primary ad server use [EGOCache library](https://github.com/enormego/EGOCache) to safely cache bids with a 6 minute expiry. This code caches the full bid object and generates cacheId client side.

For pubs with MoPub, still cache the bid on server side, send cache_markup = 1 through to PBS.

Deprecate old registerAdUnits API in favor of the new one where we can cache locally for DFP bids.

### DIFF
--- a/sdk/PrebidMobile/PrebidURLProtocol.m
+++ b/sdk/PrebidMobile/PrebidURLProtocol.m
@@ -1,0 +1,26 @@
+/*   Copyright 2017 APPNEXUS INC
+ 
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+ 
+ http://www.apache.org/licenses/LICENSE-2.0
+ 
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+#import "PrebidURLProtocol.h"
+
+@interface PrebidURLProtocol () <NSURLConnectionDelegate>
+
+@property (nonatomic, strong) NSURLConnection *connection;
+
+@end
+
+@implementation PrebidURLProtocol
+
+@end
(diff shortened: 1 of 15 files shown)


sdk/PrebidMobile/PrebidURLProtocol.m
```diff
@@ -0,0 +1,92 @@
...
+
+- (void)startLoading {
+    NSMutableURLRequest *newRequest = [self.request mutableCopy];
+    [NSURLProtocol setProperty:@YES forKey:@"PrebidURLProtocolHandledKey" inRequest:newRequest];
+
+    NSRange range = [newRequest.URL.absoluteString rangeOfString:@"uuid="];
+    if (range.location == NSNotFound) {
+        NSLog(@"No uuid in URL");
+    }
+    NSString *uuid = [newRequest.URL.absoluteString substringFromIndex:(range.location + 5)];
```
anwzhang: we can't assume that uuid is by the end of the URL.

sdk/PrebidMobile/PrebidURLProtocol.m
```diff
@@ -0,0 +1,92 @@
...
+
+    NSDictionary *dict = (NSDictionary *)[[EGOCache globalCache] objectForKey:uuid];
+    if (dict != nil) {
+        NSData *data = [NSJSONSerialization dataWithJSONObject:dict
+                                                       options:NSUTF8StringEncoding
+                                                         error:nil];
+        NSURLResponse* response = [[NSHTTPURLResponse alloc] initWithURL:[self.request URL] statusCode:200 HTTPVersion:@"HTTP/1.1" headerFields:@{@"Access-Control-Allow-Origin": @"https://pubads.g.doubleclick.net", @"Access-Control-Allow-Credentials" : @"true"}];
```
anwzhang: Should MoPub origin be added here too?

sdk/PrebidMobile/PrebidURLProtocol.m
nhedley: Why not? We're in complete control of the cache call from pbm.js and when we call from native code for FB so we know what the url looks like exactly.

sdk/PrebidMobile/PrebidURLProtocol.m
nhedley: Yup will add. Need to test MoPub still.

nhedley: Added MoPub support with Prebid Cache since we can't use local cache solution as MoPub uses WKWebView.

PR #52, opened 2017-12-15
Closes #51: Add option to set prebid server host url

### LINKED ISSUE #51: Add prebid server host option in SDK
Instead of defaulting to AppNexus' hosted version of prebid server, we want to give the devs an option of using another prebid server host, such as Rubicon's.

sandraleon:
- Devs must now explicitly set the prebid server host they wish to use. If one is not set they will get an error. It does not default to any particular prebid server host.
- Also updated the header files with the new copyright message.


nhedley:
@sandraleon Do you mind updating the copyright message in all files in a separate PR? Will make this one much easier to review.

sandraleon:
@nhedley absolutely. done.

nhedley:
As discussed in the phone call yesterday, if you could update this PR to use a config management system for vendor, rather than a url that would be preferred. 

Ideally, we have a static mapping somewhere of a vendor name (ie Rubicon, AppNexus) to the associated vendor urls (ie prebid server url, prebid cache url) and any additional configuration information that is relevant to that vendor.

Just to note on this PR, by changing the prebid server url and prebid cache url, we need to make sure the prebid cache url is passed through from Prebid Server to the creative where pbm.js is loaded in order to load the creative from the correct cache url. 

Let me know if you have any questions or need any clarifications!

djsu:
@nhedley Just to clarify this section:

> we need to make sure the prebid cache url is passed through from Prebid Server to the creative where pbm.js is loaded in order to load the creative from the correct cache url.

The current implementation uses a macro in the creative that only replaces the cache ID. Depending on what hosted Prebid Server is being used, Publishers still need to manually change the URL where pbm.js is hosted within the creative code.

Is the suggestion to change the creative setup to use one single holistic creative code? Then, depending on what URL is set in the SDK, the SDK would pass the appropriate Prebid Cache URL in the ad request to DFP/MoPub and ultimately to the creative?

@randeepsiddhu @sandraleon

nhedley:
This was discussed last week on the call, but just wanted to comment here. I suggest that we have the publisher define a vendor that they would like to work with (ie, AppNexus, Rubicon) and have that vendor statically map to Prebid Cache URL, Prebid Server URL, and any other necessary info.

We can update pbm.js to use the correct URL per vendor, so publishers should be able to use similar creative code.

anwzhang:
@ppuviarasu resolve merge conflict on this one and merge.

PR #55, opened 2018-02-05
batch send server requests & fix unit test cases

ppuviarasu:
Added logic to batch the adunits to be sent only 10 per request
fixed the broken unit test cases

### DIFF
--- a/sdk/PrebidMobileTests/Stubbing/ANHTTPStubURLProtocol.h
+++ b/sdk/PrebidMobileTests/Stubbing/ANHTTPStubURLProtocol.h
@@ -1,8 +1,4 @@
-<<<<<<< HEAD
-/*   Copyright 2014 Prebid.org, Inc.
-=======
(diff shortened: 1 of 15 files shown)


sdk/PrebidMobileTests/Stubbing/ANHTTPStubURLProtocol.h
```diff
@@ -1,8 +1,4 @@
-<<<<<<< HEAD
-/*   Copyright 2014 Prebid.org, Inc.
-=======
 /*   Copyright 2017 APPNEXUS INC
```
anwzhang: Use Prebid.org, Inc instead.

anwzhang: And we should probably change all files to use Prebid.org, need to confirm with Matt.

PR #64, opened 2018-04-16
Set gdpr settings open rtb

ppuviarasu:
{
  "device": {
    "w": 414,
    "h": 736,
    "ifa": "8A84CD22-CDCE-491A-8A84-8000BED7B08D",
    "devtime": 1523896447,
    "osv": "11.2",
    "lmt": 0,
    "connectiontype": 1,
    "os": "iOS",
    "pxratio": 3,
    "make": "Apple",
    "ua": "Mozilla/5.0 (iPhone; CPU iPhone OS 11_2 like Mac OS X) AppleWebKit/604.4.7 (KHTML, like Gecko) Mobile/15C107",
    "model": "x86_64"
  },
  "source": {
    "tid": "123"
  },
  "test": 1,
  "app": {
    "bundle": "com.AppNexus.PrebidMobileValidator",
    "ext": {
      "prebid": {
        "version": "0.2.1",
        "source": "prebid-mobile"
      }
    },
    "publisher": {
      "id": "aecd6ef7-b992-4e99-9bb8-65e2d984e1dd"
    },
    "ver": "1.0"
  },
  "id": "B0A63760-125D-47CB-9FDE-B5910B3E02BF",
  "imp": [
    {
      "id": "HomeScreen",
      "ext": {
        "prebid": {
          "storedrequest": {
            "id": "eebc307d-7f76-45d6-a7a7-68985169b138"
          }
        }
      },
      "secure": 1,
      "banner": {
        "format": [
          {
            "w": 300,
            "h": 250
          }
        ]
      }
    }
  ],
  "ext": {
    "prebid": {
      "targeting": {
        "pricegranularity": "medium",
        "lengthmax": 20
      }
    }
  },
  "regs": {
    "ext": {
      "gdpr": 1
    }
  },
  "user": {
    "yob": 1993,
    "ext": {
      "consent": ""
    },
    "gender": "F"
  }
}


anwzhang:
You added the test app in the same review. should be removed.

sdk/PrebidServerAdapter/PrebidServerModule/PBServerRequestBuilder.m
```diff
@@ -260,6 +261,17 @@ - (NSDictionary *)openrtbGeo {
     }
 }
 
+-(NSDictionary *) openrtbRegs {
+    
+    NSMutableDictionary *regsDict = [[NSMutableDictionary alloc] init];
+    
+    BOOL gdpr = [[PBTargetingParams sharedInstance] gdpr];
+    
+    regsDict[@"ext"] = @{@"gdpr" : @(@(gdpr).integerValue)};
```
anwzhang: This should be do not send GDPR related info if Publisher does not set on the API.

ppuviarasu: By default the consent is treated as Yes if the user has not provided the value so technically we can to pass in YES here

sdk/PrebidMobile/PBTargetingParams.m
```diff
@@ -158,4 +159,29 @@ - (void)setItunesID:(NSString *)itunesID {
...
+    [[NSUserDefaults standardUserDefaults] setBool:subjectToGDPR forKey:PB_GDPR_SubjectToConsent];
+}
+
+-(void) setGdprConsentString:(NSString *)gdprConsentString{
+    
+    [[NSUserDefaults standardUserDefaults] setObject:gdprConsentString forKey:PB_GDPR_ConsentString];
+}
+
+-(BOOL) subjectToGDPR {
+    BOOL savedGDPR = YES;
```
anwzhang: should be able to return nil case, which indicates publisher didn't set any GDPR.

sdk/PrebidServerAdapter/PrebidServerModule/PBServerRequestBuilder.m
```diff
@@ -260,6 +261,17 @@ - (NSDictionary *)openrtbGeo {
     }
 }
 
+-(NSDictionary *) openrtbRegs {
+    
+    NSMutableDictionary *regsDict = [[NSMutableDictionary alloc] init];
```
anwzhang: If GDPR is not set, SDK should not be sending info about GDPR.

sdk/PrebidMobile/PBTargetingParams.m
ppuviarasu: If nil & if no values are passed in the request, the impbus treats as subjectToGDPR as `YES`... so its ok to pass in YES by default...

anwzhang: From OPENRTB Spec
The “Regs” object will signal whether or not the request is subject to GDPR regulations. It will do so via
the extension attribute “gdpr” which is an optional integer that indicates: 0 = No, 1 = Yes. Under
OpenRTB conventions for optional attributes, omission indicates Unknown.

0, 1 and unknown actually means different things to exchanges and buyers. For example, impbus will use a different way to process unknown situation.

sdk/PrebidMobile/PBTargetingParams.m
```diff
@@ -158,4 +162,38 @@ - (void)setItunesID:(NSString *)itunesID {
...
+        savedGDPR = [[NSUserDefaults standardUserDefaults] boolForKey:PB_GDPR_SubjectToConsent];
+    } else if([[NSUserDefaults standardUserDefaults] objectForKey:IAB_GDPR_SubjectToConsent] != nil){
+        self.isGDPREnabled = YES;
+        NSString *stringValue = [[NSUserDefaults standardUserDefaults] objectForKey:IAB_GDPR_SubjectToConsent];
+        savedGDPR = [stringValue boolValue];
+    }
+    return savedGDPR;
+}
+
+-(BOOL) isSubjectToGDPR {
```
anwzhang: Change the name to isGDPREnabled for better reading.

PR #84, opened 2018-07-25
PriceGranularity

AntoineJac:
Add PriceGranularity option instead of always passing Medium

### DIFF
--- a/examples/PrebidMobileDemo/PrebidMobileDemo/AppDelegate.m
+++ b/examples/PrebidMobileDemo/PrebidMobileDemo/AppDelegate.m
@@ -66,7 +66,7 @@
 
         [self setPrebidTargetingParams];
 
-        [PrebidMobile registerAdUnits:@[adUnit1, adUnit2] withAccountId:kAccountId withHost:kPBServerHost andPrimaryAdServer:PBPrimaryAdServerDFP];
+        [PrebidMobile registerAdUnits:@[adUnit1, adUnit2] withAccountId:kAccountId withHost:kPBServerHost withPriceGranularity:kPPriceGranularity andPrimaryAdServer:PBPrimaryAdServerDFP];
     } @catch (PBException *ex) {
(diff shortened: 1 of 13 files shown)


AntoineJac:
@anwzhang This is not the documented and correct behaviour for Prebid server:
 https://github.com/prebid/prebid-server/blob/master/docs/endpoints/openrtb2/auction.md

"request.ext.prebid.targeting is an optional property which causes Prebid Server to set these params on the response at response.seatbid[i].bid[j].ext.prebid.targeting."

So passing this information should override the server side properties! 

I can quickly edit my code to make sure the changes are only concerning Rubicon.
@bretg what are your thoughts?

anwzhang:
Hi @AntoineJac 
I strongly suggest you to implement this on the server side. Even if you want to override it on the client side, don't change it on the registration method, add and API to support it.

AntoineJac:
Hi @anwzhang , 
It is implemented on server side but we need to let publisher decided if they want to hardcode it.
I would like to avoid proliferating special-purpose functions that is why I first used the init function. 

But you are correct and this not the best way of doing it. I would prefer letting the init function simple.
I am going to edit my PR to create new class PBConfig that we could could use as a generic config-setting class that can handle granularity, request-level-config-id, and whatever else we throw at it in the future.

Let me know your thoughts about it?

anwzhang:
Hi @AntoineJac that sounds good to me.

AntoineJac:
@anwzhang, I agree that this should be a server side control, but I think we should still provide the option to publishers to override it if they want to be sure no one can change it by mistake on the server.

anwzhang:
I'm afraid I have to close this PR for the following reasons:
1. No confirmed publisher use case
2. Prefer not giving the opportunity to hardcode wrong values 

For the existing hardcode, I opened an issue to track: https://github.com/prebid/prebid-mobile-ios/issues/88
You're welcome to open a new PR to just fix that.

PR #86, opened 2018-07-30
Fix prebid cache

anwzhang:
PrebidCache doesn't write to disk until it's released in memory, thus causing DFP/MoPub webview unable to read from PrebidCache sometimes. Fix is only sending the cache id back after PrebidCache writes to disk, a.k.a ARC is zero for webviews caching the content.


ppuviarasu:
We have switched to server side cache & not using local cache anymore... Will revisit this later

Issue #88, opened 2018-08-03
PrebidMobile should not hardcode PriceGranularity on the client side


anwzhang:

ptomasroos:
Is this resolved now since #90 @anwzhang ?

anwzhang:
@ptomasroos yes it's resolved.

PR #90, opened 2018-08-07
fix hardcoding medium as price granularity 

anwzhang:
Issue: https://github.com/prebid/prebid-mobile-ios/issues/88


AntoineJac:
Hi @anwzhang , 

Here are my remarks on the PR:
- request.ext.storedrequest.id should not be the accountId by default. It should be defined by the publisher.
- request.ext.prebid.targeting must be present event is the value is void {}. Or Prebid Server won't set the KV on the response at response.seatbid[i].bid[j].ext.prebid.targeting

anwzhang:
Hi @AntoineJac 
I talked with PBS engineer.
You're right on the second point, it's better to add `request.ext.prebid.targeting`, I'll update the PR.
For the first point, account id is defined by the publisher, it's the id that PBS needs to pull server side settings for the SDK, I don't know what else value could it be other than account id?

PR #93, opened 2018-08-10
PBM should not save the bid if targeting keys are not returned from server

anwzhang:


sdk/PrebidServerAdapter/PBServerAdapter.m
```diff
@@ -105,17 +105,19 @@ - (void)requestBidsWithAdUnits:(nullable NSArray<PBAdUnit *> *)adUnits
...
                         }
-                        NSString *bidderCacheId = cacheIds[i];
-                        NSString *cacheIdkey =[ NSString stringWithFormat:@"%@_%@", kAPNAdServerCacheIdKey, bidsArray[i][@"seat"]];
-                        cacheIdkey = cacheIdkey.length > 20 ? [cacheIdkey substringToIndex:20] : cacheIdkey;
-                        adServerTargetingCopy[cacheIdkey] = bidderCacheId;
-                        PBBidResponse *bidResponse = [PBBidResponse bidResponseWithAdUnitId:adUnitId adServerTargeting:adServerTargetingCopy];
-                        PBLogDebug(@"Bid Successful with rounded bid targeting keys are %@ for adUnit id is %@", [bidResponse.customKeywords description], adUnitId);
-                        [bidResponsesArray addObject:bidResponse];
                     }
                     [delegate didReceiveSuccessResponse:bidResponsesArray];;
```
ppuviarasu: Why are u sending `didReceiveSuccessResponse ` when targeting keys are nil?

anwzhang: check out the full code. I'm not sending `didReceiveSuccessResponse` when targeting keys are nil, that's inside bid response retrieval.

PR #94, opened 2018-08-20
Use server cache

anwzhang:


sdk/PrebidServerAdapter/PrebidServerModule/PBServerRequestBuilder.m
```diff
@@ -87,7 +87,7 @@ - (NSDictionary *)openrtbRequestExtension: (NSString *)accountId
     NSMutableDictionary *requestPrebidExt = [[NSMutableDictionary alloc] init];
     requestPrebidExt[@"targeting"] = @{};
     requestPrebidExt[@"storedrequest"] = @{@"id" :accountId};
```
Kowshickkarthick: Looks like this enables server cache for both DFP and Mopub is it the intended behavior??

PR #97, opened 2018-09-11
Update V0.5 to select cache server or local

AntoineJac:
@anwzhang , I have used your PR and add a small parameters so Publisher can use server or local caching. Both have perks and drawback. 
Local caching seem to work fine with your PR, I have tested and see no more blank ad (when webview has not cached yet the creative before it is request by the dfp creative).

### DIFF
--- a/sdk/PrebidMobile/PBTargetingParams.h
+++ b/sdk/PrebidMobile/PBTargetingParams.h
@@ -74,6 +74,17 @@ typedef NS_ENUM(NSUInteger, PBTargetingParamsGender) {
 @property (nonatomic, readonly, assign) BOOL isGDPREnabled;
 
 /**
+ * The boolean value set by the user to collect user data
+ */
+@property (nonatomic, readwrite) BOOL localCache;
+
+/**
+ * The boolean value set by the user to define the cache type
+ */
+
+@property (nonatomic, readonly, assign) BOOL useLocalCache;
+
+/**
  * This property stores the set of custom keywords that prebid provides for targeting
  */
 @property (nonatomic, strong, readonly) NSDictionary<NSString *, NSArray *> *__nullable customKeywords DEPRECATED_ATTRIBUTE;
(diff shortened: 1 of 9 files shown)


AntoineJac:
@anwzhang , do you please have any updates on this?
Thanks,

ppuviarasu:
@AntoineJac we have taken this PR for this sprint.

Since this involves adding a new API we will review this with our product manager.

The code looks good. Its the product approval that we are looking into.

We will update you on the same in a couple of days

thanks

AntoineJac:
Thanks @ppuviarasu and @anwzhang 

Before merging this PR I have a questions, why are we using the UIWebview for DFP caching and WKWebView for MoPub. Should we not used WKWebView for both as UIWebview has been deprected for iOS12 and is provoking crash on React Native...

I have made some test and seem WKWebView is working also correctly for DFP. If no reasons I will update this PR to only use WKWebView (also in the class PBServerGlobal where UIWebview is used to get the UA).

Thanks,

ppuviarasu:
@AntoineJac 
Deprecating UIWebView is a new PR. not to be combined into this one.

For this existing PR for allowing the publisher to choose the caching option, Can you raise an issue & assign it to the product managers (@mjacobsonny for AppNexus). 

The issue has to be flagged as a feature by the product & then we can attach this PR to the issue & start looking into the same.

We cannot merge this PR until we have product signoff on the issue

AntoineJac:
@ppuviarasu , thanks will do. I will open a new PR as I have worked locally with a new code to manage the caching in a more stable way and only base on WKWebView.

Indeed even with the code from this PR, many time (over 10% on my tests!!!!), the line item is trying to display the creative before it has finished caching it. The issue is knowing when "localStorage.setItem" has finished its job. I have this used a completionHandler on it. I don't see any errors with my new code. Will push the new PR soon, I am stuck with other staff to finish before. 

Thanks,

AntoineJac:
Here is the new PR with a new way of caching:
https://github.com/prebid/prebid-mobile-ios/pull/102

When tested I don't see any crash and no KV are passed if the creative has not finishing caching which avoid calling Prebid line item and provoking a blank ad (issue with current caching class).

Issue #77, opened 2018-06-05
0.3 does not build on Xcode 10 beta (10L176w)


shujew:
hey @ppuviarasu 

Can you please take a look at the error that I'm receiving while compiling 0.3 with Xcode 10 beta (10L176w)? Seems to be an issue with the new Xcode 10 build system.

```
error: Multiple commands produce '/.../Debug-iphoneos/PrebidMobile/PrebidMobile.framework/Headers/PrebidMobile-umbrella.h':
1) Target 'PrebidMobile' has copy command from '/.../Pods/PrebidMobile/sdk/PrebidMobile/PrebidMobile-umbrella.h' to '/.../Developer/Xcode/DerivedData/xxx-biwlmbnzlziimjdmmfuhrsqpkfic/Build/Products/Debug-iphoneos/PrebidMobile/PrebidMobile.framework/Headers/PrebidMobile-umbrella.h'
2) Target 'PrebidMobile' has copy command from '/.../Pods/Target Support Files/PrebidMobile/PrebidMobile-umbrella.h' to '/.../Developer/Xcode/DerivedData/xxx-biwlmbnzlziimjdmmfuhrsqpkfic/Build/Products/Debug-iphoneos/PrebidMobile/PrebidMobile.framework/Headers/PrebidMobile-umbrella.h'
```

Thanks

ppuviarasu:
Hey @shujew 

I have tested the Prebidmobile SDK 5.0 on xcode 10.0 & the sdk builds fine.

Can you confirm the same.

Thanks

ptomasroos:
Works great for us who has been involved since 0.0.1

shujew:
@ppuviarasu I can confirm that this work as a temporary workaround. Hopefully, this issue can be addressed in the next release

ppuviarasu:
Fixed in v 0.5.1

ppuviarasu:
Are you integrating on the swift apps?

I tried this on the swift apps & Objective C apps & works fine...

wacpenales:
Still reproducible after 0.5.1 
Even after deleting Pod cache and all.

Using cocoapod version 1.5.3.
Should we update the exclude statement to 
`s.exclude_files = "sdk/PrebidMobile/PrebidMobile-umbrella.h"` ?

PR #104, opened 2018-10-08
changes for xcode 10 compatibility

ppuviarasu:


ppuviarasu:
associated to the issue https://github.com/prebid/prebid-mobile-ios/issues/77

PR #108, opened 2018-10-17
Fix bug React Native

AntoineJac:
PBServerGlobal.m:26 at UIWebView init. It is being relied on having the main thread active which is not the case with react native. I wrapped the user agent retrieval using GCD. And since it becomes async, I also call it for the first time once the PBServerGlobal class is loaded in ObjC runtime.


AntoineJac:
@anwzhang , can you please review this one. It is important as make crash react native. Thanks

PR #114, opened 2018-11-02
it appears GADSlot is renamed to GADOSlot

m44rt3n:
In the new sdk of 7.35.1 the swizzle fails because it cannot find GADSlot class, which now seems to be called GADOSlot.
In addition the keywordsForWinningBidForAdUnit also invokes a call to the prebid server, while bids are already available.
this commit should solve these issues


ppuviarasu:
Thanks for the pull request. @m44rt3n I have few concerns with the pull request

1. We have a new release 0.5.1 nothing major but a fix for the breaking crash on reactNative apps
2. Please do not include pod files in the PR
3. Also what if someone wants to use older versions of the DFP sdk? will your fix work

ppuviarasu:
we have fixed in the our new version #0.5.3

PR #118, opened 2018-11-28
Fix new DFP class GADOSlot + itunesID passed as bundle if defined + new version

AntoineJac:


AntoineJac:
@ppuviarasu , can you please have a look at this one, I have some clients blocked as it is not compliant with the last DFP SDK pod version. This PR have:
- Fix for this new class
- Also I have added the option to use itUnesId as bundle if defined in PBTargetingParmas class.
- New version 5.2.3

Can you please check and release if all ok?

Thanks,

PR #124, opened 2018-12-03
fix for zoomed tests run

ppuviarasu:
1. Fix for App camera usage
2. Fix for zoomed test run


tools/PrebidValidator/Dr.Prebid/Settings/SettingsTableViewController.m
```diff
@@ -173,6 +173,69 @@ -(void)tableView:(UITableView *)tableView willDisplayCell:(UITableViewCell *)cel
         [cell setLayoutMargins:UIEdgeInsetsZero];
     }
     
+    if(indexPath.section == 0 && indexPath.row == 0 && [cell isKindOfClass:[SegmentCell class]]){
```
anwzhang: Why moving these from didPressNext to here?

ppuviarasu: Becuase the scroll of page prevents the 1st cell values to be persisted. So basically the values present in screen only is stored & not all the scrolled up values

PR #126, opened 2018-12-04
Dr.Prebid: fix for rubicon creative not rendering

anwzhang:


ppuviarasu:
I am not seeing the creative... also no response content

PR #143, opened 2019-01-18
Fix DFP swizzling

AntoineJac:
DFP has changed their GadSlot and it would be better to stop relying on swizzling technique on 3rd party SDK (Google SDK's or Mopub). It is a dangerous practise behaviour in case on of the class or method we are swizzling is edited (that happened with GadSlot fixed in v0.5.3). Some publisher may also use a SDK rewriting the SDK api already.

The new method is adding the key value to the customTargeting array and will be easy to implement.
 
Objective-c:
.....
        self.bannerView.adUnitID = @"/7191/header-bidding";
        self.bannerView.rootViewController = self;
        DFPRequest *request = [DFPRequest request];
        [PrebidMobile setBidKeywordsOnDFPBannerView:_dfpAdView withAdUnitId:kAdUnit1Id withTimeout:600 withRequest:request];

Swift:
        banner.adUnitID = "/7191/header-bidding"
        banner.rootViewController = self
        let request = DFPRequest()
        PrebidMobile.setBidKeywordsOnDFPBannerView(self.banner, withAdUnitId: "TestUnit1", withTimeout: 600, withRequest: request)

### DIFF
--- a/sdk/PrebidMobile/Categories/NSObject+Prebid.m
+++ b/sdk/PrebidMobile/Categories/NSObject+Prebid.m
@@ -23,6 +23,11 @@
 @implementation NSObject (Prebid)
 
 @dynamic pb_identifier;
+static bool swizzlingDisable = false;
+
++ (void)disableSwizzling {
+    swizzlingDisable = true;
+}
 
 + (void)load {
     static dispatch_once_t loadToken;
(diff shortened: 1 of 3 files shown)


AntoineJac:
@ppuviarasu , for information here is a way to not removed on swizzling method.
Do you have any date or update about prebid 1.0? Thanks

ppuviarasu:
@AntoineJac the 1.0 doesnt do the swizzling and the refresh is controlled by PrebidMobile is more cleaner approach than the existing one.
We have also implemented the API in swift & the solution is same as what we have in this PR. So unless its a blocker I would suggest to wait for the 1.0. we are in the test cases writing stage so should be available soon....

AntoineJac:
@ppuviarasu , thanks for the update. No it is fine and will wait for the new 1.0 version :)
Thanks again

PR #146, opened 2019-02-18
fixing changes in Dr Prebid app

akashVermaCardinal:
Fixing changes in Dr Prebid app as per the new Swift Prebid SDK.


src/PrebidMobile/PrebidMobile/Dispatcher.swift
```diff
@@ -44,6 +44,8 @@ class Dispatcher:NSObject {
         
         self.timer = Timer.scheduledTimer(timeInterval: repeatInSeconds, target: self, selector: #selector(fireTimer), userInfo: nil, repeats: true)
         
+        RunLoop.current.add(self.timer!, forMode: .commonModes)
```
anwzhang: what is this change? do you remove it from the RunLoop in the stop method? If not, would it cause memory issues?

PR #151, opened 2019-03-04
[API 1.0] [MoPub] Banner ad is not rendered properly

yoalex5:
Fix for issue https://github.com/prebid/prebid-mobile-ios/issues/150

PR #156, opened 2019-03-12
[API 1.0] [MoPub] Banner ad is not rendered properly 

yoalex5:
duplicated
https://github.com/prebid/prebid-mobile-ios/pull/151


ppuviarasu:
merged

Issue #157, opened 2019-03-12
[API 1.0][Banner] Memory leak found


yoalex5:
There are strong references Cycles between BannerAdUnit <-> Dispatcher and Dispatcher <->Timer that also retains BannerController

<img width="289" alt="Screen Shot 2019-03-12 at 8 53 03 PM" src="https://user-images.githubusercontent.com/8141931/54227588-f12fb000-4508-11e9-82e6-b6490b3b3dae.png">
<img width="290" alt="Screen Shot 2019-03-12 at 8 52 02 PM" src="https://user-images.githubusercontent.com/8141931/54227589-f12fb000-4508-11e9-9321-3b1d75522091.png">
<img width="289" alt="Screen Shot 2019-03-12 at 8 50 41 PM" src="https://user-images.githubusercontent.com/8141931/54227590-f1c84680-4508-11e9-9c19-f7488bd9c3c8.png">
<img width="286" alt="Screen Shot 2019-03-12 at 8 50 29 PM" src="https://user-images.githubusercontent.com/8141931/54227591-f1c84680-4508-11e9-9877-4a6fff52606d.png">

PR #158, opened 2019-03-12
[API 1.0][Banner] Memory leak found

yoalex5:
Issue #157 

This variant eliminates memory leak but has a disadvantage. If a client code does not create a property on AdUnit it will be deallocated and refresh feature will not work.

### DIFF
--- a/example/Swift/PrebidDemo/PrebidDemo/BannerController.swift
+++ b/example/Swift/PrebidDemo/PrebidDemo/BannerController.swift
@@ -35,6 +35,8 @@ class BannerController: UIViewController, GADBannerViewDelegate, MPAdViewDelegat
     
     var mopubBanner: MPAdView?
     
+    var adUnit: AdUnit?
+    
     override func viewDidLoad() {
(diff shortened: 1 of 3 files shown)


ppuviarasu:
Why not create the BannerAdUnit as a member varaible instead of local and passing around? I dont understand why AdUnit needs to be created and the banner added to it

yoalex5:
@ppuviarasu 
>Why not create the BannerAdUnit as a member varaible instead of local and passing around?
This example just demonstrates that the client part must have a reference to BannerAdUnit

> I dont understand why AdUnit needs to be created and the banner added to it
In the other case deinit method will be called on AdUnit object (Dispatcher has a weak reference to DispatcherDelegate(that is BannerAdUnit -> AdUnit))

AntoineJac:
It is also fixed in my PR 155.

ppuviarasu:
should be resolved in PR https://github.com/prebid/prebid-mobile-ios/pull/160

PR #167, opened 2019-03-28
Lower the deployment target to iOS 9.0

ton-impalastudios:


bszekely1:
After consensus with the Prebid Mobil committee, the consensus was iOS versions lower than 11.0 represent less than 5% of user traffic with shrinking footprint. The extra testing and product support for earlier iOS versions were not worth effort.

Closing this issue.

PR #171, opened 2019-04-01
[MoPub] redundant `DispatchQueue.main.async` call was removed

yoalex5:


ppuviarasu:
added as part of this PR https://github.com/prebid/prebid-mobile-ios/pull/173. Hence closing

PR #184, opened 2019-04-26
Change host reference

yoalex5:
Default host was changed from PrebidHost.Appnexus to PrebidHost.Custom.
If a publisher does not set Prebid.shared.prebidServerHost the Log error will be printed
Screen Shot 2019-04-26 at 12 17 45 PM

The Doc hint looks like
Screen Shot 2019-04-26 at 12 25 26 PM


yoalex5:
@ppuviarasu Could you please review this PR ?

bszekely1:
For reference, the genesis behind this change is to default the Prebid Server Host to Custom, forcing the app developer to select a Host URL. No host provider should have any preference in an open system, in addition to it being confusing to app developers to have a default URL that may not match the hosting providing they wish to integrate with.

Issue #175, opened 2019-04-02
[AdManager][Banner] An ad is not rendered properly


yoalex5:
## Type of issue
Bug

## Description
When running a standard banner adunit it is not rendered by AdManager SDK. 

## Steps to reproduce
1 - Follow the [Guide to DFP setup](http://prebid.org/adops/step-by-step.html)
2 - Make a [Code Integration for iOS](http://prebid.org/prebid-mobile/pbm-api/ios/code-integration-ios.html)
3 - [Setup BannerAdUnit for Google Mobile Ads](http://prebid.org/prebid-mobile/pbm-api/ios/pbm-bannerad-ios.html)
4 - Run and check the result. When a banner ad is returned, the rendering fails.

## Test config
Host = Appnexus
ServerAccountId = "bfa84af2-bd16-4d35-96ad-31c6bb888df0"
configId = "6ace8c7d-88c0-4623-8117-75bc3f0a2e45"
adUnitID = /5300653/test_adunit_pavliuchyk_300x250_puc_ucTagData_prebid-server.rubiconproject.com

Charles log: [ios_Pv1.0_AMv7.42.2_prod_appnexus_rubicon-puc_ucTagData.chls.zip](https://github.com/prebid/prebid-mobile-ios/files/3038928/ios_Pv1.0_AMv7.42.2_prod_appnexus_rubicon-puc_ucTagData.chls.zip)

### Expected results
If AdManager response contains an ad it should be rendered

### Actual results
A creative is not showed

## Platform details
Prebid SDK v 1.0, git hash: beeee28699c268d021a7ad612ec6d60e64624f30
AdManager v 7.42.2

AntoineJac:
I am confirming the issue, the GADBannerView has the correct size but the GADWebAdView is keeping the 1x1 size. The "x-afma-ad-size" header used by the SDK to build the GADWebAdView, this header field is returning 1x1 instead of the real creative size. The other header "X-Afma-Ad-Slot-Size" has the correct size.
It is an issue with the DFP SDK and I raised it to google a few week ago.

I have found a fix using the mraid.resize() in the prebid creative but hope Google could fix it...

bszekely1:
To add additional details to the issue, Oleksandr who created this ticket has done some analysis on which versions of the Google ads SDK exhibit the failure to resize below:
[...]

PR #189, opened 2019-05-10
[iOS][AdManager][Banner] An ad is not rendered properly

yoalex5:
Issue #175

### DIFF
--- a/example/Swift/PrebidDemo/PrebidDemo/BannerController.swift
+++ b/example/Swift/PrebidDemo/PrebidDemo/BannerController.swift
@@ -108,6 +108,9 @@ class BannerController: UIViewController, GADBannerViewDelegate, MPAdViewDelegat
 
     func adViewDidReceiveAd(_ bannerView: GADBannerView) {
         print("adViewDidReceiveAd")
+        
+        self.dfpBanner.resize(bannerView.adSize)
+        
     }
 
     func adView(_ bannerView: GADBannerView, didFailToReceiveAdWithError error: GADRequestError) {
--- a/example/Swift/PrebidDemo/PrebidDemo/BannerController.swift
+++ b/example/Swift/PrebidDemo/PrebidDemo/BannerController.swift
@@ -109,7 +109,7 @@ class BannerController: UIViewController, GADBannerViewDelegate, MPAdViewDelegat
     func adViewDidReceiveAd(_ bannerView: GADBannerView) {
         print("adViewDidReceiveAd")
         
-        Utils.shared.resizeAdManagerBannerAdView(bannerView) { (size) in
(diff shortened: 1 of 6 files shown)


example/Swift/PrebidDemo/PrebidDemo/BannerController.swift
```diff
@@ -108,6 +108,9 @@ class BannerController: UIViewController, GADBannerViewDelegate, MPAdViewDelegat
 
     func adViewDidReceiveAd(_ bannerView: GADBannerView) {
         print("adViewDidReceiveAd")
+        
+        self.dfpBanner.resize(bannerView.adSize)
```
ppuviarasu: This needs to be documented here `http://prebid.org/prebid-mobile/pbm-api/ios/pbm-bannerad-ios.html` to include adManager and the implementation logics like these to address specific mediation partner issues

yoalex5:
The latest update to PR has the next logic:
1. When AdManager SDK retrieves a creative (`adViewDidReceiveAd` event) a client app call Utility method `findPrebidCreativeSize`  that is looking for `hb_size` key/value in the creative js code.
2. `findPrebidCreativeSize` calls Completion Handler that contains `CGSize` parameter with a value if the `hb_size` key/value was found  and `nil` in other cases

Please note that it is Utility function that is involved to call it from the client code when it is necessary

src/PrebidMobile/PrebidMobile/AdUnits/Utils.swift
```diff
@@ -89,6 +90,157 @@ public class Utils: NSObject {
...
+            let wkResult = self.findSizeInJavaScript(jsCode: value as? String)
+            
+            self.runResizeCompletion(size: wkResult, completion: completion)
+        })
+        
+    }
+    
+    func findSizeInWebViewAsync(uiWebView: UIWebView, completion: @escaping (CGSize?) -> Void) {
+        
+        Timer.scheduledTimer(withTimeInterval: 0.1, repeats: true) { (timer) in
```
ppuviarasu: This API only works for iOS 10 & above and not for iOS 9. Since we are downgrading to iOS 9, can you fix this?

yoalex5: Yes, I will upgrade the PR.

yoalex5:
@ppuviarasu, I have updated the PR. 
On Tuesday I am going to make full testing using different AdManagerSDK versions

yoalex5:
@ppuviarasu, I have finished sanity checks and it seems that it is workable fix. Could you please review this PR?

bretg:
docs PR https://github.com/prebid/prebid.github.io/pull/1352

Issue #192, opened 2019-05-13
UnitTests are broken


ThomWee:
I forked the project while updating the Swift support and need some further tweaks. But while doing so I came across the situation, that some tests are not working.

**Steps to reproduce**
Check out master. First Fix the `PrebidServerOneBidFromAppNexusOneBidFromRubicon.json` source path, since this is taken from the android project. After doing so and run all test the test `func testBidManagerRequestForBidResponseFromTwoBidders()` starts to fail: keyword.count is 15 and not as expected 16.

Now I changed the test scheme to run tests in random order and got 7 more tests failing.

Can you fix the tests so that I can continue on my fork to see that I don't break anything!?

yoalex5:
@ThomWee, I had fixed critical issues with UnitTests ant they have been merged to master. 

I am going to close this issue as well as create a new ticket with running UnitTest in random order #200

PR #193, opened 2019-05-13
Fixed unit tests

yoalex5:
The issue #192

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
 language: swift
-osx_image: xcode8
+osx_image: xcode10.2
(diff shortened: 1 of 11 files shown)


yoalex5:
@ThomWee Thanks for your review. We know about this behaviour and it is caused by using singleton pattern which has a state for all tests. We are going to fix it in the nearest future

yoalex5:
@ppuviarasu Could you please review this PR ?

yoalex5:
@ppuviarasu, thank you for approve PR. 
Could you please merge these changes?

Issue #195, opened 2019-05-17
Fix Travis CI


yoalex5:
**The problem**

Please review the latest build history
https://travis-ci.org/prebid/prebid-mobile-ios/builds

You can notice that the build status is SUCCESS. But the reality is different

For example API 1.0 PR has a green status
https://travis-ci.org/prebid/prebid-mobile-ios/builds/498809621

If you take a look at build logs you will see that 
1. buildPrebidMobile.sh script threw an error - `Error in build check log ../../build/out/log/prebid_mobile_build.log` but exit code is 0
2. testPrebidMobile.sh script finished with error - `xcodebuild: error: Unable to find a destination matching the provided destination specifier:` but exit code is 0
3. integration tests were not started because of testPrebidMobile.sh termination 

**Insight**
The build script consists of 3 main parts(except swift link and analysis tools)
1. build PrebidMobile
2. test PrebidMobile (Unit tests)
3. test PrebidDemo (Integration tests)

**I propose the next pipeline**
All steps should be done during CI process but the final result status has to take into account only `build PrebidMobile` and `test PrebidMobile`. If some of them was finished with error the build status will be `failed`. 
I don't include `test PrebidDemo` in this strict mode on purpose because integration tests sometimes can finished by `No bids` status or has a timeout. So we should first of all fix them and after that we will be able to include `test PrebidDemo` tests in this strict mode

PR #196, opened 2019-05-17
Fix Travis CI

yoalex5:
Issue #195

Notes
1. osx_image was changed from xcode8 to the latest xcode10.2
2.-arch i386 was removed because the Deployment target is 11 that doe not support 32-bit arch.
3. defines_module was changed to no to build with arch from -arch parameter
4. added exit 1 to xcodebuild command to explicitly specify an error instead of using the exit code from the latest command
5. xcpretty-travis-formatter was added
6. testPrebidMobile.sh was divided into testPrebidMobile.sh(unit tests) and testPrebidDemo.sh (integration tests) to handle exit codes separately 
7. fixed wrong unit test
8. fixed wrong file path
Added log information

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
 language: swift
-osx_image: xcode8
+osx_image: xcode10.2
 before_script: "./scripts/add-keys.sh"
 after_script: "./scripts/remove-key.sh"
 script:
-- "./scripts/swiftLint.sh"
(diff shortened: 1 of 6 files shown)


yoalex5:
@ppuviarasu could you please review this pull request ?

yoalex5:
@ppuviarasu Travis finished with success status. Could you please merge the PR ?

PR #197, opened 2019-05-22
[DemoApp] Appnexus host was added

yoalex5:
Since we changed default host to PrebidHost.Custom in PR #184 we should specify it in the client code


yoalex5:
@ppuviarasu could you please review this pull request ?

Issue #203, opened 2019-06-11
Fixing "version" value in the auction request


yoalex5:
The latest PrebidSDK version sends a wrong value "1.0" instead of "1.1" in the `app.ext.prebid.version` key

PR #204, opened 2019-06-11
Fixing "version" value in the auction request

yoalex5:
GitHub isuue #203


yoalex5:
@ppuviarasu, could you please review this PR ?

ppuviarasu:
We did a release without this 😢 
Can you check if you can revert the cocoapods push and do with the new one

yoalex5:
@bszekely1, what do you think about it ?

bszekely1:
Lets re-release this. It's typically not good practice to add items after release, however we are very close to release time and developers are not rushing to download right away. 

Please perform another release with the same tag as soon as possible. It will be included in 1.1

PR #205, opened 2019-06-11
Update getting PrebidSDK version approach

yoalex5:
Current implementation relies on autogenerated `PrebidMobileVersionNumber` field that is not workable with CocoaPods(The value is always 1).  

I propose to get the PrebidSDK version via `Bundle(for: type(of: self)).infoDictionary?["CFBundleShortVersionString"]` function. It returns value from sdk module's Info.plist that we change during new release


ppuviarasu:
The main idea about sending the version to the backend is to track how many requests we are getting from what version of the SDK. We cannot make it generic. It has to be modified for every release

yoalex5:
It seems that this PR is not ready because it sends now `1.1.0`. But in any case my changes do not break the idea of sending the actual PrebidSDK version. 
Or I have missed something ?

yoalex5:
@ppuviarasu, what do you think ? Can we send "version" in `X.Y.Z` format ? If yes we should merge this PR

ppuviarasu:
yes this should be fine

PR #211, opened 2019-06-19
Fix issue for Int.max overflow

AntoineJac:
For Iphone 5c and lower the processors is built on a 32-bit architecture. So any Int will be 32 bit. For these device we need to use Int64 as Date().timeIntervalSince1970 * 1000 will be out of range.


AntoineJac:
@yoalex5 , for review. Thanks,

yoalex5:
@AntoineJac, Thank you for your PR.
I have created an [optional change](https://github.com/AntoineJac/prebid-mobile-ios/pull/6)

yoalex5:
@ppuviarasu, could you please merge this PR ?

PR #216, opened 2019-07-04
Adapt resize feature for Objective-C client

yoalex5:
GitHub issue #215

What were changed:
1. `Utils.shared` was opened for Objective-C
2. A new method `AdViewUtils.findPrebidCreativeSize` was created to support `success` and `failure` handlers: 
3. The old resize function(`
findPrebidCreativeSize(_:success:)`) was marked as deprecated and it's still possible to use it (to support compatibility)

Screen Shot 2019-07-04 at 11 06 13 PM

4. Demo apps for Swift and Objective-C were updated to use the new function

The latest API for resize function looks like

Swift:
```
    func adViewDidReceiveAd(_ bannerView: GADBannerView) {
        print("adViewDidReceiveAd")

        AdViewUtils.findPrebidCreativeSize(bannerView,
                                            success: { (size) in
                                                guard let bannerView = bannerView as? DFPBannerView else {
                                                    return
                                                }

                                                bannerView.resize(GADAdSizeFromCGSize(size))

        },
                                            failure: { (error) in
                                                print("error: \(error)");

        })
}
```

Objective-C:
```
-(void) adViewDidReceiveAd:(GADBannerView *)bannerView {
    NSLog(@"Ad received");
    [AdViewUtils findPrebidCreativeSize:bannerView
                                   success:^(CGSize size) {
                                       if ([bannerView isKindOfClass:[DFPBannerView class]]) {
                                           DFPBannerView *dfpBannerView = (DFPBannerView *)bannerView;
                                           
                                           [dfpBannerView resize:GADAdSizeFromCGSize(size)];
                                       }
                                   } failure:^(NSError * _Nonnull error) {
                                       NSLog(@"error: %@", error);
                                   }];
}

```

Error codes:
```
unspecifiedCode = 101
noWebViewCode = 110
uiWebViewFailedCode = 121
wkWebViewFailedCode = 126
noHtmlCode = 130
noSizeObjectCode = 140
noSizeValueCode = 150
sizeUnparsedCode = 160
```

What to do:
1. Update docs

PR #220, opened 2019-07-16
Utils class improvements

Gaetan:
Formatted the Utils class.

Reworked the `findPrebidCreativeSize` method to have a success and failure handler. The method will always call one or the other.

In case of an error, the error message will be logged and passed to the failure handler.

### DIFF
--- a/src/PrebidMobile/PrebidMobile/AdUnits/Utils.swift
+++ b/src/PrebidMobile/PrebidMobile/AdUnits/Utils.swift
@@ -31,65 +31,65 @@ public class Utils: NSObject {
 
     }
 
-@objc public func removeHBKeywords (adObject: AnyObject) {
+    @objc public func removeHBKeywords (adObject: AnyObject) {
 
-    let adServerObject: String = String(describing: type(of: adObject))
-    if (adServerObject == .DFP_Object_Name || adServerObject == .DFP_O_Object_Name || 
-        adServerObject == .DFP_N_Object_Name || adServerObject == .GAD_N_Object_Name || 
-        adServerObject == .GAD_Object_Name) {
-        let hasDFPMember = adObject.responds(to: NSSelectorFromString("setCustomTargeting:"))
-        if (hasDFPMember) {
-            //check if the publisher has added any custom targeting. If so then merge the bid keywords to the same.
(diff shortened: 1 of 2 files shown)


yoalex5:
Hello @Gaetan!

Thanks for contributing. 
I have already fixed the bug with handler and it will be a part of #216 PR. Also this PR will include some internal improvements.

1. Could you please exclude refactoring changes
2. Have these changes fixed problems with random returns and empty blank ad as you described in a letter?

Thank you

Gaetan:
Hi @yoalex5,

The goal of this PR is to mainly share my vision on how I expect the `findPrebidCreativeSize` to work. I can adapt it in case you see value in the changes.

You say that #216 will include some internal improvements, does that mean that you are going to make additional changes to that PR?

1. The refactoring changes I propose here go a bit further than #216, there you would only get a message "cannot get size" in case of a failure. With my changes you can expect messages like:
 - "subView doesn't include WebView"
- "jsCode is nil"
- "Size cannot be parsed"
These messages should help the app developers to pin point what they are doing wrong.

Also at the moment, if the view does not contain a webview, the handler is not called. [Utils.swift#L108](https://github.com/prebid/prebid-mobile-ios/blob/a20e582b894ae97e4020e9fdd254585f10f7f7b7/src/PrebidMobile/PrebidMobile/AdUnits/Utils.swift#L108)
If the javascript execution in the webview fails, the handler is not called: [Utils.swift#L140](https://github.com/prebid/prebid-mobile-ios/blob/a20e582b894ae97e4020e9fdd254585f10f7f7b7/src/PrebidMobile/PrebidMobile/AdUnits/Utils.swift#L140)

2. These changes should allow us to pin point the problem we are facing. I will keep you up to date if we make progress.

Kind regards,

Gaetan

yoalex5:
I will merge your changes into my branch and after that I will propose it on review

yoalex5:
@Gaetan 

I have updated [the PR](https://github.com/prebid/prebid-mobile-ios/pull/216#commits-pushed-8ecf2d7) by your changes

Gaetan:
@yoalex5 Thank you, I will close this PR then.

PR #213, opened 2019-07-02
Fixing Unit tests

yoalex5:
GitHub issue #212

Issue #200, opened 2019-06-10
UnitTests in random order 


yoalex5:
If run UnitTests in random order some tests will be failed.

PR #221, opened 2019-07-17
Fix invalid stub for rubicon, and ensure correct prebid host in BidManagerTests.

SanderSander:
Currently the test `testRubiconBidManagerRequestForNoBidResponse` is failing silently, because the assertions pass when an invalid request is provided. 

Furthermore the tests can't be run individual because the default `PrebidHost` is custom (so empty, which results in an invalid host exception), but they succeed when executed in bulk/on travis because `Prebid.shared.prebidServerHost` is set in another test. 

This would make debugging easier!


yoalex5:
Possibly is duplicated of #213

SanderSander:
> Possibly is duplicated of #213

Yup you are right.

yoalex5:
As for running tests in a random order the #200 was created. We need some time to fix it completely.

SanderSander:
@yoalex5 

I ran all test separated, the only test outside `FetchingLogictest/BidManagerTests` is `PrebidTests::testServerCustomHost`

Where the first line is `XCTAssertEqual(Prebid.shared.prebidServerHost, PrebidHost.Appnexus)` 

A simple switch by testing in the first assert if the host is custom, and then set the host to Appnexus would do trick ;)

bszekely1:
I am closing this issue since it's a duplicate of https://github.com/prebid/prebid-mobile-ios/pull/213.

@SanderSander, please review 213.

PR #229, opened 2019-08-06
Fix: Cancel event is triggered to early on adUnit refresh

SanderSander:
Currently if you refresh the AdUnit in 1.9 seconds the timeout is triggered 0.1 seconds later (with a 2 second timeout setting) because the previous timeout is maintained and still queued as task. 

The check `if (!self.didReceiveResponse) {` also fails in this case, because by invoking `fetchDemand` the booleans are reset too false.  

This patch makes sure that the previous timeout task is canceled and a new timeout task is created with the correct timings.

### DIFF
--- a/src/PrebidMobile/PrebidMobile/AdUnits/AdUnit.swift
+++ b/src/PrebidMobile/PrebidMobile/AdUnits/AdUnit.swift
@@ -23,6 +23,8 @@ import ObjectiveC.runtime
 
     var dispatcher: Dispatcher?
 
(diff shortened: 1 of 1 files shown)


anwzhang:
Hi @SanderSander, can you resolve the merge conflict before I approve it?

bszekely1:
Hi @SanderSander I am temporarily closing this PR since this has taken no action in the last several months. If you wish to have this PR merged, please open it back up and resolve any merge conflicts.

PR #230, opened 2019-08-07
Refactoring/Improvement

yoalex5:
These changes generally refactor the FirstPartyData PR. Plus minor changes

1. `AdUnit.UserKeyword` uses `TargetingParams.UserKeyword` inside. The parameter `key` will be omitted
2. a code that is related to FPD was improved and refactored
3. tests were improved


littleV:
@yoalex5 Can you resolve the merge conflict on this PR?

anwzhang:
Hi @yoalex5, can you resolve the merge conflict on this branch?

src/PrebidMobile/PrebidMobile/AdUnits/AdUnit.swift
```diff
@@ -107,43 +105,44 @@ import ObjectiveC.runtime
...
-     * This method obtains the user keyword & value user for targeting
-     * if the key already exists the value will be appended to the list. No duplicates will be added
+     The func uses Targeting.addUserKeyword() inside.
+     - Parameters:
+        - key: parameter is omitted
+        - value: is passed to Targeting.addUserKeyword()
      */
-    @available(*, deprecated, message: "Please use Targeting.shared.addUserKeyword() method instead")
+    @available(*, deprecated, message: "Please use Targeting.addUserKeyword() method instead")
     public func addUserKeyword(key: String, value: String) {
```
anwzhang: if key is omitted, why not change the API to be just passing one String, same as Android?

bszekely1: yes, the new targeting keywords should not utilize a key. It should be a static string. I believe this may be a mistake.

yoalex5: Hello @anwzhang,

>if key is omitted, why not change the API to be just passing one String

I did not do it because it is a public API and I want to have a backward compatibility 

>same as Android

Android version has the same changes
https://github.com/prebid/prebid-mobile-android/blob/67df7e2f9be348f33147681454a1e94f66ec1a97/PrebidMobile/API1.0/src/main/java/org/prebid/mobile/AdUnit.java#L150

If you wish I can change PRs for iOS and Android what will entail breaking changes as a result we have to change a major version

yoalex5:
@anwzhang, I have resolved the merge conflicts

yoalex5:
@anwzhang, could you please review it ?

Issue #250, opened 2019-10-24
Support for US Privacy flag


bszekely1:
## Type of issue

As privacy regulation is taking on further adaption by governments, the IAB has devised a US Privacy signal to express a GDPR-like consent notice presented to the user for any US based privacy regulation. This intent to implement is created to provide support for CCPA using the US Privacy String Framework with a goal to implement before the end of 2019. 

To quote the IAB spec on the initial version of the US Privacy String:

> A U.S. Privacy String contains information about disclosures made and choices selected by a user regarding consumer data privacy under U.S. Privacy regulation. Version 1 of this specification only supports signals pertaining to the California Consumer Privacy Act (CCPA). 

>This specification was created because Digital Properties need a scalable way to establish and persist U.S. Privacy signals. Additionally, downstream vendors need a reliable way to access U.S. Privacy signals when running within a Digital Property’s website or app.

## Goals

- Provide support for California Consumer Privacy Act (CCPA)

## Proposed Design

The expectation is the publisher will have the responsibility of participating and complying with the IAB US privacy String as defined in the [IAB US Privacy User Guide ](https://docs.google.com/document/d/1vsbBROTbWgm0Z6TIHJ9tD6mzwUS_2I8WAzDrCqG9fTA/edit?pli=1#)

Workflow should be as follows:

1. Publisher collects CCPA consent from user
2. Publisher stores consent signal in [NSUserDefaults](https://developer.apple.com/documentation/foundation/nsuserdefaults#1664798?language=objc) as defined in the [IAB spec](https://docs.google.com/document/d/1BavBbqFGE0fC0-wdDG99IdyZtbifdp9Al_rZ-ra1xag/edit#heading=h.q35ewepn3965)
3. Prebid SDK reads from [NSUserDefaults](https://developer.apple.com/documentation/foundation/nsuserdefaults#1664798?language=objc) as defined in the [IAB spec](https://docs.google.com/document/d/1BavBbqFGE0fC0-wdDG99IdyZtbifdp9Al_rZ-ra1xag/edit#heading=h.q35ewepn3965) database for persistent storage
4. Prebid SDK passes signal to Prebid Server if present

### Prebid SDK Changes
*Requirements*

[...]

Issue #243, opened 2019-09-10
support for gdpr consent


seraph3b:
on Targeting.swift line 288

according to iab framework .IAB_GDPR_SubjectToConsent must be string "1" or "0", if you execute Bool(stringValue)! when stringValue is "1" or "0" will cause an exception.

it works only on stringValue = "true" or "false".

please unwrap it preventing unwanted errors and handle iab specification for consent parameter correctly

PR #256, opened 2019-11-15
Support for US Privacy flag

yoalex5:
1. CCPA was added #250
API:
``` Swift
//add
UserDefaults.standard.set("1YN", forKey: "IABUSPrivacy_String")

//remove
UserDefaults.standard.removeObject(forKey: "IABUSPrivacy_String")
```

OpenRTB request:
``` XML
"regs": {
  "ext": {
   "us_privacy": "1YN"
  }
}
```

2. fixed GDPR exception #243
3. `subjectToGDPR` can be assign to `nil` to get an IAB value
``` Swift
Targeting.shared.subjectToGDPR = nil
```
4. refactoring
5. tests were added

### DIFF
--- a/Source/RequestBuilder.swift
+++ b/Source/RequestBuilder.swift
@@ -342,22 +342,25 @@ import AdSupport
         return nil
     }
 
-    func openrtbRegs() -> [AnyHashable: Any]? {
+    func openrtbRegs() -> [AnyHashable: Any] {
 
         var regsDict: [AnyHashable: Any] = [:]
+        var ext: [AnyHashable: Any] = [:]
 
-        let gdpr = Targeting.shared.subjectToGDPR
-
-        if gdpr == true {
-            regsDict["ext"] = ["gdpr": NSNumber(value: gdpr).intValue] as NSDictionary
+        if Targeting.shared.subjectToGDPR == true {
+            ext["gdpr"] = 1
         }
+        
+        ext["us_privacy"] = StorageUtils.iabCcpa()
+        
+        regsDict["ext"] = ext
 
         let coppa = Targeting.shared.subjectToCOPPA
         if coppa == true {
             regsDict["coppa"] = NSNumber(value: coppa).intValue
         }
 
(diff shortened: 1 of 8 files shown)


Source/StorageUtils.swift
```diff
@@ -0,0 +1,83 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import Foundation
+
+class StorageUtils {
```
anwzhang: Should we call this PrivacyUtils?

yoalex5: `StorageUtils` is a general utility class to work with persistent storage. It will allow us to expand this class in future without generating a lot of small classes. If we had several big logic entities I would create `PrivacyUtils`, `StorageUtils` etc.

Please keep it as is and we will change it when it will be necessary

Source/RequestBuilder.swift
```diff
@@ -342,22 +342,25 @@ import AdSupport
...
 
-        let gdpr = Targeting.shared.subjectToGDPR
-
-        if gdpr == true {
-            regsDict["ext"] = ["gdpr": NSNumber(value: gdpr).intValue] as NSDictionary
+        if Targeting.shared.subjectToGDPR == true {
+            ext["gdpr"] = 1
         }
+        
+        ext["us_privacy"] = StorageUtils.iabCcpa()
```
ppuviarasu: Is this object `ext["us_privacy"]` created even if the ccpa is empty or nil?

yoalex5: If `StorageUtils.iabCcpa()` returns `nil` or empty string the `ext["us_privacy"]` will not be created

Check tests: `testPostDataWithEmptyCCPA` and `testPostDataWithoutCCPA`

PR #272, opened 2020-03-02
Tcf 2.0 changes

ppuviarasu:


yoalex5:
@ppuviarasu, are we going to modify test cases ?

Source/Targeting.swift
```diff
@@ -133,6 +133,17 @@ import CoreLocation
         }
     }
     
+    public var deviceAccessConsent: Bool? {
```
yoalex5: Am I correct, that a publisher has to handle `deviceAccessConsent` explicitly to get UUID  ?

ppuviarasu: yes the publisher has to explicitly set this purposeConsent to send the IDFA in the request

Source/RequestBuilder.swift
```diff
@@ -310,11 +310,15 @@ import AdSupport
         let lmtAd: Bool = !ASIdentifierManager.shared().isAdvertisingTrackingEnabled
         // Limit ad tracking
         deviceDict["lmt"] = NSNumber(value: lmtAd).intValue
-
-        let deviceId = RequestBuilder.DeviceUUID()
-        if deviceId != "" {
-            deviceDict["ifa"] = deviceId
+        
+        
+        if Targeting.shared.deviceAccessConsent == true {
```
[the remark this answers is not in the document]
ppuviarasu: handled

Source/RequestBuilder.swift
```diff
@@ -481,10 +506,7 @@ import AdSupport
 
         if (uuidString == "") {
             let advertisingIdentifier: String = ASIdentifierManager.shared().advertisingIdentifier.uuidString
-
-            if (advertisingIdentifier != .kIFASentinelValue) {
```
ppuviarasu: even if the device id is "0000" the value has to be passed in the jsonBody. removed the conditional checks

Source/RequestBuilder.swift
```diff
@@ -310,12 +310,37 @@ import AdSupport
...
+                            deviceAccessConsent=true  deviceAccessConsent=false  deviceAccessConsent undefined
+         gdprApplies=false        Yes, read IDFA       No, don’t read IDFA           Yes, read IDFA
+         gdprApplies=true         Yes, read IDFA       No, don’t read IDFA           No, don’t read IDFA
+         gdprApplies=undefined    Yes, read IDFA       No, don’t read IDFA           Yes, read IDFA
+         */
+          
+        var setDeviceId:Bool = false
+        let purposeConsents:String? = Targeting.shared.purposeConsents
+        
+        if (purposeConsents == nil && Targeting.shared.subjectToGDPR == false) {
```
yoalex5: what about the case when  `gdprApplies=undefined` and `deviceAccessConsent undefined` ?

Source/StorageUtils.swift
```diff
@@ -64,7 +79,25 @@ class StorageUtils {
...
+        var gdprConsentString:String? = getObjectFromUserDefaults(forKey: StorageUtils.IABTCF_ConsentString)!
+        
+        if(gdprConsentString == nil || gdprConsentString == String.EMPTY_String){
+            gdprConsentString = getObjectFromUserDefaults(forKey: StorageUtils.IABConsent_ConsentStringKey)!
+        }
+        return gdprConsentString
+        
+    }
+    
+    static func setPurposeConsents(value: String?) {
```
yoalex5: maybe we can use Bool instead of String ?

Source/Targeting.swift
```diff
@@ -133,6 +133,27 @@ import CoreLocation
...
+
+        get {
+            var savedConsent: String?
+            
+            if let pbString = StorageUtils.pbPurposeConsents() {
+                savedConsent = pbString
+            } else if let iabString = StorageUtils.iabPurposeConsents() {
+                savedConsent = iabString
+            }
+            if(savedConsent != nil){
```
yoalex5: can we move this logic from getter to `RequestBuilder` class ?

Source/RequestBuilder.swift
ppuviarasu: This case is handled similar to the gdprApplies = no && deviceConsent = no

Source/StorageUtils.swift
ppuviarasu: purposeConsent is a string like this "1010101001" from this only the 1st digit is the PurposeConsent all other are for different consents provided

Source/Targeting.swift
ppuviarasu: I need to check if the value is set by the publisher or are we accessing from the user preference. and the string that i extract from consent string, I only need the 1st char which represents the purposeConsents

Source/RequestBuilder.swift
yoalex5: Yes I see this behaviour in the code, the question in logic:

`gdprApplies = no && deviceConsent = no`  - result: `No, don’t read IDFA`

table above describes 
`gdprApplies = undefined && deviceConsent = undefined` - result: `Yes, read IDFA`

Source/StorageUtils.swift
yoalex5: Where can I found this information ? I mean where is described that the first char is a purposeConsent ?

ppuviarasu: https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/master/TCFv2/IAB%20Tech%20Lab%20-%20Consent%20string%20and%20vendor%20list%20formats%20v2.md

Source/StorageUtils.swift
```diff
@@ -28,6 +28,12 @@ class StorageUtils {
 
     static let IABConsent_ConsentStringKey = "IABConsent_ConsentString"
     
+    //TCF 2.0 variables
+    static let IABTCF_ConsentString = "IABTCF_TCString"
+    static let IABTCF_SubjectToGDPR = "IABTCF_gdprApplies"
+    
+    static let IABTCF_DeviceAccessConsent = "IABTCF_DeviceAccessConsent"
```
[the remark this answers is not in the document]
ppuviarasu: modified...

yoalex5:
@ppuviarasu, could you please review #279 ?

yoalex5:
@ppuviarasu, should we add a PB setPurposeConsents() method ?

ppuviarasu:
isnt that not doable using the property `purposeConsents` in targeting. I think we have a setter and getter there

yoalex5:
We have an `IAB` getter and do not have a `PB` setter

Source/Targeting.swift
```diff
@@ -133,6 +127,53 @@ import CoreLocation
         }
     }
     
+    // MARK: - TCFv2
+    
+    public var purposeConsents: String? {
+        set {
+            StorageUtils.setPurposeConsents(value: newValue)
```
ppuviarasu: Isnt this doing the PB setter?

Source/StorageUtils.swift
```diff
@@ -64,7 +97,26 @@ class StorageUtils {
...
+        if (gdprConsentString ?? "").isEmpty {
+            gdprConsentString = getObjectFromUserDefaults(forKey: StorageUtils.IABConsent_ConsentStringKey)
+        }
+
+        return gdprConsentString
+        
+    }
+    
+    static func setPurposeConsents(value: String?) {
+        setUserDefaults(value: value, forKey: StorageUtils.PBConsent_PurposeConsentsStringKey)
```
ppuviarasu: this does the setting. I am not sure what i am missing.

yoalex5:
Sorry, my mistake

PR #278, opened 2020-03-17
TCFv2 improvements

yoalex5:
Upstream PR #272

- `Targeting.shared.getDeviceAccessConsent()` was added 
- tests were added
- refactoring

Also I have a question regarding to `IABTCF_gdprApplies`

Official doc says `Number`, now we are using `String`. What do you think ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,10 @@ before_install:
   - "gem install cocoapods"
 before_script: 
   - "./scripts/add-keys.sh"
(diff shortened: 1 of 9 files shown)


yoalex5:
@ppuviarasu, what about a type of `IABTCF_gdprApplies` ?

ppuviarasu:
Yes doing it now...

ppuviarasu:
@yoalex5 I am providing a fix for the same. thanks for pointing out

PR #284, opened 2020-03-26
UIWebView was removed

yoalex5:
>ITMS-90809: Deprecated API Usage - Apple will stop accepting submissions of apps that use UIWebView APIs . See https://developer.apple.com/documentation/uikit/uiwebview for more information.

1. `UIWebView` was removed from SDK
2. `WKWebView` was set by default for AM SDK

**AdManager**
[AdManager Release notes](https://developers.google.com/ad-manager/mobile-ads-sdk/ios/rel-notes)
> 7.55.0 | 2020‑02‑04 | Removed all references to UIWebView. UIWebView is no longer supported.

**MoPub**
[MoPub issue](https://github.com/mopub/mopub-ios-sdk/issues/285)
[MoPub Release notes](https://developers.mopub.com/publishers/ios/changelog/)
>Version 5.9.0 (September 16, 2019) Totally remove UIWebView

`AdViewUtils.findPrebidCreativeSize` works only with `WKWebView`.
`AdManager`: `gad_preferred_webview` key with `wkwebview` value or v7.55.0(or higher)
```
//Info.plist 
<key>gad_preferred_webview</key>
<string>wkwebview</string>
```
`MoPub`: v5.9.0(or higher)

### DIFF
--- a/Example/PrebidDemo/PrebidDemoObjectiveC/Info.plist
+++ b/Example/PrebidDemo/PrebidDemoObjectiveC/Info.plist
@@ -41,8 +41,6 @@
 		<string>UIInterfaceOrientationLandscapeLeft</string>
 		<string>UIInterfaceOrientationLandscapeRight</string>
 	</array>
-	<key>GADIsAdManagerApp</key>
-	<true/>
 	<key>UISupportedInterfaceOrientations~ipad</key>
 	<array>
 		<string>UIInterfaceOrientationPortrait</string>
@@ -50,5 +48,11 @@
 		<string>UIInterfaceOrientationLandscapeLeft</string>
 		<string>UIInterfaceOrientationLandscapeRight</string>
 	</array>
+
+    <key>GADIsAdManagerApp</key>
+    <true/>
+    <key>gad_preferred_webview</key>
+    <string>wkwebview</string>
+
 </dict>
 </plist>
--- a/Source/Addendum/AdViewUtils.swift
+++ b/Source/Addendum/AdViewUtils.swift
@@ -27,19 +27,16 @@ public final class AdViewUtils: NSObject {
     @objc
     public static func findPrebidCreativeSize(_ adView: UIView, success: @escaping (CGSize) -> Void, failure: @escaping (Error) -> Void) {
         
-        let view = self.findWebView(adView) { (subView) -> Bool in
-            return isWebView(subView)
+        let view = self.findView(adView) { (subView) -> Bool in
+            return isWKWebView(subView)
         }
         
         if let wkWebView = view as? WKWebView  {
             Log.debug("subView is WKWebView")
             self.findSizeInWebViewAsync(wkWebView: wkWebView, success: success, failure: failure)
             
-        } else if let uiWebView = view as? UIWebView {
-            Log.debug("subView is UIWebView")
-            self.findSizeInWebViewAsync(uiWebView: uiWebView, success: success, failure: failure)
         } else {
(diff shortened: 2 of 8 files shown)


Example/PrebidDemo/PrebidDemoObjectiveC/Info.plist
```diff
@@ -41,14 +41,18 @@
...
 	<array>
 		<string>UIInterfaceOrientationPortrait</string>
 		<string>UIInterfaceOrientationPortraitUpsideDown</string>
 		<string>UIInterfaceOrientationLandscapeLeft</string>
 		<string>UIInterfaceOrientationLandscapeRight</string>
 	</array>
+
+    <key>GADIsAdManagerApp</key>
+    <true/>
+    <key>gad_preferred_webview</key>
```
ppuviarasu: Do we need to set this even if UIWebView is removed completely from the SDK?

Source/Addendum/AdViewUtils.swift
```diff
@@ -27,19 +27,16 @@ public final class AdViewUtils: NSObject {
     @objc
     public static func findPrebidCreativeSize(_ adView: UIView, success: @escaping (CGSize) -> Void, failure: @escaping (Error) -> Void) {
         
-        let view = self.findWebView(adView) { (subView) -> Bool in
-            return isWebView(subView)
```
ppuviarasu: I think this is still relevant. It should be isWebView irrestpective of WKWebView or UIWebView...

Source/Addendum/AdViewUtils.swift
```diff
@@ -52,22 +49,22 @@ public final class AdViewUtils: NSObject {
         failure(error)
     }
     
-    static func findWebView(_ view: UIView, closure:(UIView) -> Bool) -> UIView? {
```
ppuviarasu: Same... we are finding recursively for WebView so the naming is apt... lets keep it as is

Source/Addendum/AdViewUtils.swift
```diff
@@ -147,8 +128,8 @@ public final class AdViewUtils: NSObject {
         return matchAndCheck(regex: AdViewUtils.sizeValueRegexExpression, text: hbSizeObject)
     }
     
-    static func isWebView(_ view: UIView) -> Bool {
```
ppuviarasu: Let the naming remain same

Example/PrebidDemo/PrebidDemoObjectiveC/Info.plist
yoalex5: This case is useful when AM SDK has a version that is lower then `v7.55.0` where `UIWebView` is used by default

Source/Addendum/AdViewUtils.swift
yoalex5: In my opinion `WebView` is a common definition which covers the whole set of classes(`UIWebView`, `WKWebView`) with corresponded functionality. 

Theoretically `UIWebView` can be passed into it and success result is expected. That is why I decided to specify a concrete type(`WKWebView`) for this logic

Source/Addendum/AdViewUtils.swift
yoalex5: This function can be used for different purposes 

For example:
``` Swift
let view = self.findWebView(myView) { (subView) -> Bool in
    return subView is UIButton
}
```

Source/Addendum/AdViewUtils.swift
yoalex5: `isWKWebView` has more clear naming

yoalex5:
@ppuviarasu, could you please review my comments ?

PR #315, opened 2020-08-27
Dr.prebid native support

ppuviarasu:
1. Support for NativeAssets in Objective C
2. Dr.Prebid updated to support Native testing


Source/AdUnits/Native/NativeAsset.swift
```diff
@@ -123,29 +123,20 @@ public class NativeAsset: NSObject {
     }
 }
 
-@objc public enum ImageAsset: Int {
-    case Icon = 1
-    case Main = 3
-    case Custom
+public class ImageAsset: SingleContainerInt {
```
yoalex5, line 126: why do you extend SingleContainerInt ?

ppuviarasu: I had replaced all enums with class objects as it was not compatible with ObjectiveC apps. hence took the SingleContainerInt out of signals so that it can be used by all asset objects

anwzhang (approved):

yoalex5:
@ppuviarasu, are you going to merge this PR ?

ppuviarasu:
Done..

PR #322, opened 2020-09-15
Dr.prebid custom host

anwzhang:


ppuviarasu (approved):

yoalex5 (commented):
I noticed two UI bugs:
1. SettingsTableViewController is not full screen(from 1OS 13)
2. "Run Test" button is disabled when you fill in all fields at first time

PR #326, opened 2020-09-24
ATT and IDFV

yoalex5:


ppuviarasu (approved):

Source/RequestBuilder.swift
```diff
@@ -492,6 +496,12 @@ class RequestBuilder: NSObject {
         deviceExtPrebid["interstitial"] = deviceExtPrebidInstlDict
         deviceExt["prebid"] = deviceExtPrebid
 
+        if #available(iOS 14, *) {
+            deviceExt["atts"] = ATTrackingManager.trackingAuthorizationStatus.rawValue
+        }
+        
+        deviceExt["ifv"] = UIDevice.current.identifierForVendor?.uuidString
```
ppuviarasu, line 503: The intent to implement states that the IDFV will be got from an API exposed to the publisher. But this does the other. Can we change the logic?

ppuviarasu (dismissed):

Source/RequestBuilder.swift
bszekely1: The intent to implement states it is the job of the bid adapter if they wish to read this, not so much provide a toggle to allow publishers to opt-in allowing Prebid SDK to read IDFV. I do agree it should be in the spirit Prebid to provide the flexibility to publisher if they wish to pass Bid Adapters the IDFV, especially since Apple remains clear on their direction to not use IDFV for ad serving purposes: https://developer.apple.com/app-store/user-privacy-and-data-use/

While Apple policy is not law, Prebid SDK should not put publishers in such a compromising position by default. 

IDFV was not part of the purpose of the iOS14 release. Lets remove the IDFV logic for now and release just the atts and lmt items. We can discuss IDFV for another day.

yoalex5: `idfv` was removed from this PR

anwzhang (approved):

bszekely1 (approved):

bszekely1 (approved):

PR #343, opened 2021-01-08
Log prebid version instead of printing it

marcelklinek:
Proposing that we log version of Prebid instead of printing it. This way we can control whether we want to see it in console or not. If you feel this is unnecessary information that is just hanging there by mistake, please consider to drop these two lines altogether.


yoalex5 (approved):

marcelklinek:
@bszekely1 thanks for moving this PR forward! Can we trigger the CI tests again so that merging is not blocked? I checked Travis output and the failing tests do not seem to be related to this PR.

Feel free to tell me what to do so that this makes it into a release. This is my first time contributing and I have zero knowledge of the process 🤷‍♂️

bszekely1:
I will defer to @yoalex5 to help resolve any errors.

PR #344, opened 2021-01-11
SKAdNetwork POC

yoalex5:
Publisher API:
``` Swift
Targeting.shared.itunesID = "880047117"
Targeting.shared.skAdNetVersions = ["2.0", "2.1"]
Targeting.shared.skAdNetListMax = 306
Targeting.shared.skAdNetListExcl = [2, 8, 10, 55]
Targeting.shared.skAdNetListAddl = ["cDkw7geQsH.skadnetwork", "qyJfv329m4.skadnetwork"]

class BannerController: UIViewController {
    private let prebidSKAdNetworkHelper = PrebidSKAdNetworkHelper()
    
    override func viewWillAppear(_ animated: Bool) {
        prebidSKAdNetworkHelper.viewAppear()
    }

    override func viewDidDisappear(_ animated: Bool) {
        prebidSKAdNetworkHelper.viewDisappear()
    }

    //MARK: - GADBannerViewDelegate
    func bannerViewDidReceiveAd(_ bannerView: GADBannerView) {
        if #available(iOS 14.0, *) {
            prebidSKAdNetworkHelper.subscribeOnAdClicked(viewController: self, adView: bannerView)
        }
    }
}
```

Main concern in the current implementation is  - polling click handler. SDK starts a timer for polling but it should be managed in a correct way
1. User clicked on ad -> user was redirected to another app(e.g. Safari). Timer should be stopped because publisher's app is on background
2. use case 1 -> user came back from Safari to publisher's app(foreground state). Timer should be restarted because user can click on the rendered ad again


itiagrawal1:
> Main concern in the current implementation is - polling click handler. SDK starts a timer for polling but it should be managed in a correct way
> 
> 1. User clicked on ad -> user was redirected to another app(e.g. Safari). Timer should be stopped because publisher's app is on background
> 2. use case 1 -> user came back from Safari to publisher's app(foreground state). Timer should be restarted because user can click on the rendered ad again

We can have one SKStoreViewController and timer for one ad Object. The flow can look like the following:
1. Create 3Party adObject.
2. Call load
3. load Successful.
4. Check all the skanParams, if SDK is capable of doing on-click attribution. Inform JS to intercept all the appstore entry points.
5. Create SKStoreViewcontroller, call loadProduct but dont show the VC.
6. Create timer, start polling to check if any entry point is triggered.
7. If openSKStore is triggered, stop timer and show SKStoreViewcontroller.
8. User exit SKStoreViewcontroller, start timer again.
9. App goes in background, stop timer.
10. App comes in foreground with ad on it, start timer.
11. Tightly couple timer and adObject, with adobject destroy timer as well.

yoalex5:
@itiagrawal1, thanks for your response

>9. App goes in background, stop timer.
10. App comes in foreground with ad on it, start timer.

which exactly techniques do you propose to use to recognise background and foreground events?

>11. Tightly couple timer and adObject, with adobject destroy timer as well.
From Prebid SDK perspective `adObject` is `AdUnit` object? And do you propose to override `deinit()` method to destroy timer ?

Iti, can I ask you to duplicate 3 main points of improving this PR which you presented on our call ?

Thank  you in advance

itiagrawal1 (commented):
> @itiagrawal1, thanks for your response
> 
> > 1. App goes in background, stop timer.
> 
> 1. App comes in foreground with ad on it, start timer.
> 
> which exactly techniques do you propose to use to recognise background and foreground events?
> 
> > 1. Tightly couple timer and adObject, with adobject destroy timer as well.
> >    From Prebid SDK perspective `adObject` is `AdUnit` object? And do you propose to override `deinit()` method to destroy timer ?
> 
> Iti, can I ask you to duplicate 3 main points of improving this PR which you presented on our call ?
> 
> Thank you in advance

We can use observer to get notification when app goes foreground or background.
eg: 
NotificationCenter.default.addObserver(self, selector: #selector(stopTimer), name: UIApplication.didEnterBackgroundNotification, object: nil)
        NotificationCenter.default.addObserver(self, selector: #selector(startTimer), name: UIApplication.willEnterForegroundNotification, object: nil)

Inside "startTimer" function you can add checks for skStore VC's state, if it present on the screen then dont start the timer.


I have added all the 3 comments, which I mentioning in the meeting.

yoalex5:
@itiagrawal1, thanks

> 10. App comes in foreground with ad on it, start timer.

How do you determine that the ad is in front of the user ?

yoalex5:
@itiagrawal1, based on your comments I have updated my PR

Key points:

- Object to work with SKAdNetwork
``` Swift
let prebidSKAdNetworkHelper = PrebidSKAdNetworkHelper()
```

- Application state to manage PollingTimer is used internally and based  on
``` Swift
viewController.viewIfLoaded?.window
//and
NotificationCenter.default.addObserver(self, selector: #selector(stopTimer), name: UIApplication.didEnterBackgroundNotification, object: nil)
NotificationCenter.default.addObserver(self, selector: #selector(startTimer), name: UIApplication.willEnterForegroundNotification, object: nil)
```

- Controller state should be passed by publisher explicitly to manage PollingTimer
```
prebidSKAdNetworkHelper.viewAppear()
prebidSKAdNetworkHelper.viewDisappear()
```

- Open URL:
```
href tag
window.open
Window.mraid.open
```

- Appstore URL:
```
//itunes.apple.com
//apps.apple.com
```

yoalex5:
@ppuviarasu, proposed to use `NSURLProtocol` to intercepting network requests. Problem is that it has effect with `UIWebView` but [doesn't work with `WKWebView`](https://bugs.webkit.org/show_bug.cgi?id=138169) which is used by GAM

Source/Addendum/AdViewUtils.swift
```diff
@@ -15,10 +15,286 @@
...
+
+                            console.log(\"default window.open\");
+                            window.open_(url)
+                        }
+                    }
+
+                    // window.mraid.open
+                    window.mraidOpen_ = window.mraid.open;
+                    window.mraid.open = function(url) {
+                        if (url.includes(\"//apps.apple.com\") || url.includes(\"//itunes.apple.com\")) {
```
yoalex5, line 145: @YuriyVelichkoOpenX mentioned that this code does not solve redirection issue

Source/Addendum/AdViewUtils.swift
```diff
@@ -124,16 +407,32 @@ public final class AdViewUtils: NSObject {
         return matchAndCheck(regex: AdViewUtils.sizeObjectRegexExpression, text: text)
     }
     
+    static func findHbIdObject(in text: String) -> String? {
+        return matchAndCheck(regex: "ucTagData.uuid = \"(.*?)\"", text: text, rangeAt: 1)
```
yoalex5, line 411: @YuriyVelichkoOpenX mentioned that we should take into account `ucTagData.targetingMap` instead because `ucTagData.uuid` can be missing

bszekely1: The targetingMap is not a good place to look because of the unpredictable behavior GAM has with the targetingMap object. If GAM receives too much data in the payload to the ad server, some data is truncated. It's best to pass the data to the GAM creative as follows:

  ```
var ucTagData = {};
  ucTagData.adServerDomain = "";
  ucTagData.pubUrl = "%%PATTERN:url%%";
  ucTagData.adId = "%%PATTERN:hb_adid%%";
  ucTagData.cacheHost = "%%PATTERN:hb_cache_host%%";
  ucTagData.cachePath = "%%PATTERN:hb_cache_path%%";
  ucTagData.uuid = "%%PATTERN:hb_cache_id%%";
  ucTagData.mediaType = "%%PATTERN:hb_format%%";
  ucTagData.env = "%%PATTERN:hb_env%%";
  ucTagData.size = "%%PATTERN:hb_size%%";
  ucTagData.winurl = "%%PATTERN:hb_winurl%%";
  ucTagData.winbidid = "%%PATTERN:hb_bidid%%";
```

Also, even if you retrieve the data in GAM from the targeting map, the ucTagData should have the uuid as an object stored there.

**Comment by YuriyVelichkoOpenX** [comment]
regarding the discussion on the PMC call, I would like to note two suggestion which will improve the current implementation:

1.  Do not check the click URL for affiliation to the App Store. If the script detects the external URL for opening - set the isClicked to true.
2. Instead of polling the status of the isClicked property continuously add a UITapGestureRecognizer to the WKWebView and **check the status only after respective event**. 

These steps should make the solution more simple without losing functionality.

**Comment by yoalex5** [comment]
Banner ad solution without polling
 - `navigationDelegate` is used 
 - polling is removed

This fix doesn't have effect on video ad type(out-stream, instream)
We have next problems:
- It is not possible to identify if the ad is a Prebid ad. Since VAST tag URL is used inside GAM creative it is not possible to check in runtime if it has some Prebid markers(e.g. `hb_` is used in  banner ad format)
- It seems that `Learn More` button uses native bridge for opening URL

File: Source/Addendum/AdViewUtils.swift [hunk 3]
```diff
@@ -15,10 +15,162 @@
...
+            return AdViewUtils.isWKWebView(subView)
+        }
+        
+        guard let wkWebView = view as? WKWebView else {
+            Log.warn("View doesn't contain WKWebView")
+            return
+            
+        }
+
+        wkWebView.navigationDelegate = self
```
YuriyVelichkoOpenX, line 45: In this point, GMA SDK losses the "connection" with web view and GMA SDK is not able to process navigation events from the own web view anymore. Is it expected behavior?

yoalex5: It is true 

Do you think that GAM uses `navigationDelegate` with some vital logic? If I were GAM developer I would not expect a big profit of the approach which can  be overrode by any developer out of SDK

YuriyVelichkoOpenX: The MRAID and click behaviour is based on this delegate.

When user taps on the ad the rendering SDK analyses the event and performs the respective action, f.e - opens a clickthrough with the click URL or performs MRAID action.

Here is an example of this method in the rendering module
https://github.com/openx/prebid-mobile-ios/blob/feature-rendering-module/PrebidMobile/Rendering/PrebidMobileRendering/PrebidMobileRendering/AdTypes/AdView/PBMWebView.m#L277

Here is MoPub's one:
https://github.com/mopub/mopub-ios-sdk/blob/3e1382ed6686c59cf83d259f2bd8796de6c88c55/MoPubSDK/Internal/HTML/MPWebView.m#L387

yoalex5: I see, then I should rollback my changes

YuriyVelichkoOpenX: Have you considered the method swizzling?

https://nshipster.com/method-swizzling/

I do not really like this approach but It could be a good option in this case.

yoalex5: which method do you propose to swizzle ?

YuriyVelichkoOpenX: However, it  is not clear what to "swizzle" here  :(

YuriyVelichkoOpenX: we can search the needed class, as described here:
https://stackoverflow.com/questions/30044366/method-swizzling-for-protocols

but it looks too tricky.

PR #354, opened 2021-02-25
2.0 release

ppuviarasu:


yoalex5:
Hello @ppuviarasu 
1. Why does it contains so much commits ? As I see it contains only `in app native` feature. Was rebase used ?
2. Why it has `2.0` release ? I thought that it should `1.2`

ppuviarasu:
Invalid

PR #356, opened 2021-02-26
upgraded the min deployment target to 10.0

ppuviarasu:
- ERROR | [PrebidMobile/core] xcodebuild:  PrebidMobile/Source/TrackerInfo.swift:34:33: error: 'scheduledTimer(withTimeInterval:repeats:block:)' is only available in iOS 10.0 or newer


Kowshickkarthick (approved):

yoalex5 (approved):
If there is no other way and based on
https://developer.apple.com/support/app-store/

PR #370, opened 2021-04-03
Location Manager to use Last known location

yoalex5:
`CLLocationManager().location` is used instead of starting LocationManager


YuriyVelichkoOpenX (approved):
LGTM

ppuviarasu (approved):

PR #371, opened 2021-04-04
Bitcode issue

yoalex5:
- `BITCODE_GENERATION_MODE` used-defined setting was added for PrebidMobile and core modules with `marker` by default
- `buildPrebidMobile` script was improved for asking about bitcode


rakesthedon:
Why not just enable bitcode as a default and not go through the script? what is the advantages?

yoalex5:
I want to propose a choice for consumers. If you decide not to use bitcode you should not edit script

rakesthedon:
It is mandatory since iOS 10. i think this should be the enabled by default. most if not all third party framework are already enabling bitcode support.

yoalex5:
@rakesthedon, can you point me out of the official resource about mandatory bitcode ?

I found next:
>For iOS apps, bitcode is the default, but optional. For watchOS and tvOS apps, bitcode is required.

https://help.apple.com/xcode/mac/current/#/devbbdc5ce4f

rakesthedon:
It was my understanding but stand corrected. Can the default value be set to yes then?

yoalex5:
done

rakesthedon:
All that missing is the documentation update but it seems correct to me

ppuviarasu (approved):

Issue #347, opened 2021-02-04
User ID suport


bszekely1:
## Type of issue

Intent to implement

## Description

As the advertising ecosystem evolves, publishers have been working with identity vendors in prebid javascript to pass buyers user identifiers. Prebid Mobile faces a similar challenge as desktop environments, however mobile apps do not utilize cookies to store user identifiers, but instead rely on the hard-coded device IDs. The availability of device IDs (used for targeting, frequency capping and attribution), are under threat from regulation and other exterior forces (such as Apple). The intent here is to replicate the user ID support found in prebid.js and prebid server over to Prebid SDK.

## Goals

- Define a publisher facing interface to allow publishers to pass user IDs derived from vendor integrations
- Pass Prebid Server all vendor IDs using the pre-defined user-ID namespace

## Proposed Design

The job of Prebid SDK:
1. Provide an API interface where publishers can pass a user ID from one or more vendors both real-time or in a local storage
1. The public facing API requirements are the following:
    1. This public interface shall support multiple vendors, example a publisher can work with two different venders who each have their own ID schema
    1. Identifier information will be free form, meaning one vendor may require an attribute the other does not
        - Prebid SDK will not hard code any pre-defined vendor seutp
    1. Provide a global namespace in the Prebid SDK API that applies to all ad unit types at runtime
1. Provide a local storage schema for publishers to follow:
	1. Any identify vendor's details in local storage will be sent over to PBS as is unadultered
	1. If data is sent in the API and entered in local storage, the API detail will prevail
1. Prebid SDK should still send the bid request to PBS in LMT, DNT, ATTS != 3 and even in privacy regulation modes as PBS will enforce any annonymization of the request

The job of the publisher:
1. Seek arrangements / integration with vendors to fetch / generate vendor user ID
1. Provide Prebid SDK via local storage or real-time each vendor ID
1. Enforce any expiration or opt out as described with each vendor
    - Prebid Server will do it's part to annonymize requests pertaining to user IDs in scenarios such as LMT or other privacy regulation

The job of Prebid Server:
1. PBS should forward all `user.ext.eids[]` objects to all bidders for that auction if they are recieved in the bid request
1. PBS should suppress IDs from bidders in the following cases:
    - LMT or DNT =1
    - In Privacy Regulation modes such as GDPR or CCPA
    - Publisher restricts eid to specific bidders

### Call to PBS
Below is an example call to PBS:

```
{
    "source": {
[...]

PR #373, opened 2021-04-05
Local storage option to store External UserId Solutions

avermaiOS:
### Issue
Implement the local storage option to store External UserId Solutions in Prebid Mobile SDK	

### Description of Fix
https://corpwiki.xandr-services.com/display/CT/%5BiOS%5D+Local+Storage+Schema+for+External+Ids+in+Prebid+Mobile+SDK

### Unit Tests
Done

### DIFF
--- a/Source/RequestBuilder.swift
+++ b/Source/RequestBuilder.swift
@@ -479,8 +479,9 @@ class RequestBuilder: NSObject {
         var externalUserIdArray = [ExternalUserId]()
         if Prebid.shared.externalUserIdArray.count != 0 {
             externalUserIdArray = Prebid.shared.externalUserIdArray
-        }else if let externalUserIds = Targeting.shared.externalUserIds, externalUserIds.count != 0{
-            externalUserIdArray = externalUserIds
+        }
+        else if Targeting.shared.externalUserIds.count != 0{
+            externalUserIdArray = Targeting.shared.externalUserIds
         }
         var transformedUserIdArray = [[AnyHashable: Any]]()
         for externaluserId in externalUserIdArray {
--- a/Source/RequestBuilder.swift
+++ b/Source/RequestBuilder.swift
@@ -475,7 +475,13 @@ class RequestBuilder: NSObject {
     }
     
     func getExternalUserIds() -> [[AnyHashable: Any]]? {
-        let externalUserIdArray : [ExternalUserId] = Prebid.shared.externalUserIdArray
+       
+        var externalUserIdArray = [ExternalUserId]()
+        if Prebid.shared.externalUserIdArray.count != 0 {
+            externalUserIdArray = Prebid.shared.externalUserIdArray
+        }else if let externalUserIds = Targeting.shared.externalUserIds, externalUserIds.count != 0{
+            externalUserIdArray = externalUserIds
+        }
         var transformedUserIdArray = [[AnyHashable: Any]]()
         for externaluserId in externalUserIdArray {
             var transformedeuidDic = [AnyHashable: Any]()
--- a/Source/Targeting.swift
+++ b/Source/Targeting.swift
@@ -181,6 +181,20 @@ import CoreLocation
         }
     }
     
+    // MARK: - External UserIds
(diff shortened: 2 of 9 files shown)


ppuviarasu (commented):
1. dont change the sample app code in this PR

Source/RequestBuilder.swift
```diff
@@ -475,7 +475,14 @@ class RequestBuilder: NSObject {
     }
     
     func getExternalUserIds() -> [[AnyHashable: Any]]? {
-        let externalUserIdArray : [ExternalUserId] = Prebid.shared.externalUserIdArray
+       
+        var externalUserIdArray = [ExternalUserId]()
+        if Prebid.shared.externalUserIdArray.count != 0 {
+            externalUserIdArray = Prebid.shared.externalUserIdArray
+        }
+        else if Targeting.shared.externalUserIds.count != 0{
```
ppuviarasu, line 483: why is the externalUserId in 2 places? Prebid.shared & Targeting.shared?

Source/Targeting.swift
```diff
@@ -181,6 +181,34 @@ import CoreLocation
         }
     }
     
+    // MARK: - External UserIds
```
ppuviarasu, line 184: why is this entire logic needed when you already have the externalUSerID setter & getter in Prebid.shared

avermaiOS: Prebid.shared contains an Array of External UserIds same as what we implemented in MobileSDK whereas Targeting.shared contains setter & getter specific to each External UserId same as the way we implemented in GDPR.
So both serve different purposes.

Source/RequestBuilder.swift
avermaiOS: ExternalUserId can be set either by Prebid.shared.externalUserIdArray API or by local storage option from Targeting.shared getter & setter.
And based on details here https://github.com/prebid/prebid-mobile-ios/issues/347 we'll give first priority to API and after that to local storage

Source/Targeting.swift
ppuviarasu: 👍

Source/Targeting.swift
```diff
@@ -181,6 +181,40 @@ import CoreLocation
...
+    var externalUserIds = [ExternalUserId]()
+    /**
+     * This method allows to save External User Id in the User Defaults
+     */
+    public func setExternalUserId(_ externalUserId: ExternalUserId) {
```
avohraa, line 189: @avermaiOS @ppuviarasu 
Shouldn't we allow over-writing the external user id for a given source?

Tests/FetchingLogictests/RequestBuilderTests.swift
```diff
@@ -331,6 +333,63 @@ class RequestBuilderTests: XCTestCase, CLLocationManagerDelegate {
 
     }
     
+    //MARK: - Targeting External UserIds UserDefault
+    func testPostDataWithTargetingExternalUserIds() throws {
```
avohraa, line 337: Please add a test for ExternalUserId, where it is initialised with source = identifier = atype = ext = nil

avohraa (changes requested):

Source/Targeting.swift
avermaiOS: Done

Source/Targeting.swift
avermaiOS: Yes, we should. Made the changes accordingly

Source/Targeting.swift
```diff
@@ -181,6 +181,34 @@ import CoreLocation
...
+    }
+    
+    public func getExternalUserId(_ source : String)->ExternalUserId? {
+        guard let array = StorageUtils.getExternalUserIds(), let externalUserId = array.first(where: {$0.source == source}) else{
+            return nil
+        }
+        return externalUserId
+    }
+    
+    public func resetExternalUserIds() {
```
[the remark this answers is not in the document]
avermaiOS: Done

PR #372, opened 2021-04-04
Enabling Prebid Carthage integration for PrebidMobile framework without the script

rakesthedon:
Hi!

This will in my opinion make the use of carthage more inline with how it is suppose to work.

Advantages:
- Simplify carthage usage
- ~~No need for a script~~
- Better CI for people using carthage

Disadvantages:
- Both framework will be build even if only one is needed.

Issue #368, opened 2021-04-02
Rearchitect multi module structure


yoalex5:
**History:**
First iteration was implemented in #239

![PrebidMobile module v2](https://user-images.githubusercontent.com/8141931/115599110-df25bb80-a2e3-11eb-8b97-598d62e26741.png)

**Current situation:**
We have two targets(PrebidMobile and PrebidMobileCore) which use the same product name and module name(PrebidMobile). At the moment these two modules are the same, they include the same source files. We considered PrebidMobile as a full SDK

Usage:
```
import PrebidMobile
```

Advantages:
- `import PrebidMobile` is used for all products

Disadvantages:
- every source  file should be manually added in every target where it is used, as a result it simple to miss some file or add unnecessary file. It is hard to find this mistake
- hard to support new modules. When new module is creating you should add necessary files(at least from Core module) manually
- is not supported by Carthage(we use custom script as a workaround #277)
- is not supported by SPM

**Proposition:**

1. Use more native approach
2. Set a unique product name and module name for each product:
    - PrebidMobile instead of PrebidMobileCore. It will allows us to seamless migration because it doesn't cause breaking changes. All publishers which now use `PrebidMobileCore` or `PrebidMobile` don't need to replace module names in source files
    - PrebidMobileFull instead of PrebidMobile. PrebidMobileFull will not include any source files it will be just a name of dependency which will be responsible to add all sub-dependencies 

3. Set dependencies between them
4. Edit final distribution binaries(FatFramework, CocoaPods, Carthage, SPM, cross-project dependency)

Usage:
```
import PrebidMobile
import PrebidMobileRendering
```

@bszekely1 , @YuriyVelichkoOpenX what do you think ?

YuriyVelichkoOpenX:
I totally support the proposed approach. Each module should consist unique set of files and have dependencies on other modules if needed. 

I would like to suggest to review the structure of the Smaatto SDK and use the same approach:
https://developers.smaato.com/publishers/nextgen-sdk-ios-integration/

OpenWrap has the similar one in the scope of event handlers:
https://community.pubmatic.com/display/IDFP

So in the result I would like to see such dependencies tree:

PrebidMobile (includes all ad formats except event handlers)
PrebidMobileBanner
PrebidMobileInterstitial
PrebidMobileNative
PrebidMobileInstream
.....

PrebidMobileEventHandlerGAM
PrebidMobileEventHandlerMoPub

[...]

PR #374, opened 2021-04-05
Enabling Prebid Carthage integration for PrebidMobile framework

rakesthedon:
Hi!

This will in my opinion make the use of carthage more inline with how it is suppose to work.

Advantages:
- Simplify carthage usage
- Better CI for people using carthage
Disadvantages:
- Both framework will be build even if only one is needed.


yoalex5:
Hello @rakesthedon 
I still see a simple framework

Screenshot 2021-04-06 at 11 20 11

rakesthedon:
since both framework have the same name when they are built one is overriding the other, that can be fix if we change the name of the product name of PrebidMobileCore to PrebidMobileCore.framework

yoalex5:
I have already mentioned this behaviour in
https://github.com/prebid/prebid-mobile-ios/pull/372#issuecomment-813392180

Technically product name should be the same as module name. That is why you can not have different product names with the same module name

rakesthedon:
My bad i didn't see that part, but may i ask why? I don't think this a common convention. What was the decision factor for this decision?

yoalex5:
Main reason was to reusing modules. Publisher can switch(add/remove) between different modules without updating class files

Now we are discussing a possibility to rearchitect modules #368

rakesthedon:
I see. Then may be it is too early for this PR then.

PR #378, opened 2021-04-13
Disallow cookies in HTTP when deviceAccessConsent=false

Kowshickkarthick:
## Issue
If Prebid SDK doesnot have device access consent(as per GDPR/TCF2.0), then it should not be reading/writing cookies while making HTTP requests.

## Description of Fix
Check for deviceAccessConsent and disallow cookies in the request deviceAccessConsent=false. 
Similar check already exists in Android. This is just a mirror of what is already there in Prebid Android


yoalex5:
@Kowshickkarthick, please review #381

ppuviarasu (approved):

yoalex5 (approved):

Issue #377, opened 2021-04-09
Align PrebidMobile and PrebidMobileCore Compiled sources 


yoalex5:
PrebidMobile and PrebidMobileCore different amount of classes but they should be equal

rakesthedon:
Does this mean that **PrebidMobile** and **PrebidMobileCore** are now 2 different frameworks?

Issue #369, opened 2021-04-02
Migrate Universal Framework to XCFramework


yoalex5:
We support creating fat framework which is used by publishers

Disadvantages:
- Memory Footprint. It contains several architectures(e.g. arm and x86) which are included into final(app) binary

Proposition:
Replace fat framework by XCFramework

Issue #345, opened 2021-01-12
Add Swift Package Manager support


tahirmt:
**Is your feature request related to a problem? Please describe.**
Swift package manager is the officially supported package manager and we should add support for it.

bszekely1:
Hi @tahirmt, we will place this item in the roadmap for Prebid SDK.

tahirmt:
@bszekely1 thank you

PR #384, opened 2021-04-27
Rearchitect multi module structure

yoalex5:
main changes:

- Rearchitect multi module structure (#368)
- XCFramework is used instead of UniversalFramework(#369)
- Swift Package Manager(SPM) support(#345)
- Carthage support without extra scripts
- Stared test source files were moved into a separate project - `TestUtils`
- all targets of PrebidMobile project are located in the root folder
- minor changes(#377)


YuriyVelichkoOpenX (commented):
Hi everyone! I did not have much time today for a deep review, so I'm still in progress with this.

However, I would like to clarify that after this PR we will get one module for all integration kinds instead of improving the granularity of modules. Therefore publishers who want to integrate the prebid SDK only for in-stream video will get all other ad formats as well in the framework. Is it an intended future behavior or just an intermediated step of the modularization initiative? 

Which modules are you going to get in the result of modularization?

yoalex5:
> I would like to clarify that after this PR we will get one module for all integration kinds instead of improving the granularity of modules. 

The main target of this PR is to get rid of using single module name for all modules in future

> Therefore publishers who want to integrate the prebid SDK only for in-stream video will get all other ad formats as well in the framework.

Current module includes all ad formats (Banner, Interstitial, Video) inside corresponding `AdUnit` successors (BannerAdUnit, InterstitialAdUnit, VideoAdUnit...) which allows publishers to make auction request in a simple manner. In my opinion all these classes solve a single task - provide publisher a winning bid. I don't think that this module should be divided into smaller modules because these classes are not heavy and do not have others dependencies

>  Is it an intended future behavior or just an intermediated step of the modularization initiative?

It is expected behaviour. At least for now

bszekely1:
It is for this reason we would need a module overview page and a description of the module architecture. By default Prebid SDK has minimal logic. As we add more complex functionality and additional weight as Oleksandr has pointed out, we can break them out as modules.

YuriyVelichkoOpenX:
@yoalex5 @bszekely1 thanks for the clarification. Everything sounds reasonable. And yes we definitely need a precise overview of modularisation.

YuriyVelichkoOpenX (dismissed):
LGTM

yoalex5:
@YuriyVelichkoOpenX, could you please check my latest commits. 
I added `TestUtils` module

YuriyVelichkoOpenX:
@yoalex5, the new module looks good!

However, I can't run unit tests nor from Xcode nor from script:

image

image

yoalex5:
@YuriyVelichkoOpenX 

Could you please remove `Derived Data` folder and build a test target ?
CI does not have such problem

YuriyVelichkoOpenX:
@yoalex5 , cleaning of Derived Data didn't help but rebooting helped. 

Everything works!

YuriyVelichkoOpenX (approved):
LGTM

Issue #393, opened 2021-07-06
Introduce the consistent refresh policy 


YuriyVelichkoOpenX:
# Motivation

There is a [pull request](https://github.com/prebid/prebid-mobile-android/pull/267) that highlights the difference in the docs and implementation respectively to the management of ad unit refreshing.

The issue is absolutely relevant but the proposed fix is not applied to the iOS and still has differences with documentation. 

This request is intended to provide the general spec for the refresh policy on both platforms and the steps to apply it.

# Policy

This policy is based on the current behavior and takes into account the necessity of the start/stop feature. It is not intended to change the API even if it would make sense to make the refresh more convenient. Eventually, on merging with the rendering module, the refresh policy will be revisited.

- By default, the autorefresh is off for all ad units kinds. 
- To activate the autorefresh the publisher should set the refresh delay using the method `setAutoRefreshMillis`.
- The minimum refresh interval is `30 sec` or `30000 ms`. The interval less than minimal will be ignored and autorefresh won't run. 
- Publisher should set the refresh interval before calling the `fetchDemand` method. The refresh timer starts when the application calls the `fetchDemand` method, without waiting for the response.
- Calling of the `setAutoRefreshMillis` after the `fetchDemand` stops the current refresh cycle and starts the new one respectively to the start/stop policy described below. 
- The current implementation can't stop and resume refresh respectively to the visibility of the ad view.
- In order to stop autorefresh publishers should call the `stopAutoRefresh()`. 
- In order to resume autorefresh publisher should call the `resumeAutoRefresh()` method. Calling of `resumeAutoRefresh()` is ignored if autorefresh has not been stopped before with `stopAutoRefresh()`. 
- `stopAutoRefresh()` tracks the reminded time for the current refresh period.
- `resumeAutoRefresh()` starts the refresh timer respectively to the remined time of the stoped cycle.
- The `setAutoRefreshMillis` won't resume the stopped refresh. 

# Objectives

## iOS 

- Introduce the `resumeAutoRefresh()` method satisfying the above policy.
- Update the stop/resume logic respectively to the above policy. Implementation details should be similar to the current implementation on Android.

## Android

- Introduce the `resumeAutoRefresh()` method satisfying the above policy.

## Tests

- All changes should be covered with unit tests.

## Docs 

- Update the public docs with the current policy.

PR #398, opened 2021-08-02
Implement start/stop/resume autorefresh functionality

VadimKhohlovOpenX:
This PR introduces support of start/stop/resume autorefresh functionality according to the description:
https://github.com/prebid/prebid-mobile-ios/issues/393

- added code to support start/stop/autorefresh
- updated the Swift demo app for demonstration the introduced functionality for the banner sample
- updated unit-tests


YuriyVelichkoPI (approved):
LGTM

maddapper:
My only suggestion is to try using a DispatchSourceTimer instead of the Timer class, since I think it is more CPU efficient.

PR #401, opened 2021-10-18
Add OpenRTB app:content:url property into BidRequest

marcelklinek:
This PR aims to provide more precise targeting information with bid request. When PBS asks for bids for an ad slot, bidders have very little information about the user. This is caused by latest developments where users choose to opt-out of sharing their IDFA and such.

We can at least supply bidders with some context, like what content the user is looking at when they see an ad. Google ad manager calls this technique content mapping: https://support.google.com/admanager/answer/11050896?hl=en

Publishers roughly map views in their apps to pages on their websites. Then the targeting information already collected for pages on publisher’s website can be used to better target ad slots that appear in the apps. For this to work bidders needs to receive information which page would this ad slot appear on if this was a website and not an app. This page’s URL is set into ContentUrl property. Google’s ad manager has already implemented this property: https://developers.google.com/ad-manager/mobile-ads-sdk/ios/targeting#content_url

Our Prebid server is connected to a couple DSPs that are able to target their bids based on contentUrl. We identified an OpenRTB property that could carry this data. See Bid request -> App -> Content -> Url in https://www.iab.com/wp-content/uploads/2016/03/OpenRTB-API-Specification-Version-2-5-FINAL.pdf

Unfortunately we could not find a way to pass this property to the PBS using the SDK. This PR is an attempt to fix it. We would like to be able to pass this from the SDK because it would allow to re-use the same ad slot on different views with different content within an app.

### DIFF
--- a/PrebidMobile/AdUnits/AdUnit.swift
+++ b/PrebidMobile/AdUnits/AdUnit.swift
@@ -16,6 +16,8 @@ import ObjectiveC.runtime
 @objcMembers public class AdUnit: NSObject, DispatcherDelegate {
 
     public var pbAdSlot: String? = nil
+    
+    public var content: ContentType?
 
     private static let PB_MIN_RefreshTime = 30000.0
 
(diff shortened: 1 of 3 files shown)


marcelklinek:
Sorry, forked this one from wrong account. Please see https://github.com/prebid/prebid-mobile-ios/pull/403

PR #446, opened 2021-11-19
Fix MoPub examples

OlenaPostindustria:
## Description of changes
- Update delegate methods in MoPub adapters;
- update delegate methods in Internal Test App;
- show button status depends on MPRewardedAds.hasAdAvailable now.


YuriyVelichkoPI (changes requested):
I don't see the renaming of UI labels for respective buttons in the Internal Test App. The storyboard file should be changed.

YuriyVelichkoPI:
@MaksPetrukhnoPI, @ValentinPostindustria  what about Android?
- are listeners work correct? 
- wich MoPub API is used? Deprecated or a new one? 

We should have a consistent implementation on both platforms.

YuriyVelichkoPI (approved):
LGTM

PR #448, opened 2021-11-22
Changes: Video Outstream feed 

MaksPetrukhnoPI:
- Changed Video outstream feed list size from 10000 to 4 (for tests to work correctly)
- Added placement parameter to Video Outstream Feed (GAM)


YuriyVelichkoPI (approved):
LGTM

PR #468, opened 2021-12-08
Support app: content object

OlenaPostindustria:
- Added support of app: content object into original sdk; 
- Added support of app: content object into rendering part;
- Unit tests for app: content object were written.


YuriyVelichkoPI (approved):
LGTM

PR #481, opened 2021-12-21
Refactor Mediation API

OlenaPostindustria:
## Description of changes 
- MediationDelegate protocol has been changed;
- get rid of adObject in fetchDemand call, we inject it to MediationDelegate object now;
- MoPub Mediation Utils were changed - there are separate Mediation Utils for each type of ad(banner, interstitial, rewarded);
- Unit Tests have been changed.


YuriyVelichkoPI (approved):
LGTM

PR #482, opened 2021-12-23
[AdMob] Introduce the Banner Adapter

OlenaPostindustria:


EventHandlers/PrebidMobileAdMobAdapters/PrebidAdMobBannerAdapter.swift
```diff
@@ -0,0 +1,124 @@
...
+            return
+        }
+        
+        guard let keywords = request.userKeywords as? [String] else {
+            let error = AdMobAdaptersError.emptyCustomEventExtras
+            delegate?.customEventBanner(self, didFailAd: error)
+            return
+        }
+        
+        for parameter in jsonServerParameters {
```
YuriyVelichkoPI, line 74: This part is a cornerstone of the AdMob adapter. We should move it into the shared method since it will be used in all adapters, and we should test it as best as possible.

EventHandlers/PrebidMobileAdMobAdapters/PrebidAdMobBannerAdapter.swift
```diff
@@ -0,0 +1,124 @@
...
+            guard keywords.contains(parameter) else {
```
YuriyVelichkoPI, line 75: We should introduce a more reliable way for matching the targeting keywords and the Ad Source parameters. Verification of the whole string makes as extremely format-sensitive. The single extra space inside the JSON structure will break the result. 
Need to compare string literals of the keys and values separately. And provide respective unit tests.
We should introduce a more reliable way for matching the targeting keywords and the Ad Source parameters. Verification of the whole string makes as extremely format-sensitive. The single extra space inside the JSON structure will break the result. 
Need to compare string literals of the keys and values separately. And provide respective unit tests.

EventHandlers/PrebidMobileAdMobAdapters/PrebidAdMobBannerAdapter.swift
```diff
@@ -0,0 +1,124 @@
...
+                let error = AdMobAdaptersError.emptyCustomEventExtras
```
YuriyVelichkoPI, line 76: The error is irrelevant, the extras are not empty. They just do not match the Ad Source params.

Example/PrebidDemo/PrebidDemoSwift/BannerController.swift
```diff
@@ -333,6 +352,28 @@ class BannerController:
...
+        
+        let size = CGSize(width: 320, height: 50)
+        
+        mediationDelegate = AdMobMediationBannerUtils(gadRequest: gadRequest, bannerView: gadBanner)
+        prebidAdMobMediaitonAdUnit = MediationBannerAdUnit(configID: "50699c03-0910-477c-b4a4-911dbe2b9d42", size: size, mediationDelegate: mediationDelegate)
+        
+        prebidAdMobMediaitonAdUnit.fetchDemand { [weak self] result in
+            let extras = GADCustomEventExtras()
+            let prebidExtras = self?.mediationDelegate.getEventExtras()
+            extras.setExtras(prebidExtras, forLabel: "PrebidAdMobCustomEvent")
```
YuriyVelichkoPI, line 371: Need to introduce the constant for "PrebidAdMobCustomEvent". And let's rename it to "PrebidAdMobCustomEventExtras" since it serves as a key for prebid's extras data.

YuriyVelichkoPI (changes requested):

YuriyVelichkoPI (changes requested):

YuriyVelichkoPI (approved):
LGTM

PR #501, opened 2022-01-27
[AdMob] Add video interstitial cases to the test apps

OlenaPostindustria:
- Added video interstitial cases to Internal Test App
- Video interstitial cases was added to PrebidDemo apps earlier


YuriyVelichkoPI (approved):
LGTM

PR #508, opened 2022-01-31
Mediation API: enrich the keywords and extras of third party libs in more gentle way.

OlenaPostindustria:
- rebuild ad object set up logic;
- add unit tests;
- fix AdMob adaptive banner sizes

PR #509, opened 2022-01-31
AdMob Mediation Utils: optimize the noBids case 

OlenaPostindustria:
- moved the checking of targeting keywords to the first place in the adapter's methods requestAd;
- updated keywords set up logic in #508


YuriyVelichkoPI (approved):
LGTM

PR #519, opened 2022-02-08
Remove Native Ads components from the Rendering API

### LINKED ISSUE #476: Remove Native Ads components from the Rendering API
## Motivation

According to the product requirements for 1.14 we do not need Native functionality from the rendering API anymore. So we should remove it from the codebase. 

## Prerequisites 

1. #478
2. #480
2. #479

## Objectives

Remove the code for Native Ads from the rendering codebase. 

**Note**: it is not clear from the first glance what we should do with each inconsistent case during replacement. So let's discuss and make decisions online during implementation.

OlenaPostindustria:
Closes #476


YuriyVelichkoPI (approved):
LGTM

PR #532, opened 2022-02-18
Fix bundle short version

### LINKED ISSUE #514: Info.plist bundle short version is wrong
**Describe the bug**
The SDK version which is send alongside a request for bids always send as SDK version the text "1.0" This is not true since we already have a version 1.12.1 as stable available.

**To Reproduce**
1. Build a test app using latest Prebid Mobile SDK
2. Run the app and trigger a bid request
3. Check the request which is send. The JSON object under **app.ext.prebid** contains two entries. The source and version.

**Expected behavior**
A new SDK should also send this detail within its JSON request.

**Screenshots**
image

 - Device: iPhone 7
 - OS: 15.3
 - Native App
 - SDK Version 1.12.1

OlenaPostindustria:
Closes #514

### DIFF
--- a/PrebidMobile/Constants.swift
+++ b/PrebidMobile/Constants.swift
@@ -16,6 +16,7 @@
 import UIKit
 
 public let PrebidLocalCacheIdKey = "hb_cache_id_local"
+public let PREBID_VERSION = "1.12.1"
 
 extension String {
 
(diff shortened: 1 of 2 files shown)


YuriyVelichkoPI (approved):
LGTM

ThomWee:
Thanks for fixing this. Just wondering is the version string update a manual step or do you use a tool updating the version info in both files now?

PR #547, opened 2022-02-23
First party data fixes

OlenaPostindustria:
- remove `Object` suffix from naming;
- revert changes to user.ext.data and app.ext.data;
- fix process argument parsing;
- change process argument names;
- fix user.ext.data setup bug in rendering.


YuriyVelichkoPI (approved):
LGTM

PR #560, opened 2022-03-15
[CODE UNIFICATION] Use bid requesting engine from the rendering codebase in the original API

OlenaPostindustria:


YuriyVelichkoPI (approved):
LGTM

PR #593, opened 2022-03-30
Migration: Display interstitial from mock server to pbs on aws

MaksPetrukhnoPI:


InternalTestApp/PrebidMobileDemoRendering/Model/TestCasesManager.swift
```diff
@@ -807,107 +807,110 @@ struct TestCaseManager {
...
                     return
                 }
                 let interstitialController = PrebidInterstitialController(rootController: adapterVC)
-                if AppConfiguration.shared.useMockServer {
                     interstitialController.prebidConfigId = "imp-prebid-no-bids"
-                } else {
-                    Prebid.shared.accountID = "1768035c-74d3-4786-b056-13bd41f34bde"
-                    interstitialController.prebidConfigId = "28259226-68de-49f8-88d6-f0f2fab846e3"
-                }
+                         Prebid.shared.storedAuctionResponse="response-prebid-no-bids"
```
YuriyVelichkoPI, line 835: We should pass storedAuctionResponse into respective controllers. 
The controller should use this property during its own lifecycle and clear it in the end. So the global values won't affect other cases.

PR #606, opened 2022-04-05
Send nested targeting params from the original API in the Bid Requests

### LINKED ISSUE #594: Send nested targeting params from the original API in the Bid Requests
## Background

Original API objects have targeting properties introduced in the nested `Parameters` classes. 

## Objectives

Review original API classes and make sure that all properties from nested `Parameters`  classes appear in the bid requests.

OlenaPostindustria:
Closes #594 

- use original banner and video parameters during building bid request; 
- use original banner and video parameters in rendering adUnits;
- replace maxVideoDuration and placementType with similar properties from original video parameters;
- remove default values for maxDuration and placement video properties;
- add & update unit tests;
- delete unnecessary In-App purchases from Internal Test App.


YuriyVelichkoPI (approved):
LGTM

PR #624, opened 2022-04-26
Restore and migrate demo apps to PBS on AWS

OlenaPostindustria:
- restore original API broken cases in Objective C app;
- restore native original API cases in Swift Demo;
- migrate all cases in Objective C app to PBS on AWS;
- add missing cases for banner video;
- add MAX cases


YuriyVelichkoPI (approved):
LGTM

PR #626, opened 2022-04-28
Restore parameters classes in original API & corrections

OlenaPostindustria:
- restore Parameters classes in BannerBaseAdUnit and VideoBaseAdUnit and mark it deprecated;
- add storedResponse for original GAM video interstitial integration;
- add storedResponse for original GAM banner video integration.


YuriyVelichkoPI (approved):
LGTM

PR #629, opened 2022-05-03
Fix video click events & corrections

OlenaPostindustria:
- fix video click events;
- add more SKAdN 2.2 examples;
- set auto display on load for interstitial ads to false to prevent showing ad without explicitly calling `show` method.


ValentinPostindustria (approved):
LGTM

ValentinPostindustria (approved):
LGTM

PR #630, opened 2022-05-05
Fix creativeDidComplete missing call in rewarded ad with end card

### LINKED ISSUE #450: Rewarded Adapter - rewardedAdShouldReward is not called for the ads with End Card
## Description

For the ad with companion creative, the delegate method `rewardedAdShouldReward` is not called due to the intricate flow of delegate calls. 

The start point for investigation should be a `FIXME` comment in `modalManagerDidFinishPop` method in the `PBMHTMLCreative`.  

## Objective

Refine the flow of the ad state signaling (delegate calls) in order to make it more convenient and fix the current behavior.

OlenaPostindustria:
Closes #450


YuriyVelichkoPI (approved):
LGTM

PR #662, opened 2022-06-03
Remove support of calendar and storePicture features of MRAID

### LINKED ISSUE #656: App Store Connect binary upload fails because of NSCalendarsUsageDescription
**Describe the bug**
After updating the `PrebidMobile` sdk from version `1.12.2` to `2.0.0` we were not able anymore to successfully upload a binary to App Store Connect. The binary gets rejected with the following email from Apple:

> We identified one or more issues with a recent delivery for your app. Please correct the following issues, then upload again.
>
> **ITMS-90683: Missing Purpose String in Info.plist** - Your app's code references one or more APIs that access sensitive user data. The app's Info.plist file should contain a NSCalendarsUsageDescription key with a user-facing purpose string explaining clearly and completely why your app needs the data. Starting Spring 2019, all apps submitted to the App Store that access user data are required to include a purpose string. If you're using external libraries or SDKs, they may reference APIs that require a purpose string. While your app might not use these APIs, a purpose string is still required. You can contact the developer of the library or SDK and request they release a version of their code that doesn't contain the APIs. Learn more (https://developer.apple.com/documentation/uikit/core_app/protecting_the_user_s_privacy).

**To Reproduce**
1. Use the `PrebidMobile` sdk version `2.0.0` 
2. Upload a binary to App Store Connect

**Expected behavior**
The binary upload should not be affected by the `PrebidMobile` sdk. It should not be necessary to grant the calendar permission.

OlenaPostindustria:
Closes #656


YuriyVelichkoPI (approved):
LGTM

PR #671, opened 2022-06-30
Add missing xcassets resources import on .podspec

jovencinho:
Without this, the ads do not show icons like close, skip, sound, etc.


OlenaPostindustria (approved):
LGTM

PR #675, opened 2022-07-15
Fix Content-Type value check 

OlenaPostindustria:


YuriyVelichkoPI (changes requested):
we need a unit test here in order to avoid regression in future

Kowshickkarthick (approved):
Thanks for making the changes.
The changes look good.
Approved :)

YuriyVelichkoPI (approved):
LGTM

PR #679, opened 2022-08-15
Fix not clickable video rewarded

### LINKED ISSUE #678: Not clickable Ad (video rewarded)
When this [change](https://github.com/prebid/prebid-mobile-ios/blob/3c96799b7941c56ac28682792ad80192cb2eab99/PrebidMobile/PrebidMobileRendering/AdTypes/AdView/PBMVideoView.m#L217-L221) was added

```
    // For rewarded ad learn more should be hidden
    if (self.creative.creativeModel.adConfiguration.isOptIn) {
        self.showLearnMore = NO;
        return;
    }
```
The possibility for the user to click on the ad was removed at all, if you don't have the `learn more` button when the video is playing and the behavior of a video reward without end card is to close automatically

It can be produced using this example rewarded video
simulator_screenshot_D2FDC6A9-FFC4-456B-9EB6-EA780BE7228B

OlenaPostindustria:
Closes #678


YuriyVelichkoPI (approved):
LGTM

PR #683, opened 2022-09-13
fix(render): automatic close of video creative is avoided when it does not have a companion ad

jovencinho:


YuriyVelichkoPI:
Let's keep it on hold. Motivation in the comments for similar changes: #684

PR #684, opened 2022-09-20
fix: automatic pop to root of the interstitial after clickthroughs is prevented

jovencinho:
The behavior is changed so that when a fullscreen ad opens the browser after clicking on it and after closing the browser, the fullscreen ad does not auto-close


YuriyVelichkoPI:
Hi @jovencinho ! Thanks for the update. 

The behavior of the rendering engine is not strictly defined in the community yet. I would like to commit to such changes only after committee approval.  

But the first step should be capturing the current behavior: #693. Only after can we change it and apply it on both platforms. Otherwise, we'll add a lot of inconsistency and mess into UX. Also, we can provide behavior flags so publishers can customize actions.

I'm going to discuss UI/UX polices on the next PMC call. I'll keep you posted.

PR #700, opened 2022-11-01
Add support of the latest GMA SDK in the Original API

### LINKED ISSUE #691: Support the latest GMA SDK in the Original API
## Objectives

- Ensure that all methods and classes that are used in the original API via reflection are present in the latest version of the GMA SDK
- If not, update the behavior respectively to the changes in the GMA SDK
- Add a check method to the `initializeSDK` , which verifies the GMA SDK version. 
  - If the version is unsupported - print a respective error message into the log "The current version of GMA SDK is not supported! No ads from Prebid will be shown. Please contact the Prebid Mobile committee for the details". The unsupported version means that it doesn't have classes or methods that are used in the reflection calls. 
  - If the GAM SDK version is higher than the last supported one - print the warning message: "The current version of Prebid SDK is not validated with the latest version of GMA SDK. Please update the Prebid SDK or post a ticket on the github. "

OlenaPostindustria:
Closes #691


YuriyVelichkoPI (approved):
LGTM

PR #722, opened 2022-11-22
Corrections for TCF & CCPA

### LINKED ISSUE #715: Corrections for TCF & CCPA
### TCF

1. Follow the "fallback for iab" approach. The method `Targeting.purposeConsents` should check the prebid (api-based) consent first, and iab consent next. 
2. Remove code duplication in `PBMUserConsentResolver` and `Targeting.getDeviceAccessConsent` and related methods. Update tests, do not loose functionality. 
3. Remove duplication in `Targeting` and `PBMUserConsentDataManager`. Use the single class. 
4. API-provided `Subject` and `Consent` shouldn't be placed into persistent storage. In this case, we have a risk that the publisher may stop using API for the TCF and start using CMP SDK. As a result, the old data from the persistent storage will signal to the Prebid SDK that there is API-provided consent. And CMP-provided consent won't be used.  
5. Drop support of TCF v1.1
6. Follow this [doc](https://docs.prebid.org/prebid-mobile/privacy-regulation.html#sending-device-information) for **reading** IDFA. Make the reading of `advertisingIdentifier`  inside `PBMDeviceAccessManager` explicitly depend on the user's consent.

### CCPA Compliance

1. Remove duplication in `PBMUserConsentDataManager` and `StorageUtils` for processing CCPA content.

OlenaPostindustria:
Closes #715


YuriyVelichkoPI (dismissed):
LGTM

YuriyVelichkoPI (approved):
LGTM

PR #728, opened 2022-11-24
Support different versions of xcode

hugolm84:
Hi!

OS X Ventura and Xcode>=14 have made some changes that requires some updates in the build scripts.

For example, Xcode 14 only supports iOS >11 and BITCODE is deprecated, thus, `find` will fail and the build script will error out.

These changes will allow users to build against xcode14 and xcode13. This can of course be expanded to support versions even further back but this seems to be a good start.

I added `runSwift.sh` for building and launching the PrebidSwiftDemo application in a simulator of choice.

To build prebid-mobile-ios and launch the swift application, first build the project:

`./script/buildPrebidMobile.sh`

then build and launch the PrebidSwiftDemo

`./script/runSwift.sh "iPhone 14 Pro Max"`

**Note:**
I have limited experience with iOS and Xcode, let me know if something does not make sense.

Some of the projects are not listed in he Podfile and thus cannot be modified to in the `post-install` step. These project needs to be added to the Podfile or users need to update them manually. Warnings that would eventually lead to errors are:

```
prebid-mobile-ios/EventHandlers/EventHandlers.xcodeproj: warning: The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 10.0, but the range of supported deployment target versions is 11.0 to 16.1.99. (in target 'PrebidMobileGAMEventHandlers' from project 'EventHandlers')
```

For Xcode 14, `IPHONEOS_DEPLOYMENT_TARGET` should be set to `12`.

My knowledge with Podfiles are limited, so I am not sure what would be the best way to include these project files in the post-install step.

Update these files:

```
EventHandlers/EventHandlers.xcodeproj/project.pbxproj
PrebidMobile.xcodeproj/project.pbxproj

```

with

```
-                               IPHONEOS_DEPLOYMENT_TARGET = 10.0;
+                               IPHONEOS_DEPLOYMENT_TARGET = 12;
```

To fix all warnings, also change these:

```
-                               CLANG_WARN_QUOTED_INCLUDE_IN_FRAMEWORK_HEADER = YES;
+                               CLANG_WARN_QUOTED_INCLUDE_IN_FRAMEWORK_HEADER = '$(inherited)';
```
 

image

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -1,5 +1,14 @@
-platform :ios, '12.0'
+$xcode_version = %x[xcrun xcodebuild -version | head -1 | awk '{print $2}']
+#'
+# ^ %x '' conflict with syntax highlight
+# Xcode 14: targets iOS 11-
+# Xcode 12,13: targets iOS 9-
+$iosVersion = '12.0'
+if Gem::Version.new($xcode_version) < Gem::Version.new('14')
+  $iosVersion = '10.0'
+end
 
+platform :ios, $iosVersion
 workspace 'PrebidMobile'
 
 project 'PrebidMobile.xcodeproj'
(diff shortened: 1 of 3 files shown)


YuriyVelichkoPI:
Hi @hugolm84! Thanks for escalating the Xcode compatibility case.

Please, describe the problems that you faced in the GitHub issue first. Some of the changes that you made in the PR have been applied or will be introduced in the scope of other tasks. 

Also, we should keep the repo generic for all users and as much dry as possible, so, for example, the script `runSwift` is an additional dependency that should be supported by the development team eventually. But it doesn't play a significant role in the building or testing of the SDK. On the other hand, we also have an ObjC demo, so we need the respective script to run the app. 

In order to apply patch changes, let's keep the flow: 

GitHub issue -> community discussion and development of requirements -> PR with respective changes. 

@OlenaPostindustria, which version of the Xcode do we use to build and release v 2.0.5?

OlenaPostindustria:
@YuriyVelichkoPI, we use Xcode 14.1

hugolm84:
Hi @YuriyVelichkoPI and @OlenaPostindustria , thanks for your reply!

Since yesterdays updates on master, I can now build and run demos without `post-install` steps and xcode version checks. I still experience some issues and open #729 to follow the flow.

PR #741, opened 2022-12-06
Setup ad unit parameters in internal test app for original api according to the docs

MaksPetrukhnoPI:


OlenaPostindustria (approved):
LGTM

PR #750, opened 2022-12-16
Cleanup original bid requests

### LINKED ISSUE #742: Cleanup original bid requests
## Motivation

Do not send properties irrelevant to the Original API integration with GAM.

## Objectives

- Do not add skadn object to the bid response since we are not able to handle it with the original API
- Do not send internal omid signals. They should be set by publishers respectively to their GMA SDK version and setup. 
- Do not send "yob" if it is not set by the publisher. 
- Do not send empty objects
- the value for `clickbrowser` should be 1 for the original API
- `displaymanager` and `displaymanagerver` should be assigned to respective properties of GMA SDK (need to discuss how to do it in the proper way)
- review the `imp.id` implementation in the original SDK. The standard says `A unique identifier for this impression within the context of the bid request (typically, starts with 1 and increments). But we use UUID.
- review the `secure` flag. I'm not sure that it should be 1. What value is used in the 1.2?
- don't set `omidpn` and `omidpv` unless thy are provided via Targeting API
- drop default value for `linearity`
- review the usage of `playbackend` in 1.12. If it was absent - do not send it for the Original API 
- review the usage of `pos` in 1.12. If it was absent - do not send it for the Original API 
- For video ads don't send any default values if a publisher hasn't provided `VideoParameters.` Let's send only `w`, `h`, `delivery`, and `placement` because it is basically all that we know about default integration.

OlenaPostindustria:
Closes #742


PrebidMobile/ConfigurationAndTargeting/PrebidInternal.h
```diff
@@ -0,0 +1,28 @@
...
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+  */
+
+// TODO: Merge with Prebid class after migrating codebase to Swift
+
+/// Internal class
+/// isOriginalAPI - true if it is used Original API, false if Rendering
+@interface PrebidInternal : NSObject
```
YuriyVelichkoPI, line 20: This approach may fail if the publisher use mixed integration scenarious. For instance if he decided to experiment and added one pure In-App banner to the app. In this case the global settings will break the behaviour. 

Can we use `AdUnitConfig` (add property to this class) to indicate the integration kind on the Ad Unit level?

YuriyVelichkoPI (changes requested):

PrebidMobile/ConfigurationAndTargeting/PrebidInternal.h
OlenaPostindustria: Sure, I'll make appropriate changes

PR #776, opened 2023-01-26
Remove hb_size from winning bid markers

### LINKED ISSUE #775: Do not treat `hb_size` as a marker of a winning bid
## Motivation

The size property doesn't make sense for native ads, and an rtb response doesn't contain such a field. 
Also, looking at the PBS logic, we can see that hb_size can be missed for other ad formats as well. So PBS can skip adding the `hb_size` to the targeting keywords of the winning bid. 

## Objectives

- Remove `hb_size` from the list of markers of the winning bid. 
- Update Unit tests
- Introduce respective cases in the Internal test app

OlenaPostindustria:
Closes #775

### DIFF
--- a/PrebidMobile/PrebidMobileRendering/Prebid/PBMCore/Bid.swift
+++ b/PrebidMobile/PrebidMobileRendering/Prebid/PBMCore/Bid.swift
@@ -82,7 +82,7 @@ public class Bid: NSObject {
             return false
         }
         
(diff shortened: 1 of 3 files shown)


YuriyVelichkoPI (approved):
LGTM

YuriyVelichkoPI (approved):
LGTM

PR #787, opened 2023-02-07
Update SDK reliance on PBS /status endpoint

### LINKED ISSUE #786: Update SDK reliance on PBS /status endpoint
## Motivation

Look at https://github.com/prebid/prebid-mobile-android/issues/499

## Objectives

- Introduce a new public property in the Prebid object - `customStatusEndpoint`
    - SDK expects to receive the valid URL for the /status endpoint of PBS. Not only the path but the whole URL. F.e.: http://prebid.adnxs.com/pbs/v1/status
- During the SDK initialization, the given property should be used to check the PBS status.
- If the publisher didn't provide `customStatusEndpoint,` use the current approach. 

Update the process of checking server status:
- Introduce the new case for `PrebidInitializationStatus` - `serverStatusWarning`.
- In the `checkServerStatus` function, change the reporting of status checking errors - use the `.serverStatusWarning` for all "not succeed" cases. 
- Fix the typo in the `PrebidInitializationStatus` case `successed` -> `succeeded`
- Change the definition of `succeeded` status. We should treat any of `2xx` return codes as _succeeded_.

- Update the description of the `initializeSDK` method. 
    - Describe the necessity of `customStatusEndpoint` and its usage. 
    - Describe the values of `PrebidInitializationStatus`, its meaning, and how SDK continues to work each status respectively.

OlenaPostindustria:
Closes #786

### DIFF
--- a/PrebidMobile/ConfigurationAndTargeting/PrebidServerStatusRequester.swift
+++ b/PrebidMobile/ConfigurationAndTargeting/PrebidServerStatusRequester.swift
@@ -35,7 +35,8 @@ class PrebidServerStatusRequester {
             if customStatusEndpoint.isValidURL() {
                 serverEndpoint = customStatusEndpoint
             } else {
-                Log.warn("The provided Prebid Server custom status endpoint is not valid. The default endpoint will be used.")
+                let endpointMessage = serverEndpoint == nil ? "There is no status endpoint to use." : "The '\(serverEndpoint ?? "")' endpoint will be used."
+                Log.warn("The provided Prebid Server custom status endpoint is not valid. \(endpointMessage)")
             }
         }
     }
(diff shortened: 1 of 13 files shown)


PrebidMobile/ConfigurationAndTargeting/PrebidServerStatusRequester.swift
```diff
@@ -0,0 +1,60 @@
...
+    // MARK: - Internal Methods
+    
+    func requestStatus(_ completion: @escaping PrebidInitializationCallback) {
+        guard let serverEndpoint = serverEndpoint else {
+            completion(.serverStatusWarning, PBMError.error(description: "The SDK failed to get Prebid Server status endpoint."))
+            return
+        }
+        
+        ServerConnection.shared.get(serverEndpoint) { serverResponse in
+            guard (200...299).contains(serverResponse.statusCode) else {
```
YuriyVelichkoPI, line 52: Let's make this check the part of `ServerResponse` class so that we can reuse it like this:

`serverResponse.isOkStatusCode`

YuriyVelichkoPI (changes requested):

YuriyVelichkoPI (approved):
LGTM

PR #794, opened 2023-02-15
Deprecate context methods

### LINKED ISSUE #792: Deprecate `Context*` methods
## Motivation

SDK API enhancements #785

## Objectives

Since we moved data out from the `context` field it makes sense to rename the `Context*` methods and properties to `Ext*` to keep the codebase consistent with the RTB schema. 

let's do it by deprecating of `Context` methods and introducing the `Ext` methods. The `Context` should just call the respective `Ext` method inside. In this way, users will be aware of changes in the method's behavior.

OlenaPostindustria:
Closes #792

### DIFF
--- a/PrebidMobile/AdUnits/AdUnit.swift
+++ b/PrebidMobile/AdUnits/AdUnit.swift
@@ -183,76 +183,144 @@ import ObjectiveC.runtime
         }
     }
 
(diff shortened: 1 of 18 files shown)


YuriyVelichkoPI (approved):
LGTM

PR #796, opened 2023-03-06
[Rendering API] Migrate from internal clickthrough browser to SFSafariViewController

### LINKED ISSUE #793: [Rendering API] Migrate from internal clickthrough browser to `SFSafariViewController`
## Motivation

Prebid iOS SDK uses an internal browser based on the WebView to open the `clickURL` of the ad. Eventually, the user can switch to the external browser.  It is a legacy approach that is extremely outdated nowadays. 

Investigating the behavior of other SDKs, you can see that they use SFSafariViewController as a click browser.

## Objectives

Replace the internal clickthrough browser with [SFSafariViewController](https://developer.apple.com/documentation/safariservices/sfsafariviewcontroller).

OlenaPostindustria:
Closes #793


PrebidMobile/ConfigurationAndTargeting/Prebid.swift
```diff
@@ -110,7 +110,6 @@ public class Prebid: NSObject {
     public var creativeFactoryTimeoutPreRenderContent: TimeInterval = 30.0
     
     //Controls whether to use PrebidMobile's in-app browser or the Safari App for displaying ad clickthrough content.
-    //For original API should be true
     public var useExternalClickthroughBrowser = false
```
YuriyVelichkoPI, line 113: This property looks outdated now. 

**Rendering API**. SDK opens an URL only in the `SFSafariViewController`, right? It means the `imp.clickbrowser` will always be `1`. 

**Original API**. This scenario assumes rendering in any ad server SDK. We can't know which browser will be used. But we should give publishers ability to clarify the behavior if needed. Since the Original API is mostly ised with GMA SDK, we can assume that the click URL will be opened in  `SFSafariViewController` as well. So the default value should be `1` too. 

The name is irrelevant anymore too. There is no attempt to open an URL in something _external_. Let's rename this property to something more OpenRTB-focused, like `impClickbrowser` with default value `@1` and ability to customize it only for Original API integration. The comment to his property should point to the description of the respective field in the OpenRTB standard.

PrebidMobile/PrebidMobileRendering/Prebid/PBMCore/PBMSafariOpener.m
```diff
@@ -34,10 +33,9 @@
 #import "PBMMacros.h"
 
 
-@interface PBMClickthroughBrowserOpener ()
+@interface PBMSafariOpener ()
```
YuriyVelichkoPI, line 36: Let's rename to `PBMSafariVCOpener`. Because there is a difference between opening in the `SFSafariViewController` and the Safari application.

YuriyVelichkoPI (changes requested):

PR #810, opened 2023-03-17
Fix VAST without end card closing issue

### LINKED ISSUE #808: Impossibility to close a vast without end card (under certain conditions)
**Describe the bug**
There are some Vast creatives that when they finish playing, a black screen is shown and they cannot be closed (because the close button does not appear).
The problem is that in this method when creating the companion, it could become nil and later, the creative model asks for `hasCompanionAd`, but it does not have more creatives to show.
https://github.com/prebid/prebid-mobile-ios/blob/41dc565c128d9f9c52335f01ba382d9a0076fe2b/PrebidMobile/PrebidMobileRendering/AdTypes/AdView/PBMCreativeModelCollectionMakerVAST.m#L108-L122

**To Reproduce**
You can use any VAST creative that does not have an end card (e.g. https://prebid-mobile-ad-assets.s3.amazonaws.com/scripts/vasts/vast.xml) and add an empty ad companion creative to it.
```
<Creative>
    <CompanionAds>
    </CompanionAds>
</Creative>
```

This is not the only way, any VAST configuration that causes it to return nil in this method would also pass
https://github.com/prebid/prebid-mobile-ios/blob/41dc565c128d9f9c52335f01ba382d9a0076fe2b/PrebidMobile/PrebidMobileRendering/AdTypes/AdView/PBMCreativeModelCollectionMakerVAST.m#L173-L244

**Solution**
Change to true `hasCompanionAd` only if a not nil companion was obtained.
```
    if (companionItems.count > 0) {
        // Now try to create the companion items creatives.
        // Create a model of the best fitting companion ad.
        PBMCreativeModel *creativeModelCompanion = [self createCompanionCreativeModelWithAd:vastAd
                                                                               companionAds:companionItems
                                                                                   creative:creative];
        if (creativeModelCompanion) {
            // There is at least 1 companion.  Set the flag so that when the initial video creative has completed
            // display, the appropriate view controllers will prevent the "close" button and the learn more after the video has
            // finished, it will instead display the endcard.
            creativeModel.hasCompanionAd = YES;
            [creatives addObject:creativeModelCompanion];
        }
    }
```

OlenaPostindustria:
Closes #808

### DIFF
--- a/PrebidMobile.xcodeproj/project.pbxproj
+++ b/PrebidMobile.xcodeproj/project.pbxproj
@@ -66,9 +66,12 @@
 		533135C6282A869800AA1E4D /* BidTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 533135C5282A869800AA1E4D /* BidTest.swift */; };
 		53322AA3282D45EE0049229D /* PrebidServerEventTracker.swift in Sources */ = {isa = PBXBuildFile; fileRef = 53322AA2282D45EE0049229D /* PrebidServerEventTracker.swift */; };
 		53322AA8282E29910049229D /* EventManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = 53322AA7282E29910049229D /* EventManager.swift */; };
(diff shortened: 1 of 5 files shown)


YuriyVelichkoPI (approved):
LGTM

PR #811, opened 2023-04-04
Support multiformat ad units

### LINKED ISSUE #809: Support multiformat adunits
We have a publisher that wants to create an adunit that supports both banner and outstream video. Prebid.js and Prebid Server call this a 'multiformat' adunit. It's supported in OpenRTB requests by specifying both imp.banner and imp.video in the same imp object.

This is not currently possible with the Prebid SDK

Please update the SDK, both iOS and Android, to allow publishers to declare both banner params and video params for the same adunit.

And of course add test cases and documentation on docs.prebid.org for how to form the request and set up the line items for the response side.

OlenaPostindustria:
Closes #809


PrebidMobile/AdUnits/BannerAdUnit.swift
```diff
@@ -15,16 +15,34 @@
...
+        set { adUnitConfig.adConfiguration.videoParameters = newValue }
+    }
+    
+    public var adFormats: Set<AdFormat> {
+        get { adUnitConfig.adFormats }
+        set { adUnitConfig.adFormats = newValue }
+    }
     
     public init(configId: String, size: CGSize) {
         super.init(configId: configId, size: size)
```
YuriyVelichkoPI, line 37: Can we make the `adFormats` an obligatory property of AdUnit initializer so we are sure that this property will never be undefined?

PrebidMobile/AdUnits/BannerBaseAdUnit.swift
```diff
@@ -12,8 +12,14 @@ limitations under the License.
 
 import Foundation
 
+@objc public protocol BannerBasedAdUnitProtocol {
+    var bannerParameters: BannerParameters { get set }
+}
+
+@available(*, deprecated, message: "This class is deprecated.")
```
YuriyVelichkoPI, line 19: We should inform publishers which class they should use instead of the deprecated one.

PrebidMobile/AdUnits/Parameters/VideoParameters.swift
```diff
@@ -41,7 +41,7 @@ public class VideoParameters: NSObject {
      * "video/mp4"
      * "video/x-ms-wmv"
      */
-    public var mimes: [String]?
+    public var mimes: [String]
```
YuriyVelichkoPI, line 44: Let's put a comment on why the `mimes` can't be optional.

PrebidMobile/AdUnits/RewardedVideoAdUnit.swift
```diff
@@ -15,21 +15,24 @@ limitations under the License.
 
 import UIKit
 
-public class RewardedVideoAdUnit: VideoBaseAdUnit {
+@objcMembers
```
YuriyVelichkoPI, line 18: Just curious - is it mean that this class, and interstitial ad unit as well, hasn't been used in Obj-c apps?

PrebidMobile/PrebidMobileRendering/ORTB/PBMORTBBidRequestExtPrebid.m
```diff
@@ -14,10 +14,19 @@
...
-#import "PBMORTBAbstract+Protected.h"
 
 @implementation PBMORTBBidRequestExtPrebid : PBMORTBAbstract
 
+- (instancetype)init {
+    if (!(self = [super init])) {
+        return nil;
+    }
+   
+    self.targeting = [PBMMutableJsonDictionary new];
```
YuriyVelichkoPI, line 25: It's not super clear from the code, just, please make sure that targeting object is absent in the request if it doesn't have any internal properties.

PrebidMobile/PrebidMobileRendering/Prebid/PBMCore/PBMPrebidParameterBuilder.m
```diff
@@ -81,6 +81,10 @@ - (void)buildBidRequest:(nonnull PBMORTBBidRequest *)bidRequest {
         bidRequest.extPrebid.cache = cache;
     }
     
+    if (adFormats.count >= 2) {
```
YuriyVelichkoPI, line 84: Please comment on why we have this condition here and not just send the `includeformat` for all cases.

YuriyVelichkoPI (changes requested):
Please address several minor comments.

PrebidMobile/AdUnits/RewardedVideoAdUnit.swift
OlenaPostindustria: Actually, there is no need for this attribute in these classes. They inherit from AdUnit class, which already provides this attribute that also applies to subclasses. Removed it from AdUnit subclasses.

PrebidMobile/PrebidMobileRendering/ORTB/PBMORTBBidRequestExtPrebid.m
OlenaPostindustria: Sending this object is required for getting targeting keywords in response. More info [here](https://docs.prebid.org/prebid-server/endpoints/openrtb2/pbs-endpoint-auction.html#targeting).

YuriyVelichkoPI (approved):
LGTM

PR #814, opened 2023-04-05
Fix zsh redirection

ptomasroos:
For some reason the output redirection doesn't work in zsh / mac osx

I've tried to maintain the same feature while support both zsh + bash

Before this change running buildPrebidMobile.sh continuously fails in zsh with the following error message

```
BUILD PREBID MOBILE


./scripts/buildPrebidMobile.sh: line 130: syntax error near unexpected token `<'
```
https://github.com/prebid/prebid-mobile-ios/blob/master/scripts/buildPrebidMobile.sh#L130

### DIFF
--- a/scripts/buildPrebidMobile.sh
+++ b/scripts/buildPrebidMobile.sh
@@ -123,11 +123,12 @@ do
 	# find all .bcsymbolmap and concatinate -debug-symbols
 	debugSymbolsBcsymbolmap=""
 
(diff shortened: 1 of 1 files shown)


OlenaPostindustria:
Hi, @ptomasroos!
Thank you for the fix. 

However, it makes sense to entirely remove the support of building PrebidMobile with bitcode(described in #815) as bitcode started being deprecated since Xcode 14.  This means that a failing piece of code in `buildPrebidMobile.sh` should be removed and after these changes `buildPrebidMobile.sh` should run successfully under both zsh + bash.

ptomasroos:
Great! Missed that!

PR #823, opened 2023-04-20
Add `PBM` prefixes to the `@objc` classes

### LINKED ISSUE #821: Prebid Mobile SDK 2.0.0+ produces Objective-C Log class with generic name in global namespace
**Describe the bug**
I am trying to upgrade our Prebid Mobile SDK from `1.12` to `2.0.8`, however due to the generic naming of the generated Objective-C [Log](https://github.com/prebid/prebid-mobile-ios/blob/6ad2e16b59188ea4e3bd8b9a89d302ca7c38ea4e/PrebidMobile/Logging/Log.swift#L18-L19) class, there is a naming clash with another problematic SDK and compilation fails.

**To Reproduce**
Steps to reproduce the behaviour:
1. Open a project that uses PrebidMobile SDK version 2.0.0 or higher
2. Import another framework that has an `@objc`/`@objcMembers` class named `Log`, or create a sub module in the project with a `Log` class, eg:
```swift
@objcMembers public class Log: NSObject {
  public static func debug(_ message: String) {
    print("LOG DEBUG: \(message)")
  }
}
```
3. Build project
4. Observe Swift compile error

**Expected behavior**
Project should build successfully without any errors. Any Objective-C code (generated or otherwise) in the SDK should follow Objective-C naming conventions to ensure unique class names in the global application namespace - see [Apple's conventions](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/Conventions/Conventions.html).

**Screenshots**
Screen Shot 2023-04-18 at 1 28 57 PM

 - OS: macOS Monterey Version 12.6.5
 - Xcode 14.1

**Additional context**
Since Objective-C code is globally name spaced, all `@objc`/`@objcMembers` Swift classes should prepend a module identifier to the generated objc class through the `@objc` attribute, eg:
```swift
@objc(PBMLog) @objcMembers
public class Log: NSObject { // ...
```

The same should be applied if any other classes use `@objc` or `@objcMembers`

OlenaPostindustria:
Closes #821

### DIFF
--- a/PrebidMobile.xcodeproj/project.pbxproj
+++ b/PrebidMobile.xcodeproj/project.pbxproj
@@ -60,7 +60,7 @@
 		47D9AFD469AA0D5B6CE79D43 /* MockDelayedReachability.swift in Sources */ = {isa = PBXBuildFile; fileRef = 47D9AF39FFB17935DB0C57B5 /* MockDelayedReachability.swift */; };
 		47D9AFFB2F592206309C5FF0 /* ACJInterstitial.json in Resources */ = {isa = PBXBuildFile; fileRef = 47D9ADF3B4DA145410B2AFF6 /* ACJInterstitial.json */; };
 		53019E6127E9E40700D509C1 /* LogTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 53019E6027E9E40700D509C1 /* LogTest.swift */; };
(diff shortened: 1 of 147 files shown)


YuriyVelichkoPI (approved):
LGTM

Issue #840, opened 2023-05-04
Memory Leak when fetch Demand


ianq:
**Describe the bug**
When through instrument and code a memory leak whenever ```fetchdemand``` is called

**To Reproduce**
See Stack Trace for the memory leak

<img width="510" alt="Screenshot 2023-05-04 at 5 34 57 PM" src="https://user-images.githubusercontent.com/19857086/236168121-724533e6-ac60-41ad-b277-4ddd776a844f.png">

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**
 - Device: [e.g. iPhone6] iPhone12
 - OS: [e.g. iOS8.1] iOS16.2

**Additional context**

YuriyVelichkoPI:
Hi @ianq ! Thanks for the report! 

Could you share a piece of code describing the usage of the `fetchDemand`? Or can it be reproduced in the Demo App?

OlenaPostindustria:
Hi, @ianq ! Thanks for the report.
I suppose you are using Original API. I managed to reproduce the memory leak in `PrebidDemo` and fixed it in #849. Please, let me know if there are any other issues with leaking.

PR #849, opened 2023-05-19
Fix Dispatcher leak

OlenaPostindustria:
#840

### DIFF
--- a/PrebidMobile/AdUnits/AdUnit.swift
+++ b/PrebidMobile/AdUnits/AdUnit.swift
@@ -74,6 +74,10 @@ public class AdUnit: NSObject, DispatcherDelegate {
         // PBS should cache the bid for original api.
         Prebid.shared.useCacheForReportingWithRenderingAPI = true
     }
+    
+    deinit {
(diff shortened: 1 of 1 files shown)


PrebidMobile/AdUnits/AdUnit.swift
```diff
@@ -74,6 +74,10 @@ public class AdUnit: NSObject, DispatcherDelegate {
         // PBS should cache the bid for original api.
         Prebid.shared.useCacheForReportingWithRenderingAPI = true
     }
+    
+    deinit {
+        dispatcher?.invalidate()
```
YuriyVelichkoPI, line 79: Can we invalidate / stop dispatcher in the `deinit` of `Dispatcher` itself?

OlenaPostindustria: Unfortunately, we can't cause `Timer` holds a strong reference to `Dispatcher`, which means `deinit` is never called until `Timer` gets invalidated.

YuriyVelichkoPI (approved):
LGTM

PR #884, opened 2023-07-24
Change default omidpv value to Prebid SDK version in rendering API

jsligh:
Changed to use Prebid SDK Version (like it does on Android).  Removed the omidversion function as it was a hardcoded value and no longer needed as the Prebid SDK version is always used.  Also updated the tests to use the SDK version and ran all tests and they passed.

Addresses this issue: https://github.com/prebid/prebid-mobile-ios/pull/884


YuriyVelichko (approved):
LGTM

YuriyVelichkoPI (approved):
LGTM

YuriyVelichkoPI:
@jsligh, the changes look good. However, the tests failed, could you check the reason and fix? 

From time to time some steps failed and we just restart them. I've tried to restart one without success still see an error. Will try one more time.

Issue #876, opened 2023-07-06
Implement third-party tracker when clicked - Link.ClickTrackers


YuriyVelichkoPI:
## Objectives

Need to implement functionality requested in this issue: https://github.com/prebid/prebid-mobile-android/issues/649 and implemented in the scope of this PR: https://github.com/prebid/prebid-mobile-android/pull/652

cc: @clho40

YuriyVelichkoPI:
Hi @jsligh ! Not exactly. The code you referenced is responsible for the _impression_ trackers - when the ad appears on the screen (becomes visible). 

The original issue is about the _click trackers_ when the user taps on the ad. 

These trackers are placed in the `NativeAdMarkup.link.clicktrackers` object. And should be processed in the `NativeAd.handleClick()` method. 

A detailed description of the Native Ads can be found here:
chrome-extension://efaidnbmnnnibpcajpcglclefindmkaj/https://iabtechlab.com/wp-content/uploads/2016/07/OpenRTB-Native-Ads-Specification-Final-1.2.pdf 

In this case, we extend the support of the `link` object, described in the section `5.7 Link Response Object`, by implementing the requirements for `clicktrackers` object.

PR #885, opened 2023-07-25
Implement third-party tracker when clicked - Link.ClickTrackers

jsligh:
This has been implemented on Android and is now firing the events for Native Ads on iOS.

This is for this task/issue: https://github.com/prebid/prebid-mobile-ios/issues/876


PrebidMobile/AdUnits/Native/NativeAd.swift
```diff
@@ -296,6 +296,21 @@ import UIKit
                 Log.debug("Could not open click URL: \(clickUrl)")
             }
         }
+        if let clickTrackers = nativeAdMarkup?.link?.clicktrackers {
```
YuriyVelichkoPI, line 299: Let's process the click trackers in the same way as implemented in Android PR - once the click URL is successfully opened: 

https://github.com/prebid/prebid-mobile-android/pull/652/files#diff-9cf1ccf8aa5d0eb0a2a198c1261d70fc83d465a1ea3f48351def5e72b971148eR420

In our case is when `openURLWithExternalBrowser` returns `true`.

PrebidMobile/AdUnits/Native/NativeAd.swift
```diff
@@ -296,6 +296,21 @@ import UIKit
...
+            fireClickTrackers(clickTrackersUrls: clickTrackers)
+        }
+    }
+    
+    
+    private func fireClickTrackers(clickTrackersUrls: [String]) {
+        if clickTrackersUrls.count > 0 {
+            TrackerManager.shared.fireTrackerURLArray(arrayWithURLs: clickTrackersUrls) { [weak self] _ in
+                guard self != nil else {
```
YuriyVelichkoPI, line 308: It looks like this code does not perform any meaningful job. Let's leave the completion handler empty.

YuriyVelichkoPI (changes requested):

YuriyVelichkoPI (approved):
LGTM

PR #887, opened 2023-07-26
Support GMA SDK 10.8.0

### LINKED ISSUE #886: Support GMA SDK 10.7.0 & 10.8.0
## Motivation
A new GMA SDK version was released.

## Objectives
- investigate changes in GMA SDK 10.7.0 and 10.8.0;
- if needed, update the behavior in Original API, AdMob Adapters, and GAMEventHandlers respectively to the changes in the GMA SDK.

OlenaPostindustria:
Closes #886 

- Deprecated `Prebid.initializeSDK(_ gadMobileAdsObject: AnyObject? = nil, _ completion: PrebidInitializationCallback? = nil)` initializer
- Add new initializer - `Prebid.initializeSDK(gadMobileAdsVersion: String? = nil, _ completion: PrebidInitializationCallback? = nil)`
- Updated dependencies 
- Updated latest tested GMA version
- Updated GAM version checking logic
- Add unit tests


PrebidMobile/ConfigurationAndTargeting/Prebid.swift
```diff
@@ -183,16 +180,27 @@ public class Prebid: NSObject {
     /// - Parameters:
     ///   - gadMobileAdsObject: GADMobileAds object
     ///   - completion: returns initialization status and optional error
+    @available(*, deprecated, message: "If the GMA SDK version is higher than 10.7.0 you should use `initializeSDK(gadMobileAdsVersion:, _ completion:)` method in order to initialize Prebid SDK.")
```
YuriyVelichkoPI, line 183: We shouldn't deprecate this method so far because publishers still can use previous GMA SDK versions, and we shouldn't push them to migrate (via warning). 

We should log the message in the `checkGMAVersion` method if the publisher updated the GMA SDK but still use this initializer.

YuriyVelichkoPI (changes requested):

YuriyVelichkoPI:
cc: @jsligh 

This PR is intended to fix UI tests.

Issue #888, opened 2023-07-27
Incorrect if #available macro usage for SKAdNetwork


friar6123:
**Describe the bug**
SKAdNetwork is available in 16.1, not 16.0 minimum which the prebid code has. Xcode suggests fix which allows prebid to build. So I think the value should be changed from 16.0 to 16.1

**To Reproduce**
1. Configure iOS target with minimum deployment of ios 14
2. Try to build

**Expected behavior**

**Screenshots**
<img width="833" alt="Screenshot 2023-07-27 at 1 55 25 PM" src="https://github.com/prebid/prebid-mobile-ios/assets/102054729/c3cc6b12-eb80-415c-a1db-5a5058dc2e69">

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #890, opened 2023-08-01
Fix SKAdN sourceIdentifier availability check

OlenaPostindustria:
#888


YuriyVelichkoPI (approved):
LGTM. 

But just want to mention that building with Xcode 14.* does not leads to this error because: 

image

PR #896, opened 2023-08-04
Fix double encoding and replace bid price on Native Ad

Macjon:
This PR fixes a bug where the {AUCTION_PRICE} template in a url was not replaced with the bid price. 
And fixes a double encoding of the click url from the native ad.

### DIFF
--- a/PrebidMobile/AdUnits/Native/NativeAd.swift
+++ b/PrebidMobile/AdUnits/Native/NativeAd.swift
@@ -107,7 +107,11 @@ import UIKit
         }) else {
             return nil
         }
-        
+
+        let macrosHelper = PBMORTBMacrosHelper(bid: rawBid)
+        rawBid.adm = macrosHelper.replaceMacros(in: rawBid.adm)
+        rawBid.nurl = macrosHelper.replaceMacros(in: rawBid.nurl)
+
         let ad = NativeAd()
         
         let internalEventTracker = PrebidServerEventTracker()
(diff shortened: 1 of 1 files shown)


OlenaPostindustria (approved):
LGTM

Macjon:
@OlenaPostindustria I removed the line `let clickUrlString = clickUrl.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed),` in the click handler. This line percent encoded the url. But our url was already encoded but as you also said I didn't find it either where the url is encoded.

OlenaPostindustria:
Removing that line of code may cause an issue with URLs that have not been encoded yet

YuriyVelichkoPI:
Hi @Macjon! It's a good point that the URL can be already encoded. However, it is not an obligatory requirement for the OpenRTB spec for the NativeAds. See section 5.7 of the following spec: 

https://iabtechlab.com/wp-content/uploads/2016/07/OpenRTB-Native-Ads-Specification-Final-1.2.pdf

I haven't found any mentions of the requirements for the `link.url` string in the native ad markup. If you know such a requirement, please point me, and we will remove the encoding with a respective comment so no one else returns it back later.

So the most appropriate way will be to support both cases. The SDK should check whether the `url` is already encoded and apply encoding only if it isn't. 

In this way, we will support both kinds of demand partners: those who send encoded URLs and those who do not. 

Also, we should add tests to be sure that all kinds of URLs are processed correctly.

YuriyVelichkoPI:
We can handle the encoding issue in the following way:

1. Try to create an `URL` as is without encoding. 
2. If the URL is created without errors - just use it
3. IF the URL is not created due to the error - try to encode it

Precondition - all macros should be resolved before converting the string to the URL. 

It would be great if we apply this approach to all places in the SDK where we encode the URL's string now.

YuriyVelichkoPI:
@Macjon, we introduced the needed fixes in this PR: #902.

PR #902, opened 2023-08-14
Handle URL double encoding & replace bid price on Native Ad

### LINKED ISSUE #901: URL double encoding issue
## Objectives 

The PrebidMobile SDK encodes URLs, which have been already encoded. 

We can handle this issue in the following way:

1. Try to create an `URL` as is without encoding. 
2. If the URL is created without errors - just use it
3. IF the URL is not created due to the error - try to encode it

Precondition - all macros should be resolved before converting the string to the URL. 

It would be great if we apply this approach to all places in the SDK where we encode the URL's string now.

_Originally posted by @YuriyVelichkoPI in https://github.com/prebid/prebid-mobile-ios/issues/896#issuecomment-1672988038_

OlenaPostindustria:
Closes #901


YuriyVelichkoPI (approved):
LGTM

PR #914, opened 2023-09-21
Implement multiformat ad unit for Original API

### LINKED ISSUE #913: Multiformat ad unit for Original API
## Feature Request

Introduce the ability to make multiformat bid requests for Original API. The supported formats should be:

- Banner
- Video
- Native

## Motivation

One of Magnite's clients uses multiformat ad places in their apps. They follow the `Combining Custom-Rendered Native Ad and Banner Ad Requests` guide from Google to show the banner and native ads in the same slots. Also, they have long-term plans to add the outstream video to the same slot. 

In order to support such a setup, Prebid SDK should provide the ability to perform banner + native (+ video) bid requests and to get a single winning bid with a respective type.    

## Proposal

The current Original API has the following structure: 

|Android|iOS|
|-------|---|
| 
Original API - Current Android
| 
Original API - Current iOS
|

We propose to make it more generic, extending with the `PrebidAdUnit` and `PrebidRequest` classes as shown on the following diagram (note that the diagram is based on iOS architecture, but the Origianal API for Android will stay the same plus new classes introduced bellow): 

Multifiormat Ad Unit - Original API - New

The new approach will allow publishers to combine any formats for the bid request, providing respective properties to `PrebidRequest`. The `PrebidAdUnit` will only be responsible for managing the bid-requesting process. 

In order to not modify the current approach, the `PrebidAdUnit` will use the existing parent class `AdUnit` to perform all needed work. 

In addition, to provide the bid response in a generic essential way, the proposed implementation will absorb the implementation of #912 introducing the `BidInfo` class, providing all the needed info about the auction result and the winning bid.

Even though these changes aim to introduce the approach for multiformat bidding, it's easy to see that it deprecates the current API classes such as `BannerAdUnit` and `InterstitialAdUnit`. Since the Original API, in a nutshell, is dedicated to making a bid request and providing the info about the auction, the new  `PrebidAdUnit` and `PrebidRequest` objects will cover all the integration cases (specific ad unit). Moreover, it will allow to implement #877 in the more RTB-manner way by adding the JSON-configuration properties to the `PrebidRequest` components. 

The proposal assumes no breaking changes in the API classes or bidding algorithms. 

However, the current API can or should eventually be deprecated and removed to support and develop the single generic multiformat integration approach.

## Objectives

Following the class diagram above, perform the following changes. 

SDK: 

- Introduce the `NativeParameters` class to aggregate the properties needed to create RTB request for the Native Ad. 
- Introduce the `PrebidRequest` class - the container for ad properties. 
- Introduce the `PrebidAdUnit` class - the manager of bid requests for multiformat ad places. 
- Introduce the `BidInfo` class to provide comprehensive information about the bid response. 
- Modify the `BannerParameters` class by adding properties needed in multiformat requests. Do not change anything else. 
- Modify the `VideoParameters` class by adding properties needed in multiformat requests. Do not change anything else. 
- Modify the `AdUnit` base class by adding new `fetchDemand` methods. The new methods must use but not change the current `fetchDemand` behavior. 

Applications:

- TBD: the integration case design. Introducing the dynamic case where the user can select which formats to add to the request might make sense. 
- Introduce the case with multiformat bid request for Original API.
- Introduce UI tests for the new cases. 

SDK Tests:

- Make sure that there is no regression in the SDK behavior after implementing the multiformat ad unit. All current tests should pass without modifications. 
- Introduce Unit Tests tests for all new classes and properties.
- Introduce  Tests tests for all new classes and properties.

Documentation:

- Add integration examples and description to the SDK docs based on the Demo App.

OlenaPostindustria:
Closes #913


PrebidMobile/AdUnits/MultiformatAdUnit/BidInfo.swift
```diff
@@ -0,0 +1,35 @@
...
+limitations under the License.
+*/
+
+import Foundation
+
+@objcMembers
+@objc(PBMBidInfo)
+public class BidInfo: NSObject {
+  
+    public private(set) var result: ResultCode
```
YuriyVelichkoPI, line 22: Please rename it to `resultCode`. The word result is to abstract for this property.

YuriyVelichkoPI (approved):
LGTM

PR #928, opened 2023-10-10
Support impression-level GPID parameter

### LINKED ISSUE #925: Support impression-level GPID parameter

Publishers want to be able to specify the Global Placement ID (GPID) on the ad call. This is not currently possible through the SDK.

Publishers should be able to specify the GPID for each impressions, and it gets passed to Prebid Server as imp[].ext.gpid

Originally, I was hoping that we could make this a generic ORTB parameter interface, but conversations around issue https://github.com/prebid/prebid-mobile-ios/issues/877 revealed that the size of that request was slowing this smaller enhancement.

OlenaPostindustria:
Closes #925


PrebidMobile/AdUnits/AdUnit.swift
```diff
@@ -378,6 +378,12 @@ public class AdUnit: NSObject, DispatcherDelegate {
         adUnitConfig.clearUserData()
     }
     
+    // MARK: GPID
+    
+    public func setGPID(_ gpid: String?) {
```
YuriyVelichkoPI, line 383: Let's add a getter to be consistent with other properties.

PrebidMobile/AdUnits/MultiformatAdUnit/PrebidRequest.swift
```diff
@@ -45,6 +47,12 @@ public class PrebidRequest: NSObject {
         super.init()
     }
     
+    // MARK: GPID
+    
+    public func setGPID(_ gpid: String?) {
```
YuriyVelichkoPI, line 52: We should keep API consistent. Since it is a new API Class, we can consider using only setters for request properties. However, in this case, we should remove other setters like `getExtData` and align this approach with Android.

YuriyVelichkoPI (changes requested):

YuriyVelichkoPI (approved):
LGTM

PR #931, opened 2023-10-11
Surface win URL via SDK

### LINKED ISSUE #926: Surface win URL via SDK

One of our publishers wants to render without using the Prebid Universal Creative, but also to use Magnite's analytics system. To achieve this, they will need to invoke the analytics event in their rendering process. Prebid Server returns a field that contains the URL they need to hit:

```
seatbid[].bid[].ext.prebid.events.win
```

We are asking for a way for the app to obtain this Prebid Server field from the SDK for each bid response. The app will be responsible for hitting the URL at the appropriate time.

OlenaPostindustria:
Closes #926


YuriyVelichkoPI (approved):
LGTM

PR #941, opened 2023-12-13
Support Collecting Arbitrary ORTB Parameters

### LINKED ISSUE #877: Support collecting GPID
We're getting requests from publishers pass GPID through the bidding ecosystem.

There should be a way in the Prebid SDK to enter `imp[n].ext.gpid`. Not clear whether this is a new one-off function or even better, it would be great to be able to pass arbitrary ORTB fields into the request. The SDK could support flexible ORTB. e.g.

```
   addOrtbGlobal('{"ext":{"prebid":{"debug":1,"trace":"verbose"}}}');
   adUnit = BannerAdUnit(configId: CONFIG_ID, size: adSize, ORTB_OBJECT)
```

This can be done by adding imp.ext.gpid to the stored request, but this is a hassle because it creates the need for users to enter the GPID somewhere so it can be loaded into the PBS DB. That duplication of data entry creates a place for discrepancies to happen.

**Additional context**

Prebid.js has supported accepting arbitrary ORTB for a year now and it's worked quite well to handle new use cases.

jsligh:
Closes #877

### DIFF
--- a/PrebidMobile/AdUnits/AdUnit.swift
+++ b/PrebidMobile/AdUnits/AdUnit.swift
@@ -24,7 +24,7 @@ public class AdUnit: NSObject, DispatcherDelegate {
     var adSizes: [CGSize] {
         get { [adUnitConfig.adSize] + (adUnitConfig.additionalSizes ?? []) }
     }
-    
+        
     private static let PB_MIN_RefreshTime = 30000.0
     
     private(set) var dispatcher: Dispatcher?
@@ -389,22 +389,12 @@ public class AdUnit: NSObject, DispatcherDelegate {
     
     // MARK: Global ORTBObject
     
-    public func setGlobalOrtbObject(_ ortbObject: [String: Any]?) {
-        adUnitConfig.globalOrtbObject = ortbObject
+    public func setOrtbConfig(_ ortbObject: [String: Any]?) {
+        adUnitConfig.ortbConfig = ortbObject
     }
     
-    public func getGlobalOrtbObject() -> [String: Any]? {
-        return adUnitConfig.globalOrtbObject
-    }
-    
-    // MARK: imp-level ORTBObject
-    
-    public func setImpOrtbObject(_ ortbObject: [String: Any]?) {
-        adUnitConfig.impOrtbObject = ortbObject
-    }
-    
-    public func getImpOrtbObject() -> [String: Any]? {
-        return adUnitConfig.impOrtbObject
+    public func getOrtbConfig() -> [String: Any]? {
+        return adUnitConfig.ortbConfig
     }
     
     // MARK: - others
--- a/PrebidMobile/PrebidMobileRendering/ORTB/PBMORTBBidRequest.m
+++ b/PrebidMobile/PrebidMobileRendering/ORTB/PBMORTBBidRequest.m
@@ -68,9 +68,58 @@
     ext[@"prebid"] = [[self.extPrebid toJsonDictionary] nullIfEmpty];
     ret[@"ext"] = [[ext pbmCopyWithoutEmptyVals] nullIfEmpty];
     
-    for (id key in self.ortbObject)
-        [ret setObject:[self.ortbObject objectForKey:key] forKey: key];
+    //remove "protected" fields from ortbObject then do a merge but merge ret into the ortbObject (addEntriesFromDictionary)
+    NSMutableDictionary *o = [self.arbitraryJsonConfig mutableCopy];
+    if (o[@"regs"]) {
+        o[@"regs"] = nil;
+    }
+    if (o[@"device"]) {
+        o[@"device"] = nil;
+    }
+    if (o[@"geo"]) {
+        o[@"geo"] = nil;
+    }
+    //merge with config from API/JSON
+    ret = [self mergeDictionaries: ret joiningArgument2: o joiningArgument3: false];
     
+    NSMutableDictionary *o2 = [self.ortbObject mutableCopy];
+    
+    if (o2[@"regs"]) {
+        o2[@"regs"] = nil;
+    }
+    if (o2[@"device"]) {
+        o2[@"device"] = nil;
+    }
+    if (o2[@"geo"]) {
+        o[@"geo"] = nil;
+    }
+    //merge with ortbConfig from SDK
+    ret = [self mergeDictionaries: ret joiningArgument2: o2 joiningArgument3: true];
+    
+    ret = [ret pbmCopyWithoutEmptyVals];
+    
+    return ret;
+}
+
+- (nonnull PBMMutableJsonDictionary *)mergeDictionaries:(NSMutableDictionary*)dictionary1 joiningArgument2:(NSMutableDictionary*)dictionary2
+                                       joiningArgument3:(Boolean)firstHasPriority{
+    PBMMutableJsonDictionary *ret = dictionary1;
+
+    for (id key in dictionary2)
+        if ([ret objectForKey: key]){
+            if ([[ret objectForKey: key] isKindOfClass: [NSDictionary class]]) {
+                //if is dictionary, need to call this method recursively for ret object for key and dictionary2 for key
+                [ret setObject:[self mergeDictionaries:[ret objectForKey: key] joiningArgument2: [dictionary2 objectForKey: key] joiningArgument3:firstHasPriority] forKey: key];
+            } else {
+                if (!firstHasPriority) {
+                    [ret setObject:[dictionary2 objectForKey: key] forKey:key];
+                }
+            }
+            //not sure what to do if array of objects
+        } else {
+            [ret setObject:[dictionary2 objectForKey:key] forKey: key];
+        }
+        
     ret = [ret pbmCopyWithoutEmptyVals];
     
     return ret;
@@ -101,7 +150,7 @@
     
     _extPrebid = [[PBMORTBBidRequestExtPrebid alloc] initWithJsonDictionary:jsonDictionary[@"ext"][@"prebid"] ?: @{}];
     
-    _ortbObject = jsonDictionary;
+    _arbitraryJsonConfig = jsonDictionary;
     
     return self;
 }
--- a/PrebidMobile/PrebidMobileRendering/Prebid/PBMCore/AdUnitConfig.swift
+++ b/PrebidMobile/PrebidMobileRendering/Prebid/PBMCore/AdUnitConfig.swift
@@ -80,10 +80,10 @@ public class AdUnitConfig: NSObject, NSCopying {
     
     public var gpid: String?
     
-    public var globalOrtbObject: [String: Any]?
-    
-    public var impOrtbObject: [String: Any]?
-
+    public var ortbConfig: [String: Any]? {
+        get {adConfiguration.ortbConfig}
+        set {adConfiguration.ortbConfig = newValue}
+    }
     
     // MARK: - Public Methods
     
--- a/PrebidMobile/PrebidMobileRendering/Prebid/PBMCore/PBMPrebidParameterBuilder.m
+++ b/PrebidMobile/PrebidMobileRendering/Prebid/PBMCore/PBMPrebidParameterBuilder.m
@@ -73,7 +73,6 @@
     bidRequest.extPrebid.storedAuctionResponse  = Prebid.shared.storedAuctionResponse;
     bidRequest.extPrebid.dataBidders            = self.targeting.accessControlList;
     bidRequest.extPrebid.storedBidResponses     = [Prebid.shared getStoredBidResponses];
-    bidRequest.ortbObject = self.adConfiguration.globalOrtbObject;
     
     if (Prebid.shared.useCacheForReportingWithRenderingAPI) {
         PBMMutableJsonDictionary * const cache = [PBMMutableJsonDictionary new];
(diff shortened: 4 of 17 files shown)


PrebidMobile/PrebidMobileRendering/Prebid/PBMCore/AdUnitConfig.swift
```diff
@@ -79,6 +79,10 @@ public class AdUnitConfig: NSObject, NSCopying {
     }
     
     public var gpid: String?
+    
+    public var ortbObject: [String: Any]?
+    
+    public var WTF: String?
```
YuriyVelichkoPI, line 85: :D

PrebidMobile/AdUnits/AdUnit.swift
```diff
@@ -387,6 +387,16 @@ public class AdUnit: NSObject, DispatcherDelegate {
         return adUnitConfig.gpid
     }
     
+    // MARK: ORTBObject
+    
+    public func setOrtbObject(_ ortbObject: [String : Any]?) {
```
YuriyVelichkoPI, line 392: Should the ortbObject be a string? So, publishers can provide the raw JSON.

PrebidMobile/PrebidMobileRendering/Prebid/PBMCore/PBMPrebidParameterBuilder.m
```diff
@@ -178,6 +178,7 @@ - (void)buildBidRequest:(nonnull PBMORTBBidRequest *)bidRequest {
         nextImp.extPrebid.storedAuctionResponse = Prebid.shared.storedAuctionResponse;
         nextImp.extPrebid.isRewardedInventory = self.adConfiguration.adConfiguration.isOptIn;
         nextImp.extGPID = self.adConfiguration.gpid;
+        nextImp.extOrtbObject = self.adConfiguration.ortbObject;
```
YuriyVelichkoPI, line 181: Hi @jsligh , @bretg !

I would like to clarify the scope of the implementation to provide the proper review. 

For now, I see support only for the `imp.ext` field. So, the publisher can set the value of this by providing the respective JSON. 

However, from my understanding of the feature request, this new API should provide the ability to set _any_ part of the request with the given objects. 

For instance, the publisher should be able to set the whole imp object like: 

```
adUnit.setOrtb("
{
 "imp": [
    {
      "id": "1",
      "bidfloor": 0.03,
      "banner": {
        "h": 250,
        "w": 300,
        "pos": 0
      }
    }
  ]
}
")
```

Or set the app info like: 

```
adUnit.setOrtb("
{
"app": { "id": "agltb3B1Yi1pbmNyDAsSA0FwcBiJkfIUDA", "name": "Yahoo Weather", "cat": [ "IAB15", "IAB15-10" ], "ver": "1.0.2", "bundle": "12345", "storeurl": "https://itunes.apple.com/id628677149", "publisher": { "id": "agltb3B1Yi1pbmNyDAsSA0FwcBiJkfTUCV", "name": "yahoo", "domain": "www.yahoo.com" } },
})
```

Note: these examples are copy-pasts from the Opertb doc just for demo purposes. Don't treat them as an implementation spec.

So, if the request is about the second variant, the implementation should be extended.

bretg: You're right @YuriyVelichkoPI that the goal is for the SDK to be able to supply any arbitrary ORTB parameter.

 My suggestion in the original issue was to break this into two parts: the "global" part and the "imp-level" part:
```
   addOrtbGlobal('{"ext":{"prebid":{"debug":1,"trace":"verbose"}}}');
   adUnit = BannerAdUnit(configId: "my-stored-request", size: adSize, "{instl:1, ext:{ gpid: "222", data: { foo: "bar" }})
```

I don't see them every wanting to supply the entire `imp` object, but they need to be able to add any field under imp, not just in imp.ext.

bretg: The global `ext` object (and any other global params) would be set in a new function, e.g. `addOrtbGlobal`

PrebidMobileTests/RenderingTests/Tests/ParameterBuilderTests/PrebidParameterBuilderTest.swift
```diff
@@ -784,6 +784,24 @@ class PrebidParameterBuilderTest: XCTestCase {
...
     }
+    
+    func testArbitraryORTBParams() {
+        let gpid = "/12345/home_screen#identifier"
+        let globalOrtb = ["arbitraryparam1": "param1", "arbitraryparam2": 2, "arbitraryparam3": false] as [String : Any]
+        let impOrtb = ["arbimpparam1": "impparam1", "arbitraryparam2": 2, "arbitraryparam3": false] as [String : Any]
+        let adUnit = AdUnit(configId: "test", size: CGSize.zero, adFormats: [.banner])
+        adUnit.setGPID(gpid)
+        adUnit.setGlobalOrtbObject(globalOrtb)
+        adUnit.setImpOrtbObject(impOrtb)
```
YuriyVelichkoPI, line 795: Hi @jsligh, @bretg ! 

The current implementation looks good, and it can address the needs described in the ticket. But also, it seems too fragile and publisher-dependent in the long-term perspective. 

For example, by providing an API to set the Global object and Imp object independently sooner or later, we will get a situation where the publisher mixes these approaches, and the final request will have the wrong structure. 

Also, correct me if I am wrong, but from what I see in the `PBMORTBAbstractTest.swift`, if a publisher provides, for example, the same `ext` object via the `ortbObject` and via `AdUnit` API, the filed will be duplicated in the final bid request. Instead, the data provided in ortbObject should override any other data. At the same time, we can't just ignore data received from API fields because they can be set by 3p services, like identity providers, so the SDK update may lead to losing data in the bid request silently.

Also, there are no changes for Rendering and Mediation API in this PR. 

As I told you from the very beginning, it is an extremely complex task that assumes a lot of design work first. So, instead of searching for implementation gaps, let's design the implementation spec first, and if it covers all the objectives, is robust for integration, and is unambiguous for utilization, we'll update the PR accordingly. 

I can provide you with a draft version early next week so we can discuss it. 

cc: @alexsavelyev , @mmullin

bretg: Thanks for the thoughts Yuriy. My take is based on the experience we've had with Prebid.js:

- ORTB is a standard. Developers have to understand it anyhow in order to debug.
- Things in our industry change too quickly for an API to keep up
- Therefore, instead of trying to design a new API function everytime there's a new feature, letting developers have direct access to ORTB is better in every way: smaller API footprint, more robust to continuing evolution, developers have less to learn.

I agree that the "rendering API" also needs this functionality, though as we discussed earlier today, I frankly have issues with a core assumption of that API -- that Prebid can always choose the winner. I'm going to open a separate issue to discuss that.

YuriyVelichkoPI: I agree with all points and statements except: 

> Developers have to understand it anyhow in order to debug.

With full respect, but from all our experience in mobile development on both sides - publishers and the ad tech industry, I should say that developers of mobile apps don't have a knowledge of ORTB. 

Open RTB, as usual, is used in s2s cases. In the mobile world, the ad request (SDK --> SSP) has a custom format depending on each particular implementation. App developers, in their own turn, work only with the API of each particular SDK that provides a certain set of parameters to tune. This API and integration documentation is the final frontier of their knowledge of ad tech and what is happening behind the ad requests to the SSP (or other platforms).

Of course, with exceptions when publishers have an AdOps department that directly works on monetization. These guys have a great understanding of the ecosystem. 

So, in order to prevent logical errors, it makes sense to restrict the ability to shoot themselves in the foot as much as possible.

YuriyVelichkoPI: @bretg @jsligh please review the proposal. For now, it is in the doc so you can easily edit it. Once everything is aligned we can copy the text here: 

https://docs.google.com/document/d/1cRWDNO6i7wLyQdQA9ZzmK0QumIr2vvC40r5GhHhOSz4/edit?usp=sharing

Sorry if the intro sections are too long. I tried to highlight the context and pitfalls that impact the proposed design. 

cc: @alexsavelyev @mmullin

PrebidMobile/AdUnits/AdUnit.swift
```diff
@@ -387,6 +387,16 @@ public class AdUnit: NSObject, DispatcherDelegate {
         return adUnitConfig.gpid
     }
     
+    // MARK: Global ORTBObject
+    
+    public func setOrtbConfig(_ ortbObject: [String: Any]?) {
```
YuriyVelichkoPI, line 392: The custom request should be set as a String. 

Publishers should just put the JSON string as a configuration option and not a dictionary.

PrebidMobile/PrebidMobileRendering/ORTB/PBMORTBBidRequest.m
```diff
@@ -68,6 +68,58 @@ - (nonnull PBMJsonDictionary *)toJsonDictionary {
     ext[@"prebid"] = [[self.extPrebid toJsonDictionary] nullIfEmpty];
     ret[@"ext"] = [[ext pbmCopyWithoutEmptyVals] nullIfEmpty];
     
+    //remove "protected" fields from ortbObject then do a merge but merge ret into the ortbObject (addEntriesFromDictionary)
+    NSMutableDictionary *o = [self.arbitraryJsonConfig mutableCopy];
```
YuriyVelichkoPI, line 72: We should avoid such names for variables.

PrebidMobile/PrebidMobileRendering/ORTB/PBMORTBBidRequest.m
```diff
@@ -68,6 +68,58 @@ - (nonnull PBMJsonDictionary *)toJsonDictionary {
...
+        if ([ret objectForKey: key]){
+            if ([[ret objectForKey: key] isKindOfClass: [NSDictionary class]]) {
+                //if is dictionary, need to call this method recursively for ret object for key and dictionary2 for key
+                [ret setObject:[self mergeDictionaries:[ret objectForKey: key] joiningArgument2: [dictionary2 objectForKey: key] joiningArgument3:firstHasPriority] forKey: key];
+            } else {
+                if (!firstHasPriority) {
+                    [ret setObject:[dictionary2 objectForKey: key] forKey:key];
+                }
+            }
+            //not sure what to do if array of objects
```
YuriyVelichkoPI, line 118: Not sure about this comment. Does it mean that the algorithm doesn't merge part of the data?

PrebidMobile/PrebidMobileRendering/ORTB/PBMORTBBidRequest.m
```diff
@@ -68,6 +68,58 @@ - (nonnull PBMJsonDictionary *)toJsonDictionary {
...
+    }
+    //merge with ortbConfig from SDK
+    ret = [self mergeDictionaries: ret joiningArgument2: o2 joiningArgument3: true];
+    
+    ret = [ret pbmCopyWithoutEmptyVals];
+    
+    return ret;
+}
+
+- (nonnull PBMMutableJsonDictionary *)mergeDictionaries:(NSMutableDictionary*)dictionary1 joiningArgument2:(NSMutableDictionary*)dictionary2
```
YuriyVelichkoPI, line 104: In general, we can merge dictionaries. This path is faster to implement. However, it makes the solution restricted to such things as 
- RTB protocol validation (proper structure, objects, and types)
- Warnings for objects prohibited from merging. 
- Upgrading protocol version. See the item 1 - the validation issue. 

It is not critical, but I would still merge serialized objects.  It is a more reliable approach from the support perspective.

PrebidMobileTests/RenderingTests/Tests/PBMORTBAbstractTest.swift
```diff
@@ -174,6 +174,11 @@ class PBMORTBAbstractTest : XCTestCase {
         pbmORTBBidRequest.test = 2
         
         codeAndDecode(abstract: pbmORTBBidRequest, expectedString: "{\"id\":\"\(uuid)\",\"imp\":[{\"clickbrowser\":0,\"ext\":{\"dlp\":1},\"instl\":0,\"secure\":0}],\"test\":2,\"tmax\":2000}")
+        
+        pbmORTBBidRequest.ortbObject = ["arbitraryparamkey1": "arbitraryparamvalue1", "tmax": 3000, "id": "1231234", "imp": ["hi": "hello"]]
+        
+        codeAndDecode(abstract: pbmORTBBidRequest, expectedString: "{\"arbitraryparamkey1\":\"arbitraryparamvalue1\",\"id\":\"\(uuid)\",\"imp\":[{\"clickbrowser\":0,\"ext\":{\"dlp\":1},\"instl\":0,\"secure\":0}],\"test\":2,\"tmax\":2000}")
```
YuriyVelichkoPI, line 180: We need to cover more use cases here, including incorrect JSONs. Not overriding of reserved objects. Also the custom parameter should be a String but not Dictionary.

YuriyVelichkoPI (changes requested):
Except for the requested changes, I want to ask you to redo the branch and make it clean from changes from the master. For now, I see which changes are related to the parent task and which were merged from the master. But later whey can confuse. 

The best way to keep the history clear is to follow the trunk-based development methodology: https://trunkbaseddevelopment.com/  as we did in mobile repositories. 

It will help any collaborator to see the clear history that is especially important for the open source project.

PR #949, opened 2024-01-11
Stop ViewabilityTimer when ViewForTracking has become nil

### LINKED ISSUE #908: Expose unregisterViewFromTracking

I'm trying to place an ad as a `UITableViewCell` for every 4 contents. It works great by using 
```swift
nativeAd.registerView(view: containerView, clickableViews: [ctaButton])
```
but when I scroll down, the log becomes too noisy because invisible ads keep their viewability timer running.
I feel it's not CPU-friendly when I have more than 10 ads in a list.

Can we expose `unregisterViewFromTracking` in `NativeAd` so I can call it like below?
```swift
func tableView(
    _ tableView: UITableView,
    didEndDisplaying cell: UITableViewCell,
    forRowAt indexPath: IndexPath
) {
    nativeAdForCell(cell).unregisterViewFromTracking()
}

// Or if the cell is never reused, on UITableViewCell deinit
```

I've tried using
```swift
nativeAd.registerView(view: UIView(), clickableViews: [])
```
but the timer keeps running.

jsligh:
Closes #908

### DIFF
--- a/PrebidMobile/AdUnits/Native/NativeAd.swift
+++ b/PrebidMobile/AdUnits/Native/NativeAd.swift
@@ -176,10 +176,14 @@ public class NativeAd: NSObject, CacheExpiryDelegate {
             Log.error("The native Ad is expired, cannot use it for tracking")
             return false
         }
-        viewForTracking = view
-        setupViewabilityTracker()
-        attachGestureRecognizersToNativeView(nativeView: view, withClickableViews: clickableViews)
-        return true
+        if (viewForTracking != nil || impressionHasBeenTracked) {
+            return false
+        } else {
+            viewForTracking = view
+            setupViewabilityTracker()
+            attachGestureRecognizersToNativeView(nativeView: view, withClickableViews: clickableViews)
+            return true
+        }
     }
     
     private func unregisterViewFromTracking() {
@@ -219,6 +223,9 @@ public class NativeAd: NSObject, CacheExpiryDelegate {
                 return
             }
             strongSelf.checkViewability()
+            if (strongSelf.viewForTracking == nil) {
+                timer.invalidate()
+            }
         }
     }
     


PrebidMobile/AdUnits/Native/NativeAd.swift
```diff
@@ -36,7 +36,13 @@ public class NativeAd: NSObject, CacheExpiryDelegate {
     private var viewabilityTimer: Timer?
     private var viewabilityValue = 0
     private var impressionHasBeenTracked = false
-    private weak var viewForTracking: UIView?
+    private weak var viewForTracking: UIView? {
+        didSet {
+            if (viewForTracking == nil) {
+                invalidateTimer(viewabilityTimer)
```
YuriyVelichkoPI, line 42: In this case, we shouldn't provide an "unregister" functionality. Otherwise, the publisher can just not register the view. And that is it. 

In the case of an unregistered view, the impression event may not be tracked when the ad view appears on the screen. It is the biggest disadvantage.  

Instead, I suggest to review the following aspects:
- invalidation of the timer when the impression is tracked. It is already implemented in the method `trackImpression`. So we just need to make sure that it works and the viewability checker timer is stopped after the impression is tracked. 
- review the timer period. For now, it is really small. And I believe it is because of the bug.

The spec for `scheduledTimer` states 

> /// - parameter:  interval  The number of seconds between firings of the timer. If seconds is less than or equal to 0.0, this method chooses the nonnegative value of 0.1 milliseconds instead.

And we call this method with the following parameter:

```
    private static let nativeAdCheckViewabilityForTrackingFrequency = 0.25
```

It means that the code "wants" to fire the timer four times per second, which doesn't seem bad, but instead, it does it every 0.1 milliseconds, which is bad and unnecessary. 

So I think if we reduce the frequency and make sure that the timer stops after tracking the impression event, the behavior will be correct and will not consume CPU. 

cc @shinwan2

YuriyVelichkoPI (changes requested):
See the comment in the review

PrebidMobile/AdUnits/Native/NativeAd.swift
YuriyVelichkoPI: Sorry, @jsligh pointed out that the parameter is correct and corresponds to the spec.

I double-checked the behavior in the Internal Test App with an integration case, `Native Add Feed (In-App).` The timer works as expected - checks the viewability 4 times per second. 

It looks like a normal cadence. 

Now, we should be sure that it stops after the impression is tracked.

shinwan2:
Thank you so much for handling this @YuriyVelichkoPI and @jsligh.

If I may be so bold, below are my use cases and expectations described by unit tests for faster feedback loop.
https://github.com/prebid/prebid-mobile-ios/compare/master...smartnews:prebid-mobile-ios:feature/optimize-view-tracking?expand=1

Let me know if I can be of more help.

YuriyVelichkoPI:
@shinwan2, thank you for the clarification of the use case. 

If I understood correctly, you want to display the same ad in the feed several times. Do you want to do it regardless of the impression event, Or if the ad is tracked the impression, you don't need to display it one more time in the feed?

To support your scenario, I believe you should use some additional container view as a point of viewability tracking and add this whole container to the different cells in the feed. 

The pseudo-code solution: 
```
 let nativeAd = NativeAd.create(cacheId: cacheId!)!
 let adContainer: UIView = createMyNativeAd(from: nativeAd)
 
 nativeAd.registerView(adContainer ...)
 
 cell.addSubview(adContainer)

// Later, after scrolling do the same, the native ad will still track the viewability of the adContainer but not the cell in the feed
cell.addSubview(adContainer)
```

YuriyVelichkoPI:
In any casy after expecting the use case and the current codebase I think that we still need several additional enchancmets regarding the impression tracking of the native ads: 

- Once the `viewabilityTimer` is fired, and the `viewForTracking` is `nil` (destroyed due to the app ligic) - invalidate the timer in order to prevent further useless cycles. 
- The `registerView` function should do nothing if: 
   - the `viewForTracking` is not nil
   - the `impressionHasBeenTracked` is `true`

It will prevent
 - viewability tracking of non existing view
 - mess in timers and tracking flows
 
 Pubilshers in their turn should register the view right before it should be displayed. Even if publisher wants to change the ad view placement they still should do it as fast as possible. So the timer still shouldn't harm so much. Cahing the ad for a long time is impossible (due to the ad expiration policy) and not appreciated in any case. Still waiting for the detailed feedback about the use case.

shinwan2:
@YuriyVelichkoPI 

> Do you want to do it regardless of the impression event

Yes, this one. The ad once assigned to a given slot (e.g. 4th slot on a feed) will always be shown on that same slot (using `UITableViewCell`).

> I believe you should use some additional container view as a point of viewability tracking and add this whole container to the different cells in the feed.

I see, I believe it should be OK to clear the `UIImageView.image` (showing the ad image) to save memory when it's out of view?

YuriyVelichkoPI (approved):
LGTM

Issue #920, opened 2023-09-26
Fix CFNetwork runtime warning on Xcode 15


OlenaPostindustria:
**Describe the bug**
There is a runtime warning that appears on Xcode 15. 

**To Reproduce**
1. Run PrebidDemoSwift using Xcode 15.
2. Run any test case.
3. See runtime warning

<img width="998" alt="image" src="https://github.com/prebid/prebid-mobile-ios/assets/92721590/95c082d1-698b-4a09-acd5-3113036d77c5">

positivefx:
Hi, I am also running into this issue. Is there a status on this fix?

aiFigueiredo:
I stumbled upon this issue as a dependency I require has Prebid as a dependency. 
Would a suitable fix for this be removing L104 in `PrebidServerConnection` as the data is already passed into `uploadTask`?

aiFigueiredo:
Opened #955 to get this fixed

aiFigueiredo:
FYI: PR to address this was merged

PR #955, opened 2024-02-13
Remove data from request as its already contained on the uploadTask

aiFigueiredo:
This PR is to address issue #920 where an error is shown on`session.uploadTask` when the request contains a body.

As the title states, the data should not be required in the request as its already passed to the `uploadTask` method.


aiFigueiredo:
Thanks @jsligh, what do we need to get this merged in and on a new release?

aiFigueiredo:
Thanks @jsligh, appreciate the 👀 and the info.

PR #957, opened 2024-02-23
938 Fixed Device UA being empty on First Bid Request

### LINKED ISSUE #938: device.ua always empty in the first prebid bid request
**Describe the bug**

Hi team,
we've tested the new iOS version 2.1.6, the prebid-mobile version is correct, and the device.ua field no longer includes the app version. However, we still experience an issue, in the first prebid bid request, the user agent is always empty. It seems as though, Prebid sdk is not able to detect the user agent in the first request.

**To Reproduce**

Here is the request:
{ "app": { "bundle": "484115113", "domain": "leboncoin.fr", "ext": { "prebid": { "source": "prebid-mobile", "version": "2.1.6" } }, "name": "Leboncoin", "publisher": { "id": "3296" }, "storeurl": "https:\/\/apps.apple.com\/fr\/app\/leboncoin\/id484115113", "ver": "6.89.0" }, "device": { "connectiontype": 2, "ext": { "atts": 3, "prebid": { "interstitial": { "minheightperc": 1, "minwidthperc": 1 } } }, "h": 844, "ifa": "3F6929E8-CF38-4F99-A17A-72A2902740C7", "language": "fr", "lmt": 0, "make": "Apple", "model": "iPhone", "os": "iOS", "osv": "17.0.3", "pxratio": 3, "ua": "", "w": 390 }, "ext": { "prebid": { "cache": { "bids": {}, "vastxml": {} }, "storedrequest": { "id": "3296" }, "targeting": {} } }, "id": "368AAD93-FB1F-4A75-B137-48926F5F2EBE", "imp": [{ "banner": { "format": [{ "h": 844, "w": 390 }], "pos": 7 }, "clickbrowser": 1, "ext": { "prebid": { "storedrequest": { "id": "30312907" } } }, "id": "A1252508-3C58-4829-A142-5CB17840E609", "instl": 1, "secure": 1 }], "regs": { "ext": { "gdpr": 1 } }, "source": { "tid": "3280A47A-8EE8-4D22-BA07-13746839F0B7" }, "user": { "ext": { "consent": "CPyc0sAPyc0sAAHABAFRDXCsAPLAAHLAAAAAJoNB_G_dTyPi-f59YvtwYQ1P4VQnoyACjgaNAwwJiVLBMI0EhmAIKAHqAAACIBAkICZAAQBlCAHAAAAA4IEAASMMAAAAIRAIIgCAAEAAAmJICABZC5AAAQAQgkwAABUAgAICABsgSDAAAAAAFAAAAAgAAAAAAAAAAAAAQAAAAAAAAgAAAAAAAAAAEAAEABAAAAAAAAAAAAAAAAAEEAQATDQuIAGwJGQmkDCIAACMIAgCgBABRAJCwQAEBKgAEEYACjAAAABFAAAAAAAAEBAAAAAAgAQgAAAAYEAgAAAEAAAAEAgEAAAAACAAABAAAAAEAMAAAAAAgAIAAAIAQAAhAAgAJAgACAAAAgAAAAAAAAAgEAAAAAAAAAAAAAAAAQAxQAGAAIgoDAAMAARBQIAAYAAiCgA" } } }

**Expected behavior**

For all bid requests to correctly include the device.ua field.

Thank you
Ghada.

jsligh:
Fixes #938 

Removes the race condition and uses UIWebView instead of WKWebView.  UIWebView does javascript calls synchronously while WKWebView does javascript asynchronously which is why the race condition is occurring.  While synchronous execution might take an extra second, it prevents a lot of headaches and dispatch_once makes sure this only happens once.  The UserAgent is also cached in the shared PBMUserAgentService.

### DIFF
--- a/PrebidMobile/PrebidMobileRendering/Utilities/PBMUserAgentService.m
+++ b/PrebidMobile/PrebidMobileRendering/Utilities/PBMUserAgentService.m
@@ -79,7 +79,7 @@
             numberOfLoops = [NSNumber numberWithInt:value + 1];
         }
     }
-    return [NSString stringWithFormat:@"%@", self.userAgent];
+    return [NSString stringWithFormat:@"%@", self.userAgent ?: @""];
 }
 
 #pragma mark - Private Methods


PrebidMobile/PrebidMobileRendering/Utilities/PBMUserAgentService.m
```diff
@@ -30,7 +30,7 @@
 
 @interface PBMUserAgentService()
 
-@property (nonatomic, strong) WKWebView *webView;
+@property (nonatomic, strong) UIWebView *webView;
```
YuriyVelichkoPI, line 33: We can't use UIWebView. It is deprecated. It's usage may block the submission of the app into App Store.

YuriyVelichkoPI (changes requested):

PrebidMobile/PrebidMobileRendering/Utilities/PBMUserAgentService.m
```diff
@@ -68,6 +70,13 @@ - (instancetype)init {
 #pragma mark - Public Methods
 
 - (nonnull NSString *)getFullUserAgent {
+    if (self.userAgent == nil) {
+        self.uaSemaphore = dispatch_semaphore_create(0);
```
YuriyVelichkoPI, line 74: If two ad units call this method at the same time before ua is obtained, the semaphore will be repeated, with unpredictable consequences

YuriyVelichkoPI: I'm not sure about using semaphore here. It is a tool dedicated to multithreading access to resources. But we need just to lock the process for any resource consumer. So, for me, using NSLock will be much more convenient here. It will let us avoid using the while loop for waiting for the resource. But we still need to be sure of proper unlocking.

YuriyVelichkoPI (changes requested):

PrebidMobile/PrebidMobileRendering/Utilities/PBMUserAgentService.m
YuriyVelichkoPI: NSLock doesn't lock anything itself. But yes, the requirement that lock() and unlock() should be called on the same thread makes it hardly usable in this situation. 

The current implementation looks good. However, this [loop](dispatch_semaphore_wait) looks potentially endless. In the case when the semaphore will be actually incremented before the start of the loop, the wait will always be finished with a timeout. Does it make sense to add an extra check for userAgent inside the loop or limit the number of possible loops?

Issue #976, opened 2024-05-09
The app is frozen because it is in a loop


DevOpsAppsIOL:
**Describe the bug**
After updating the Prebid library from version 2.1.6 to major versions, when requested by the adv, the app is frozen.

**The problem**
I'm stuck in this loop:
```
- (nonnull NSString *)getFullUserAgent {
    if (self.userAgent == nil) {
        NSNumber *numberOfLoops = [NSNumber numberWithInt: 0];
        while (dispatch_semaphore_wait(self.uaSemaphore, DISPATCH_TIME_NOW + 1) && [numberOfLoops intValue] < 100) {
            [[NSRunLoop currentRunLoop] runMode:NSDefaultRunLoopMode
                                     beforeDate:[NSDate dateWithTimeIntervalSinceNow:10]];
            int value = [numberOfLoops intValue];
            numberOfLoops = [NSNumber numberWithInt:value + 1];
        }
    }
    return [NSString stringWithFormat:@"%@", self.userAgent ?: @""];
}
```

In older versions there was no while loop. but a simple:
```
- (nonnull NSString *)getFullUserAgent {
    return [NSString stringWithFormat:@"%@", self.userAgent];
}
```

**Expected behavior**
I expect there are no while loops blocking the entire application.

hkimmgni:
@DevOpsAppsIOL , what version are you seeing this issue on? is it v2.2.1?

vgorloff:
Same issue here. App just stuck and does nothing. Loop jus blocks UI completely.
Would be nice if you will transform code to avoid synchronous operations, such as semaphores.

YuriyVelichkoPI:
@OlenaPostindustria, could you please take a look at the issue when you have time?

DevOpsAppsIOL:
> @DevOpsAppsIOL , what version are you seeing this issue on? is it v2.2.1?
Hi @jsligh,
I encountered the problem in versions 2.2.0 and 2.2.1. 
I'm currently using 2.1.6 which doesn't give me any problems.
I'm waiting for a new version with the fix.
Thank you.

YuriyVelichkoPI:
Hi @DevOpsAppsIOL et, al. 

Please review the alternative implementation that should fix the issue: https://github.com/prebid/prebid-mobile-ios/pull/983 
The ua will still be present in each request. 

This [comment](https://github.com/prebid/prebid-mobile-ios/pull/983#issuecomment-2122626417) shows the time of fetching the ua.

[...]

PR #983, opened 2024-05-21
Fix device UA being empty on first bid request without blocking application

OlenaPostindustria:
#976

### DIFF
--- a/PrebidMobile/PrebidMobileRendering/Utilities/UserAgentService.swift
+++ b/PrebidMobile/PrebidMobileRendering/Utilities/UserAgentService.swift
@@ -27,27 +27,29 @@ public class UserAgentService: NSObject {
     override init() {
         super.init()
         
-        generateUserAgent()
+        fetchUserAgent()
     }
     
-    public func generateUserAgent(completion: (() -> Void)? = nil) {
+    public func fetchUserAgent(completion: ((String) -> Void)? = nil) {
         // user agent has been already generated
         guard userAgent.isEmpty else {
-            completion?()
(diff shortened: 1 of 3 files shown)


PrebidMobile/PrebidMobileRendering/Utilities/UserAgentService.swift
```diff
@@ -0,0 +1,54 @@
...
+    
+    private let webView = WKWebView()
+    
+    override init() {
+        super.init()
+        
+        generateUserAgent()
+    }
+    
+    public func generateUserAgent(completion: (() -> Void)? = nil) {
```
YuriyVelichkoPI, line 33: Let's call this method `fetchUserAgent` . 

Also looking at the signature, I expect that the UA string will be a parameter of the completion handler. I see that it is not applicable for the current usage of this method, but please consider such interface changes.

YuriyVelichkoPI (changes requested):
Could you also measure the average, minimum, and maximum time for extracting the user agent string, as well as the number of errors per 1 million attempts from different web view instances?

The current solution does not have a timeout option, so we need to ensure that the value will be fetched reliably.

OlenaPostindustria:
> Could you also measure the average, minimum, and maximum time for extracting the user agent string, as well as the number of errors per 1 million attempts from different web view instances?
> 
> The current solution does not have a timeout option, so we need to ensure that the value will be fetched reliably.

min - 0.0004010200500488281 s
max - 0.521765947341919 s
average - 0.0019949429631233215 s
number of errors per 1 million attempts - 2

Issue #966, opened 2024-04-12
Crash app with ads redirect to "about:blank"


TranNghiaHiep:
**Describe the bug**
When try with ads click redirect to "about:blank" app crash with error: "*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'The specified URL has an unsupported scheme. Only HTTP and HTTPS URLs are supported.'"

TranNghiaHiep:
I tried adding a code snippet to check the URL scheme before redirecting, and it works. Will this have any impact?

Line 333 in file PBMWebView.m

if ([[url scheme] isEqualToString:@"http"] || [[url scheme] isEqualToString:@"https"]) {
    [self.delegate webView:self receivedClickthroughLink:url];
}

PR #987, opened 2024-05-27
Fix crash in SFSafariViewController

weibel:
During click handling an attempt to open SFSafariViewController could result in a crash.
`The specified URL has an unsupported scheme. Only HTTP and HTTPS URLs are supported.`

The check is added to the section of PBMSafariVCOpener where some other similar checks are carried out.

Fixes to https://github.com/prebid/prebid-mobile-ios/issues/966

### DIFF
--- a/PrebidMobile/PrebidMobileRendering/Prebid/PBMCore/PBMSafariVCOpener.m
+++ b/PrebidMobile/PrebidMobileRendering/Prebid/PBMCore/PBMSafariVCOpener.m
@@ -112,6 +112,11 @@
         return NO;
     }
      
+    if (!([strURLscheme isEqualToString:@"http"] || [strURLscheme isEqualToString:@"https"])) {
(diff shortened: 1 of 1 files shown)


PrebidMobile/PrebidMobileRendering/Prebid/PBMCore/PBMSafariVCOpener.m
```diff
@@ -112,6 +112,11 @@ - (BOOL)openURL:(NSURL *)url onClickthroughExitBlock:(nullable PBMVoidBlock)onCl
         return NO;
     }
      
+    if (!([strURLscheme isEqualToString:@"http"] || [strURLscheme isEqualToString:@"https"])) {
```
niksuprunov, line 115: if (([strURLscheme isEqualToString:@"http"] )) should suffice I think

niksuprunov (changes requested):

Issue #1012, opened 2024-06-18
Prebid SDK iOS does not send OpenRTB 'hwv' object


takenorim:
**Describe the bug**
I found Prebid SDK iOS does not send H/W information which is corresponding to `hwv` in terms of OpenRTB. 

**To Reproduce**
Use Charles iOS app to capture traffic from actual iPhone devices.

**Expected behavior**
`/device.hwv` is sending to Prebid Server.

**Additional context**
It looks the Prebid SDK iOS tries to read hardware information by the following legacy logic. But, this does not work (at least my environment). Can someone confirm it with iPhone/iPad devices, please?

- https://github.com/prebid/prebid-mobile-ios/blob/4020716722b7a73b45abb91c869cd6b0af12f046/PrebidMobile/PrebidMobileRendering/Utilities/PBMDeviceAccessManager.m#L83-L92

The following is Apple's developer guide, I'm not sure whether `hw.machine` still works as expected at recent iOS versions and iPhone/iPad devices.
- https://developer.apple.com/documentation/kernel/1387446-sysctlbyname
- https://developer.apple.com/documentation/kernel/1387446-sysctlbyname/determining_system_capabilities

OlenaPostindustria:
Hi, @takenorim. 

It turned out that the `hwv` field was not added to the request. I made a [PR](https://github.com/prebid/prebid-mobile-ios/pull/1013) to fix this issue. The detection should work fine.

YuriyVelichkoPI:
@jsligh, @OlenaPostindustria can we close this issue?

OlenaPostindustria:
The respective [PR](https://github.com/prebid/prebid-mobile-ios/pull/1013) was merged, so I assume we can close this issue.

Issue #962, opened 2024-04-03
Some Prebid SDK unit tests fail specifically on iOS 17+ 


OlenaPostindustria:
I'm running `PrebidMobileTests` on iOS 17+ and some tests fail(see attached screenshot). But if I run them on iOS version < 17, the tests successfully pass.

Additional info about my environment: 
- macOS 14.2.1
- Xcode 15.3

![image](https://github.com/prebid/prebid-mobile-ios/assets/92721590/33b69550-e267-45ce-8c73-3a1ca9e36ce9)

PR #1013, opened 2024-06-18
Add `hwv` to bid request

OlenaPostindustria:
#1012


OlenaPostindustria:
@jsligh Looks like the new GMA SDK version requires a higher version of Xcode so I had to update Circle CI config. But as I updated the version of Xcode, I ran into another problem that I described in this issue - https://github.com/prebid/prebid-mobile-ios/issues/962. In short, some tests fail specifically on iOS 17+. I'll try to take a look at this this week.

PR #1026, opened 2024-07-16
Custom logger implementation on iOS

### LINKED ISSUE #1023: setCustomLogger implementation from Android Library is not present in iOS library
Android has this method in the global Parameters section of the documentation ->
image

that can be used as so (it's an example) -> 

```
PrebidMobile.setCustomLogger(object : LogUtil.PrebidLogger
{
	override fun println(messagePriority: Int, tag: String, message: String)
	{
		when (messagePriority)
		{
			org.prebid.mobile.LogUtil.NONE    -> CustomLoggerExample.none(tag, message)
			org.prebid.mobile.LogUtil.VERBOSE -> CustomLoggerExample.v(tag, message)
			org.prebid.mobile.LogUtil.DEBUG   -> CustomLoggerExample.d(tag, message)
			org.prebid.mobile.LogUtil.INFO    -> CustomLoggerExample.i(tag, message)
			org.prebid.mobile.LogUtil.WARN    -> CustomLoggerExample.w(tag, message)
			org.prebid.mobile.LogUtil.ERROR   -> CustomLoggerExample.e(tag, message)
			org.prebid.mobile.LogUtil.ASSERT  -> CustomLoggerExample.e(tag, message)
			else            -> CustomLoggerExample.unkown(tag, message)
		}
	}
	
	override fun e(tag: String, message: String, throwable: Throwable)
	{
		CustomLoggerExample.e(tag, throwable)
	}
})
```

Have the same method with same capabilities

jsligh:
Fixes #1023

### DIFF
--- a/PrebidMobile/Logging/Log.swift
+++ b/PrebidMobile/Logging/Log.swift
@@ -15,78 +15,90 @@
 
 import Foundation
 
+/// This class serves as the central point for all logging operations within the SDK.
+/// It allows for categorized logging based on severity levels (e.g., error, warning, debug) and offers options for both console and file-based logging.
+/// It also provides the ability to set third-party logger.
 @objc(PBMLog) @objcMembers
 public class Log: NSObject {
 
     // MARK: - Public properties
-    public static var logger: PrebidLogger? = SDKConsoleLogger()
-    public static var dateFormat = "yyyy-MM-dd hh:mm:ssSSS"
-    public static var dateFormatter: DateFormatter {
-        let formatter = DateFormatter()
-        formatter.dateFormat = dateFormat
-        formatter.locale = Locale.current
-        formatter.timeZone = TimeZone.current
-        return formatter
-    }
     
-    public static func setCustomLogger(customLogger: PrebidLogger?) {
-        logger = customLogger
-    }
-    
+    /// The current logging level. Only messages at this level or higher will be logged.
     public static var logLevel: LogLevel = .debug
+    
+    /// Indicates whether logs should also be saved to a file.
     public static var logToFile = false
+    
+    /// Sets a custom logger to handle log messages.
+    public static func setCustomLogger(_ logger: PrebidLogger) {
+        self.logger = logger
+    }
 
     public static func error(_ object: Any, filename: String = #file, line: Int = #line, function: String = #function) {
-        log(object, logLevel: .error, filename: filename, line: line, function: function)
-        logger?.error(object, filename: filename, line: line, function: function)
+        logger.error(object, filename: filename, line: line, function: function)
     }
 
     public static func info(_ object: Any, filename: String = #file, line: Int = #line, function: String = #function) {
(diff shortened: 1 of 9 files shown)


PrebidMobile/Logging/Log.swift
```diff
@@ -33,38 +33,70 @@ public class Log: NSObject {
     public static var logToFile = false
 
     public static func error(_ object: Any, filename: String = #file, line: Int = #line, function: String = #function) {
-        log(object, logLevel: .error, filename: filename, line: line, function: function)
+        if (customLogger == nil) {
```
YuriyVelichkoPI, line 36: In my opinion, this kind of functionality should be implemented in the following way: 

1. Introduce the class `SDKConsoleLogger` which implements `PrebidLogger` protocol
2. Introduce internal property `var logger:  PrebidLogger? = SDKConsoleLogger()`
3. Introduce a custom setter for the `logger` so pubs can set up own one. 
4. in all functions of the `Log` class just use `logger?.dosomething( .... )`

It will alow to avoid `if ... else` and seamsly add another SDK loggers like `SDKFileLogger` if needed.

PrebidMobile/Logging/Log.swift
```diff
@@ -29,35 +29,46 @@ public class Log: NSObject {
...
     
+    public static func setCustomLogger(customLogger: PrebidLogger?) {
+        logger = customLogger
+    }
+    
     public static var logLevel: LogLevel = .debug
     public static var logToFile = false
 
     public static func error(_ object: Any, filename: String = #file, line: Int = #line, function: String = #function) {
         log(object, logLevel: .error, filename: filename, line: line, function: function)
```
YuriyVelichkoPI, line 40: this code should be in the respected function of the `SDKConsoleLogger`. 

Just move it to `error(...)` function in this case.

YuriyVelichkoPI: You should move 

```
 log(object, logLevel: .error, filename: filename, line: line, function: function)
```

into  `SDKConsoleLogger.error(_ object: Any, filename: String, line: Int, function: String)` function. 

And the same for all other functions. In this case the SDK's logging functionality will be encapsulated in `SDKConsoleLogger`. And published can replace it with a custom one.

PrebidMobile/Logging/Log.swift
```diff
@@ -29,35 +29,46 @@ public class Log: NSObject {
...
+        logger?.error(object, filename: filename, line: line, function: function)
     }
 
     public static func info(_ object: Any, filename: String = #file, line: Int = #line, function: String = #function) {
         log(object, logLevel: .info, filename: filename, line: line, function: function)
```
YuriyVelichkoPI, line 45: Now, you should remove system fogging from the Log class. Otherwise, the same messages will be logged twice.

YuriyVelichkoPI (approved):
LGTM

PR #1027, opened 2024-07-16
Added support for privacyUrl

### LINKED ISSUE #1014: Add support for privacy Url
Implement https://github.com/prebid/prebid-mobile-android/pull/754 on iOS

jsligh:
Fixes #1014

### DIFF
--- a/EventHandlers/PrebidMobileAdMobAdapters/Sources/PrebidAdMobNativeAdapter.swift
+++ b/EventHandlers/PrebidMobileAdMobAdapters/Sources/PrebidAdMobNativeAdapter.swift
@@ -45,15 +45,6 @@ public class PrebidAdMobNativeAdapter:
     
     public var store: String?
     
-    public var privacyUrl: String? {
-        set {
-            prebidNativeAd?.nativeAdMarkup?.privacy = newValue
-        }
-        get {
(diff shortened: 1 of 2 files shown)


EventHandlers/PrebidMobileAdMobAdapters/Sources/PrebidAdMobNativeAdapter.swift
```diff
@@ -45,6 +45,15 @@ public class PrebidAdMobNativeAdapter:
     
     public var store: String?
     
+    public var privacyUrl: String? {
```
YuriyVelichkoPI, line 48: This property doesn't make any sense, because it is absent in `GADMediatedUnifiedNativeAd`. So the GMA SDK won't use it. 

The sense of the request in the referenced ticket was just about adding the property to `NativeAd` class. 

```
 public var privacyUrl: String? {
        set {
            nativeAdMarkup?.privacy = newValue
        }
        get {
            nativeAdMarkup?.privacy
        }
    }
```

So publishers can use `nativeAd.privacyUrl` instead of `nativeAd.nativeAdMarkup?.privacy`.

YuriyVelichkoPI (approved):
LGTM

mdanylov-sigma (approved):

PR #1034, opened 2024-08-05
Generate PrebidMobile Documentation

OlenaPostindustria:


.github/workflows/upload-docs-ios.yml
```diff
@@ -0,0 +1,43 @@
...
+        gem install jazzy
+    - name: Generate docs
+      run: |
+        jazzy
+    - name: Checkout docs branch
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+        ref: docs
+        clean: false
```
YuriyVelichkoPI, line 28: Looks like this command doesn't work as expected. 

In the case of updating the existing doc the fetching of the `docs` branch invalidates local changes. So there is nothing to commit after that. 

It does work when the `docs` brach doesn't have the `/docs` folder. But once the folder is there: 
```
Your branch is up to date with 'origin/docs'.
nothing to commit, working tree clean
```

We need to fix it; The docs should be updated automatically. 

As an alternative, we can try to merge the `release-X.X.X` branch into `docs` and build the docs right there.

.github/workflows/upload-docs-ios.yml
```diff
@@ -0,0 +1,51 @@
...
+    - name: Commit documentation changes
+      run: |
+          if [ "$(git status --porcelain | wc -l)" -gt 0 ]; then
+            git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
+            git config --local user.name "github-actions[bot]"
+            git status
+            TAG_NAME=${{ github.ref }}
+            TAG_NAME=${TAG_NAME#refs/tags/}
+            echo "Current tag: $TAG_NAME"
+            rm -rf docs
```
bretg, line 39: Same comment as on the android side... would like to see some more defensive coding here -- check the size of `docs_new` before removing docs. Thanks.

bretg (changes requested):

OlenaPostindustria:
> Does this generate docs on every commit, on merge to master, or on release? I would think that on-release is what we'd want.

Documentation generation starts when a tag is pushed(f.e., 1.0.0), so it happens on release.

YuriyVelichkoPI (approved):
LGTM

bretg (approved):
My review was limited to the workflow script. My take is this can be released when ready. We're not going to link to the results until after the next SDK release after confirming the release did the right thing to the documentation repo.

thanks!

PR #1035, opened 2024-08-07
Implement Event Delegate

kimar:
# Rationale

We’re developing an analytics backend which depends on the OpenRTB Request and Response data from the Prebid Mobile SDK. 

To achieve this we’ve implemented a simple delegate which will be called with the request and response payloads once the HTTP request to the Prebid Server has finished.

# Next steps

Once this PR has been approved we'd take it as a blueprint for the Prebid Mobile Android SDK and implement it in a similar manner.

### DIFF
--- a/PrebidMobile/PrebidMobileRendering/Prebid/PBMCore/PBMBidRequester.m
+++ b/PrebidMobile/PrebidMobileRendering/Prebid/PBMCore/PBMBidRequester.m
@@ -159,7 +159,7 @@
         }
         
         completion(bidResponse, trasformationError);
-        [Prebid.shared callEventDelegate_prebidBidRequestDidFinishWithRequestData:rtbRequestData 
+        [Prebid.shared callEventDelegateAsync_prebidBidRequestDidFinishWithRequestData:rtbRequestData 
                                                                      responseData:serverResponse.rawData];
     }];
 }
--- a/PrebidMobile/PrebidMobileRendering/Prebid/PBMCore/PBMBidRequester.m
+++ b/PrebidMobile/PrebidMobileRendering/Prebid/PBMCore/PBMBidRequester.m
@@ -94,10 +94,12 @@
     
     const NSTimeInterval postTimeout = (dynamicTimeout_onRead ? dynamicTimeout_onRead.doubleValue : (rawTimeoutMS_onRead / 1000.0));
     
+    NSData *rtbRequestData = [requestString dataUsingEncoding:NSUTF8StringEncoding];
+    
     @weakify(self);
     NSDate * const requestDate = [NSDate date];
     [self.connection post:requestServerURL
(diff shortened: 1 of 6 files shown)


kimar:
Hello @jsligh the intention of using this delegate is for consumers of the SDK to have a plug-and-play way of doing further analytics in terms of the study mentioned by @steffenmllr.

Without this delegate quite a bit more manual work would be required to extract the relevant data, risking also incorrect data by inconsistent implementations.

I'm afraid a fork would not be very appealing for integrators, would you see any chance (even if this would require an alternative implementation) to get this (or a similar solution) merged?

YuriyVelichkoPI:
Hi @kimar, @steffenmllr, this PR was raised during yesterday's prebid mobile PMC meeting. It looks like it is stuck in the backlog, mainly due to miscommunication and a lack of resources. 

We had some concerns initially, but this functionality doesn't break the Prebid approach. Similar callbacks are present in js.  

So, we are going to review it one more time and include it in the next release.

Would you be able to introduce a similar PR for Android SDK?

PrebidMobile/PrebidMobileRendering/Prebid/PBMCore/PBMBidRequester.m
```diff
@@ -157,6 +159,8 @@ - (void)makeRequestWithCompletion:(void (^)(BidResponse *, NSError *))completion
         }
         
         completion(bidResponse, trasformationError);
+        [Prebid.shared callEventDelegate_prebidBidRequestDidFinishWithRequestData:rtbRequestData
```
YuriyVelichkoPI, line 162: I have a concern here that this call can lead to blocking behavior due to the client's implementation. Can we call this delegate asynchronously?  

The async call can be implemented inside `callEventDelegate_prebidBidRequestDidFinishWithRequestData`. In this case the method name should reflect the async behaviour. For instance, `callEventDelegateAsync_prebidBidRequestDidFinishWithRequestData`. 

Open question: on which thread should methods of `PrebidEventDelegate` be called? It shouldn't be a main thread.

YuriyVelichkoPI (changes requested):

ValentinPostindustria:
Hello. We rewrote the [PR](https://github.com/prebid/prebid-mobile-android/pull/814). Now the callback returns the request and response with the default `JSONObject` type.

YuriyVelichkoPI (approved):
LGTM

YuriyVelichkoPI:
> @OlenaPostindustria thanks for CI Fix, sorry I did not have an issues locally.
> 
> Could you please let us know when the next release will be? We would love to retire our fork as soon as possible :)

Hi @steffenmllr, once SharedID is finalized, we'll make an SDK release. I think it should be done this or early next week.

PR #1053, opened 2024-09-19
fix: make sure gppSID in InternalUserConsentDataManager is nil

Macjon:
Before this fix the gppSID from the userdefaults could be an empty string `""`. This PR fixes this and forces the gppSID variable to be nil.

We had this with a library that handles all our consents. We need this fix to use prebid again.


OlenaPostindustria (approved):
LGTM

YuriyVelichkoPI (approved):

Issue #693, opened 2022-10-27
[DRAFT] Public policy for UI/UX appearance and behaviour


YuriyVelichkoPI:
## Motivation

Prebid SDK has many stakeholders. Each of them may want to see the UI/UX of the Rendering API in a particular shape. It can lead to the pull-and-push changes for to the add appearance. 

In order to avoid it and keep everyone on the same page, we should provide a doc with a description of all UI properties and UX behavior so stakeholders can study it first. Then if some changes are still needed, they should rice the discussion with the community. If proposed changes are good for everyone, they should be reflected in the guide and implemented in the codebase on both platforms. 

## Description

YuriyVelichkoPI:
Yes. It aligned well with 3.0 UX items.

PR #1055, opened 2024-10-08
Enable landscape support and autorotation for device orientation

positivefx:


YuriyVelichkoPI:
These changes seem reasonable, but on the other hand, they will break the existing UX, which can harm publishers who rely on the current behavior. 

UI/UX is very important, but since the SDK is the same for everyone, the behavior should be captured in the public policy and modified only after the committee agreement.

We have an old issue to introduce this policy, but for now, it is on hold: https://github.com/prebid/prebid-mobile-ios/issues/693

Issue #1057, opened 2024-10-11
Prebid SDK support for SharedID


bretg:
As more of the ecosystem supports SharedID, it's time to consider adding support for it to PBSDK.

Proposal:

1. Create a way for app developers to opt into having Prebid SDK generate a shared ID. e.g. Prebid global class function createSharedId()
2. This function would need to check for permission:
    1. If LimitAdTracking (LMT) is set, there's no permission. Remove any previously stored SharedID value.
    2. If GDPR is in scope and there's not Purpose 1 consent, remove any previously stored SharedID value. See https://github.com/prebid/prebid-mobile-ios/issues/375, involving gdprApplies and deviceAccessConsent.
    3. On iOS, if [ATTS](https://github.com/prebid/prebid-mobile-ios/issues/965) is other than 3, remove any previously stored SharedID value.
4. If allowed to create SharedID and it doesn't already exist in device local storage, generate a UUID as per generateUUID in https://github.com/prebid/Prebid.js/blob/master/src/utils.js and stored it.
5. When building the ORTB, check device local storage and append the following ORTB object to the user.eids array:

```
{
  "source": "pubcid.org",
   "uids": [{
            "id": 'SHAREDID_VALUE',
            "atype": 1
        }]
}
```

jdwieland8282:
Would Yahoo offer a legal opinion of how setting a SharediD on the device would comply with the various regulatory environments, ex. gdpr, cpra, etc..

bretg:
Discussed in the Identity committee - flipped to ready-for-dev

bretg:
This was discussed in a Prebid Mobile committee meeting. There's some skepticism about the value of SharedID -- in iOS, if ATTS is 3, then bidders can have the Device ID. If that's the only scenario where we're setting SharedID, then it doesn't bring much incremental value.

Here are the ATTS values:

0 = not determined
1 = restricted
2 = denied
3 = authorized

Looking at [Apple's ATTS page](https://developer.apple.com/app-store/user-privacy-and-data-use/), seems that it's about "access their device’s advertising identifier". So requiring ATTS 3 for SharedID may be overkill. However, I can't find what "restricted" means. 

But in any case, even if we stay with the ATTS 3 requirement, perhaps there's still value in the Android environment?

bretg:
[...]

PR #1082, opened 2025-01-09
Generate SharedId and send in auction requests

jclou:
I based the implementation on this proposal #1057 with a few changes:
1. The consent checks and id generation/deletion is done lazily when the `.sharedId` property is accessed. This ensures that the latest consents are always checked before usage. However we still want to give app developers control over whether sharedId is used, so I added the flag `sharedIdEnabled` that developers need to set before we generate/send SharedIds. Maybe we want to also allow app developers to clear a previously generated id, but I did not implement this.
1. Since ATT deprecates and replaces LMT, we only need to check LMT as a fallback for older versions of iOS where ATT is not yet available (e.g. iOS 13 and below).


PrebidMobile/ConfigurationAndTargeting/Targeting.swift
```diff
@@ -255,6 +257,45 @@ public class Targeting: NSObject {
         return transformedUserIdArray.isEmpty ? nil : transformedUserIdArray
     }
     
+    // MARK: - SharedId
+    
+    public var sharedIdEnabled: Bool = false
+    
+    public var sharedId: ExternalUserId? {
+        let sharedIdAllowed: Bool = {
```
YuriyVelichkoPI, line 265: Please encapsulate the code for managing  SharedId in a standalone class. The Targeting class shoud work only as a facade.

YuriyVelichkoPI (changes requested):
Also, please, apply changes regarding to this comment: https://github.com/prebid/prebid-mobile-ios/issues/1057#issuecomment-2590886818

PrebidMobile/ConfigurationAndTargeting/Targeting.swift
```diff
@@ -255,6 +257,45 @@ public class Targeting: NSObject {
         return transformedUserIdArray.isEmpty ? nil : transformedUserIdArray
     }
     
+    // MARK: - SharedId
+    
+    public var sharedIdEnabled: Bool = false
```
YuriyVelichkoPI, line 262: Please document the public properties and methods according to this guide: https://www.swift.org/documentation/docc/writing-symbol-documentation-in-your-source-files

jclou:
Updated based on PR comments:
- Removed ATT/LMT checks since ids are not shared between different apps on the same device
- Enabled sharedId to be used with in-memory persistence when local storage is not allowed
- Added documentation for public API
- Moved implementation into standalone SharedId class

PrebidMobile/ConfigurationAndTargeting/Targeting.swift
```diff
@@ -255,6 +257,21 @@ public class Targeting: NSObject {
...
+    /// encouraged to consult with their legal team before enabling this feature. See `Targeting.sharedId`.
+    public var sharedIdEnabled: Bool = false
+    
+    /// A Prebid-owned first party identifier
+    ///
+    /// The SharedID remains consistent throughout the current app session. The same id may also persist indefinitely across
+    /// multiple app sessions when it is determined that local storage access is allowed. SharedID values are NOT consistent
+    /// across different apps on the same device.
+    public var sharedId: ExternalUserId {
+        SharedId.sharedInstance.identifier
```
YuriyVelichkoPI, line 272: It looks like this method will generate and return sharedId event if sharedIdEnabled is false.

Let's make it consistent: 
- if `sharedIdEnabled` is false the `sharedId` should be `nil`. 
- if publisher sets Targeting.sharedId = false, the SDK should clean up UserDefaults. I understand that in real life is hard to imagine the use case when publisher will switch this feature off, but it make sense to consistent solution .

PrebidMobile/StorageUtils.swift
```diff
@@ -36,6 +36,21 @@ class StorageUtils {
         }
     }
     
+    // SharedId
+    static var sharedId: String? {
```
YuriyVelichkoPI, line 40: Please add unit tests for this function.

PrebidMobile/ConfigurationAndTargeting/SharedId.swift
```diff
@@ -0,0 +1,54 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+  */
+
+import Foundation
+
+class SharedId {
```
YuriyVelichkoPI, line 18: Please add unit tests for this class.

YuriyVelichkoPI (changes requested):

PrebidMobile/ConfigurationAndTargeting/Targeting.swift
jclou: I think for most publishers, the simplest integration of just turning `sharedIdEnabled` on or off is easy and works as expected.`sharedIdEnabled` is checked when constructing the auction request so we won't actually send the id even if it was previously generated when the feature is not enabled. Beyond that I think we should be careful to not do too much and get in the way more sophisticated integrations. 

For instance at Yahoo we actually do decide to turn identifiers on/off randomly before making an auction request so that we can measure an identifier's uplift performance. So we would need a way to temporarily disable the sharedId from being added to the auction request without deleting the existing value and breaking its continuity.

Maybe we can just add a function to reset the sharedId and clarify in the documentation for `sharedId` that `sharedIdEnabled` needs to be turned on for the it to get sent in auction requests?

YuriyVelichkoPI: Thanks for the clarification. This use case makes perfect sense to me. 

In this case, we should make API more verbose about what exactly each method or property does. And document how to use it. 

The main point is when we design the solution in open-source SDK; it should be generic and clear to let other users to implement their use cases using the provided API. 

For this purpose, I suggest the following: 
- Rename `sharedIdEnabled` to `sendSharedId`. It will indicate that the property is responsible only for adding the SharedID to the bid request. The implementation of `sharedId` can stay the same in this way. 
- Add `resetSharedId()` to remove the current SharedID from the app storage. 
- Describe the use case in the doc.

OlenaPostindustria:
Hi, @jclou! This is one of the possible places where you can add unit tests for `sendSharedId` -  https://github.com/prebid/prebid-mobile-ios/blob/master/PrebidMobileTests/RenderingTests/Tests/ParameterBuilderTests/ParameterBuilderServiceTest.swift

YuriyVelichkoPI:
@jclou, sorry if I already asked. Will you or your team provide an Android PR as well?

YuriyVelichkoPI (approved):
LGTM

But should be merged once PMC aligns the approach on the legal side. 

On the last call, @Arjun Taneja wanted to consult with the legal team one more time about the dependency of SharedID on ATT.

Issue #1087, opened 2025-01-16
TargetingInfo not passed in case of no bid


ankit-thanekar007:
There is a behavior difference between Android and iOS TargetingInfo handling. 

The difference is in case of Android, the targetingInfo is passed in the Ad object regardless of whether the bid is winning or not. But in case of iOS it's only passed when the bid is a  winning bid. In our use case, we want to pass any targeting info that we get from our Prebid Server to Ad Server, regardless of whether Prebid won or not.

In case of iOS, I see it happening in this function https://github.com/prebid/prebid-mobile-ios/blob/50120081450a318e97186067304c1d01d22d6521/PrebidMobile/AdUnits/AdUnit.swift#L189

Here it returns prebidDemandNoBids in case of no winning bid

```
        guard let winningBid = bidResponse.winningBid else {
            return .prebidDemandNoBids
        }
```
Due to this it doesn't perform 
`Utils.shared.validateAndAttachKeywords(adObject: adObject, bidResponse: bidResponse)`

https://github.com/prebid/prebid-mobile-ios/blob/50120081450a318e97186067304c1d01d22d6521/PrebidMobile/AdUnits/AdUnit.swift#L198
The above func performs the following 
`adObject.setValue( existingDict, forKey: "customTargeting")` 
which attaches the targetingInfo to the ad object.

**To Reproduce**
- Check the `cust_params` in gam request for a non-winning auction. 

**Expected behavior**
- TargetingInfo should be passed regardless of bid status

YuriyVelichkoPI:
Introduced in Prebid Mobile `2.4.0`

PR #1088, opened 2025-01-16
Update TargetingInfo Behavior 

ankit-thanekar007:
- Add a flag that allows devs to use this feature. 
- If the flag is set, the targeting info will be added to adObject regardless of bid status
- If the flag is not set, it'll retain the current behavior 
- Add tests

@YuriyVelichkoPI : After our discussion, do we need the flag or should this be the default behavior ? 

Fixes Issue : https://github.com/prebid/prebid-mobile-ios/issues/1087

### DIFF
--- a/Example/PrebidDemo/PrebidDemoSwiftUITests/NativeAdsTest.swift
+++ b/Example/PrebidDemo/PrebidDemoSwiftUITests/NativeAdsTest.swift
@@ -34,6 +34,6 @@ class NativeAdsTest: BaseAdsTest {
     }
     
     override func checkAd(testCase: String) {
-        XCTAssert(app.staticTexts[testCase].waitForExistence(timeout: 10),assertFailedMessage(testCase: testCase, reason: "Prebid title is not displayed"))
+        XCTAssert(app.staticTexts["Prebid (Title)"].waitForExistence(timeout: 10),assertFailedMessage(testCase: testCase, reason: "Prebid title is not displayed"))
     }
 }
--- a/Example/PrebidDemo/PrebidDemoSwiftUITests/NativeAdsTest.swift
+++ b/Example/PrebidDemo/PrebidDemoSwiftUITests/NativeAdsTest.swift
@@ -34,6 +34,6 @@ class NativeAdsTest: BaseAdsTest {
     }
     
     override func checkAd(testCase: String) {
-        XCTAssert(app.staticTexts["Prebid (Title)"].waitForExistence(timeout: 10),assertFailedMessage(testCase: testCase, reason: "Prebid title is not displayed"))
+        XCTAssert(app.staticTexts[testCase].waitForExistence(timeout: 10),assertFailedMessage(testCase: testCase, reason: "Prebid title is not displayed"))
     }
 }
--- a/PrebidMobile/AdUnits/AdUnit.swift
+++ b/PrebidMobile/AdUnits/AdUnit.swift
@@ -188,7 +188,7 @@ public class AdUnit: NSObject, DispatcherDelegate {
     
     private func setUp(_ adObject: AnyObject?, with bidResponse: BidResponse) -> ResultCode {
         
-        if Prebid.shared.forceTargetingInfo, let adObject {
+        if let adObject {
             Utils.shared.validateAndAttachKeywords(adObject: adObject, bidResponse: bidResponse)
         }
         
--- a/PrebidMobile/AdUnits/AdUnit.swift
+++ b/PrebidMobile/AdUnits/AdUnit.swift
@@ -186,16 +186,21 @@ public class AdUnit: NSObject, DispatcherDelegate {
         })
     }
     
-    private func setUp(_ adObject: AnyObject?, with bidResponse: BidResponse) -> ResultCode {
+    func setUp(_ adObject: AnyObject?, with bidResponse: BidResponse) -> ResultCode {
         
         if let adObject {
-            Utils.shared.validateAndAttachKeywords(adObject: adObject, bidResponse: bidResponse)
+            //If Bid Present and the bid is winning
+            //OR
+            //If forceSdkChooseWinner == false : Any bid works -- winnning or loosing
+            if bidResponse.winningBid?.isWinning == true || !Targeting.shared.forceSdkToChooseWinner {
+                Utils.shared.validateAndAttachKeywords(adObject: adObject, bidResponse: bidResponse)
+            }
         }
-        
+
         guard let winningBid = bidResponse.winningBid else {
             return .prebidDemandNoBids
         }
-        
+
         if let cacheId = cacheBidIfNeeded(winningBid) {
             bidResponse.addTargetingInfoValue(key: PrebidLocalCacheIdKey, value: cacheId)
         }
--- a/PrebidMobile/AdUnits/AdUnit.swift
+++ b/PrebidMobile/AdUnits/AdUnit.swift
@@ -191,20 +191,15 @@ public class AdUnit: NSObject, DispatcherDelegate {
         if Targeting.shared.forceSdkToChooseWinner {
             Log.error("Breaking change: set Targeting.forceSdkToChooseWinner = false and test your behavior. In the upcoming major release, the SDK will send all targeting keywords to the AdSever, so you should prepare your setup.")
         }
-        
-        if let adObject {
-            //If Winning Bid is present
-            //OR
-            //If forceSdkToChooseWinner == false : Any bid works -- winnning or loosing
-            if bidResponse.winningBid != nil || !Targeting.shared.forceSdkToChooseWinner {
-                Utils.shared.validateAndAttachKeywords(adObject: adObject, bidResponse: bidResponse)
-            }
-        }
 
         guard let winningBid = bidResponse.winningBid else {
             
             //When the new behavior is active
             if !Targeting.shared.forceSdkToChooseWinner {
+                
+                if let adObject {
+                    Utils.shared.validateAndAttachKeywords(adObject: adObject, bidResponse: bidResponse)
+                }
                 //If there are no winning bids, but there are bids the SDK will send back prebidDemandFetchSuccess
                 if let bids = bidResponse.allBids, !bids.isEmpty {
                     return .prebidDemandFetchSuccess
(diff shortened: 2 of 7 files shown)


PrebidMobile/AdUnits/AdUnit.swift
```diff
@@ -187,6 +187,11 @@ public class AdUnit: NSObject, DispatcherDelegate {
     }
     
     private func setUp(_ adObject: AnyObject?, with bidResponse: BidResponse) -> ResultCode {
+        
+        if Prebid.shared.forceTargetingInfo, let adObject {
```
mdanylov-sigma, line 191: Looks like the provided solution will break the existing behaviour.
In case `forceTargetingInfo == false`, which is default behaviour, `bidResponse` will never be attached for winning bid.

ankit-thanekar007: Yeah, thanks for catching that. I removed the flag altogether and by default it'll always add targeting params.

PrebidMobile/AdUnits/AdUnit.swift
```diff
@@ -187,6 +187,11 @@ public class AdUnit: NSObject, DispatcherDelegate {
     }
     
     private func setUp(_ adObject: AnyObject?, with bidResponse: BidResponse) -> ResultCode {
+        
+        if let adObject {
```
YuriyVelichkoPI, line 191: In order to make the solution generic and not break the current behavior, we should implement it according to the following requirements: 

1. The current behavior should remain the same. It means that for the current users SKD should continue to add the targeting keywords only in the case when the bid response has a winning bid. 
2. We should inform users that in the future (3.0 release), this behavior will be changed. As usual, we do it via deprecating the method or property, so the app developers see the compile time warning. But in this case, I don't see such an option. Therefore, we should at least log some error message that the current behavior will be changed soon. Also we should put a respective warning note to the public documentation. 
3. We should add the flag to like `Targeting.forceSdkChooseWinner` which will allow adding targeting keywords regardless of the presence of the winning bid. The default value should be `true`. That was partially done in the initial commit, but it should still be adjusted according to the review comment. If forceSdkChooseWinner is `true`, the keywords should be sent only if the winning bid is present (the current behavior). If forceSdkChooseWinner is `false`,  the keywords should be sent regarding the winning bid. 
4. And the most important. If we start sending info from all bids - SDK shouldn't return `.prebidDemandNoBids` if there are bids in the response, regardless of the presence of the winning bid.  
5. And, of course, we should add unit tests for the new behavior. Also, please ensure that we have unit tests that verify that the `BidResposne.targetingInfo` really contains targeting keywords from all bids in the response.

YuriyVelichkoPI (changes requested):

PrebidMobile/AdUnits/AdUnit.swift
ankit-thanekar007: Hey @YuriyVelichkoPI , For 4
- It's possible to not have a winning bid as it doesn't have hb_pb value. It's just targeting data. 
- In that case we'll be passing the targeting data to GAM, it's not a bid or a winning bid. What should we send in return ? 
- I think 4 should be part of the 3.x release as it's scope is much bigger.

ankit-thanekar007: I've made changes for 1,3 and 5. 
Can you let me know the msg for 2 ? Should it use `Log.warn` ?

PrebidMobileTests/AdUnitTests/AdUnitTests.swift
```diff
@@ -114,6 +114,98 @@ class AdUnitTests: XCTestCase {
         XCTAssertEqual(realBidInfo?.nativeAdCacheId, expectedCacheId)
     }
     
+    //forceSdkToChooseWinner + Winner = Contains Targeting Info
+    func testForcedWinnerAndWinningBid() {
+        //given
+        Targeting.shared.forceSdkToChooseWinner = true
```
OlenaPostindustria, line 120: Please reset all global values to their default settings after the test completes (e.g. in [tearDown](https://developer.apple.com/documentation/xctest/xctest/1500463-teardown)) to prevent any unexpected behaviour.

OlenaPostindustria (changes requested):

ankit-thanekar007:
@YuriyVelichkoPI @OlenaPostindustria  Updated the PR with the fixes for comments + 1 test case fix. Please verify.

Example/PrebidDemo/PrebidDemoSwiftUITests/NativeAdsTest.swift
```diff
@@ -34,6 +34,6 @@ class NativeAdsTest: BaseAdsTest {
     }
     
     override func checkAd(testCase: String) {
-        XCTAssert(app.staticTexts["Prebid (Title)"].waitForExistence(timeout: 10),assertFailedMessage(testCase: testCase, reason: "Prebid title is not displayed"))
+        XCTAssert(app.staticTexts[testCase].waitForExistence(timeout: 10),assertFailedMessage(testCase: testCase, reason: "Prebid title is not displayed"))
```
OlenaPostindustria, line 37: Why is it necessary to modify the test conditions? `testCase` text is always visible on the screen since it’s not part of an ad, so it is not a valid indicator for detecting ad presence.

PrebidMobile/AdUnits/AdUnit.swift
```diff
@@ -186,19 +186,37 @@ public class AdUnit: NSObject, DispatcherDelegate {
...
+                //If there are no winning bids, but there are bids the SDK will send back prebidDemandFetchSuccess
+                if let bids = bidResponse.allBids, !bids.isEmpty {
+                    return .prebidDemandFetchSuccess
+                }
+            }
             return .prebidDemandNoBids
         }
-        
+
         if let cacheId = cacheBidIfNeeded(winningBid) {
```
OlenaPostindustria, line 216: We are missing the addition of `hb_cache_id_local` (`cacheId`) to the `adObject`, which is crucial for certain types of ads (e.g., native in-app).

OlenaPostindustria (changes requested):

Example/PrebidDemo/PrebidDemoSwiftUITests/NativeAdsTest.swift
ankit-thanekar007: During this flow `app.staticTexts` never have `"Prebid (Title)"` they have the `testCase` strings. The tests were failing because it could never find `"Prebid (Title)"` inside `app.staticTexts`.

ankit-thanekar007: These started failing since Yesterday.

OlenaPostindustria: This occurred due to the `hb_cache_id_local` key missing issue causing the test case to stop working.

PrebidMobile/AdUnits/AdUnit.swift
OlenaPostindustria: In this case, the `winningBid` is the bid intended for display. The SDK should cache such a bid and attach the `cacheID` to the `adObject` as `hb_cache_id_local`. Otherwise, it will break the existing functionality.

ankit-thanekar007: The 2 other tests target the old behavior

OlenaPostindustria: Yes, I mean more tests for `hb_cache_id_local`

PrebidMobileTests/TargetingTests.swift
```diff
@@ -122,6 +123,17 @@ class TargetingTests: XCTestCase {
         XCTAssertEqual(locationPrecision, Targeting.shared.locationPrecision)
     }
     
+    func testforceSdkToChooseWinner() {
+        //given
+        let forceSdkToChooseWinner = true
```
OlenaPostindustria, line 128: I think `false` would be more suitable here, as `Targeting.shared.forceSdkToChooseWinner` is true by default.

OlenaPostindustria (changes requested):

PrebidMobileTests/TargetingTests.swift
ankit-thanekar007: Updated

PrebidMobile/AdUnits/AdUnit.swift
ankit-thanekar007: I added 2 cases for native format, please can you verify those. I'm not that versed in the native format from Prebid yet

PrebidMobileTests/AdUnitTests/AdUnitTests.swift
```diff
@@ -114,6 +117,148 @@ class AdUnitTests: XCTestCase {
...
+        let rawWinningBid = PBMBidResponseTransformer.makeNativeValidResponse(bidPrice: 0.75)
+        let jsonDict = rawWinningBid.jsonDict as? NSDictionary
+        let bidResponse = BidResponse(jsonDictionary: jsonDict ?? [:])
+        
+        //when
+        let resultCode = adUnit.setUp(adObject, with: bidResponse)
+        
+        //then
+        XCTAssertNotNil(bidResponse.targetingInfo?[PrebidLocalCacheIdKey])
+        XCTAssertTrue((adObject.allKeys as? [String])?.contains("hb_bidder") ?? false)
```
OlenaPostindustria, line 233: I believe it would be helpful to check if the `hb_cache_id_local` key is attached in the ad object. This applies to both native tests.

OlenaPostindustria (changes requested):

PrebidMobileTests/AdUnitTests/AdUnitTests.swift
ankit-thanekar007: Added

OlenaPostindustria (approved):
LGTM!

ankit-thanekar007:
@YuriyVelichkoPI Can you review and let me know in case of any changes ?

YuriyVelichkoPI (approved):
LGTM

ankit-thanekar007:
@YuriyVelichkoPI / @OlenaPostindustria Can you merge the PR ? I don't have access.

PR #1089, opened 2025-01-20
Fix unit tests after merging #1035 (Implement Event Delegate)

OlenaPostindustria:
After merging  #1035 some tests began to crash in CI/CD.


YuriyVelichkoPI (approved):
LGTM

Issue #1028, opened 2024-07-17
Timeout in bannerAdUnit.fetchDemand() when Prebid.shared.timeoutMillis value is equal to bannerAdUnit.setAutoRefreshMillis value


RobertApikyan:
**Description**

Starting from the second auto refresh update, the `BannerAdUnit.fetchDemand()` method completes with `ResultCode.prebidDemandTimedOut` code when the `Prebid.shared.timeoutMillis` value is equal to what has been set to `bannerAdUnit.setAutoRefreshMillis()`.

**To Reproduce**
Originally, this example is from the Prebid's iOS swift demo app (see [GAMOriginalAPIDisplayBannerViewController](https://github.com/prebid/prebid-mobile-ios/blob/master/Example/PrebidDemo/PrebidDemoSwift/Examples/GAM/OriginalAPI/GAMOriginalAPIDisplayBannerViewController.swift))
It is slightly modified, in the `loadView()` method the `Prebid.shared.timeoutMillis = 30000` is added.

##### Important
1. In the `loadView()` method set the `Prebid.shared.timeoutMillis = 30000`, 
2. Make sure the `bannerAdUnit.setAutoRefreshMillis(time: 30000)` is set to 30000 as well.

It is important to have 30000 set to `Prebid.shared.timeoutMillis` and to `bannerAdUnit.setAutoRefreshMillis`, because the issue reproduces only when both have the same value.

After running the sample, for the first time the `bannerAdUnit.fetchDemand()` completes with `ResultCode.prebidDemandFetchSuccess` and the ad loads, but for the subsequent auto refresh updates the `bannerAdUnit.fetchDemand()` fails with `ResultCode.prebidDemandTimedOut` and the ad fails to load.

##### In the logs

```
Prebid demand fetch for GAM Prebid demand fetch successful // This for the first time
Prebid demand fetch for GAM Prebid demand timedout
Prebid demand fetch for GAM Prebid demand timedout
Prebid demand fetch for GAM Prebid demand timedout 
...
```

```
Prebid demand fetch for GAM Prebid demand timedout
GAM did fail to receive ad with error: Error Domain=com.google.admob Code=1 "Request Error: No ad to show." UserInfo={NSLocalizedDescription=Request Error: No ad to show., gad_response_info=  ** Response Info **
    Response ID: (null)

  ** Loaded Adapter Response **
    (null)

  ** Extras Dictionary **
    {
    }

  ** Mediation line items **
}
```

[...]

PR #1093, opened 2025-01-27
Resolves issue with similar timeout and auto refresh value

mdanylov-sigma:
Issue #1028 

Resolves the corner case, when `timeout` equals to `autoRefreshTime`, which triggers new demand request and previously scheduled dispatch queue task simultaneously.

### DIFF
--- a/PrebidMobile/AdUnits/AdUnit.swift
+++ b/PrebidMobile/AdUnits/AdUnit.swift
@@ -157,7 +157,18 @@ public class AdUnit: NSObject, DispatcherDelegate {
         lastFetchDemandCompletion = completion
         adServerObject = adObject
         
+        
+        let timeoutHandler = DispatchWorkItem {
+            if (!self.didReceiveResponse) {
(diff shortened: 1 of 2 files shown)


PrebidMobile/AdUnits/AdUnit.swift
```diff
@@ -157,7 +157,18 @@ public class AdUnit: NSObject, DispatcherDelegate {
...
+        let timeoutHandler = DispatchWorkItem {
+            if (!self.didReceiveResponse) {
+                self.timeOutSignalSent = true
+                completion(BidInfo(resultCode: .prebidDemandTimedOut))
+                return
+            }
+        }
+        
         bidRequester.requestBids { [weak self] bidResponse, error in
+            timeoutHandler.cancel()
```
OlenaPostindustria, line 170: The solution looks great! Could you please consider adding some unit tests for this?

OlenaPostindustria (approved):
LGTM

Issue #1063, opened 2024-11-12
Support plcmt video parameters


bretg:
**Is your feature request related to a problem? Please describe.**

The IAB has moved away from `placement` to `plcmt` for describing the video placement type. Prebid SDK has support for placement, but not plcmt.

**Describe the solution you'd like**

Add plcmt support for both iOS and Android as described in https://github.com/InteractiveAdvertisingBureau/AdCOM/blob/main/AdCOM%20v1.0%20FINAL.md#list_plcmtsubtypesvideo

YuriyVelichkoPI:
Hi @bretg! Clarification question - should the SDK deprecate and remove in the future support of `placement`? Or do we just need to add the support of `plcmt`? 

If we deprecate and remove `placement` the server should take care of converting the  `plcmt`to `placement` for OpenRTB 2.5 bidders. And I can't find such functionality in PBS PRs for now.

bretg:
> should the SDK deprecate and remove in the future support of placement

No. Many pubs and DPSs still use placement.

> the server should take care of converting the plcmtto placement

Unfortunately this isn't possible -- there's no direct mapping between them.

YuriyVelichkoPI:
Thanks for the confirmation! I'll add respective comments to the PR.

YuriyVelichkoPI:
Introduced in Prebid Mobile 2.4.0

PR #1094, opened 2025-01-27
Support video plcmnt parameter

mdanylov-sigma:
Issue #1063 

- Adds "plcmnt" parameter for video object.
- Deprecates  "placement" property.

### DIFF
--- a/PrebidMobile/AdUnits/Parameters/VideoParameters.swift
+++ b/PrebidMobile/AdUnits/Parameters/VideoParameters.swift
@@ -54,7 +54,6 @@ public class VideoParameters: NSObject {
     public var startDelay: Signals.StartDelay?
     
     /// Placement type for the impression.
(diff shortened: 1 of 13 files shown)


PrebidMobile/AdUnits/Parameters/VideoParameters.swift
```diff
@@ -54,8 +54,12 @@ public class VideoParameters: NSObject {
     public var startDelay: Signals.StartDelay?
     
     /// Placement type for the impression.
+    @available(*, deprecated, message: "This property is deprecated. Please, use plcmnt instead.")
```
YuriyVelichkoPI, line 57: We shouldn't deprecate this property in the SDK. 

The reason is we still should support OpenRTB 2.5. 

The publishers will have to set both properties `placement` and  `plcmnt`.

YuriyVelichkoPI (changes requested):

YuriyVelichkoPI (approved):
LGTM

Issue #946, opened 2024-01-05
Split account from auction settings


bretg:
**Is your feature request related to a problem? Please describe.**

During initialization, the SDK asks for "account". This value is then copied to two places in the ORTB:

- app.publisher.id
- ext.prebid.storedrequest.id

This is sub-optimal because a single account ID might want to have multiple top-level storedrequests defining different timeouts, price granularities, etc.

**Describe the solution you'd like**

I would like an optional parameter for the SDK to separate out account from "auction settings".

e.g.
```
Prebid.init(context, accountID, adServer, host, [auctionSettingsId]);
```

If specified, the auctionSettingsId is used to set ext.prebid.storedrequest.id. If not specified, it's ok to default ext.prebid.storedrequest.id to the account ID.

**Describe alternatives you've considered**

When we have the new ORTB feature defined in https://github.com/prebid/prebid-mobile-ios/issues/877, there will be a workaround where the app code can override ext.prebid.storedrequest.id. But that seems awkward.

YuriyVelichkoPI:
@mdanylov-sigma you should also prepare the docs PR describing new property.

PR #1100, opened 2025-02-04
Auction settings id 

mdanylov-sigma:
Issue #946

### DIFF
--- a/PrebidMobile/AdUnits/AdUnit.swift
+++ b/PrebidMobile/AdUnits/AdUnit.swift
@@ -147,6 +147,12 @@ public class AdUnit: NSObject, DispatcherDelegate {
             return
         }
         
+        if let auctionSettingsId = Prebid.shared.auctionSettingsId,
+           auctionSettingsId.isEmpty || auctionSettingsId.containsOnly(.whitespaces) {
+            completion(BidInfo(resultCode: .prebidInvalidAuctionSettingsId))
+            return
+        }
+        
         if !isInitialFetchDemandCallMade {
             isInitialFetchDemandCallMade = true
             startDispatcher()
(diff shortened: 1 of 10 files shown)


PrebidMobile/AdUnits/AdUnit.swift
```diff
@@ -147,6 +147,12 @@ public class AdUnit: NSObject, DispatcherDelegate {
             return
         }
         
+        if let auctionSettingsId = Prebid.shared.auctionSettingsId,
+           auctionSettingsId.isEmpty || auctionSettingsId.containsOnly(.whitespaces) {
+            completion(BidInfo(resultCode: .prebidInvalidAuctionSettingsId))
+            return
```
OlenaPostindustria, line 153: Is it necessary for the SDK to fail the entire `fetchDemand` process when the optional parameter is invalid? I would suggest logging the error in this case and proceeding with the bid request with the `prebidServerAccountId` if present.

PrebidMobile/PrebidMobileRendering/Prebid/PBMCore/PBMPrebidParameterBuilder.m
```diff
@@ -67,8 +67,13 @@ - (void)buildBidRequest:(nonnull PBMORTBBidRequest *)bidRequest {
     BOOL const isHTML = ([adFormats containsObject:AdFormat.banner]);
     BOOL const isInterstitial = self.adConfiguration.adConfiguration.isInterstitialAd;
     
+    NSString *requestID = self.sdkConfiguration.auctionSettingsId;
```
OlenaPostindustria, line 70: Would be great to have a test when both `auctionSettingsId` and `prebidServerAccountId` are present.

OlenaPostindustria (changes requested):

OlenaPostindustria (approved):
LGTM

PR #1102, opened 2025-02-10
Test plans for PR checks

mdanylov-sigma:
- Creates test plans for unit test scheme.
- Updates Circle CI config to run only quick tests on PR check. Short list of tests runs in ~40 sec
 instead of ~450 sec.
- Adds job to manually run all tests if needed.


YuriyVelichkoPI:
@mdanylov-sigma is it PR ready for review? 

From the changed code, it is not clear whether the `unit-test` is still used in some pipelines.  Will we still run all tests for release?

Issue #1009, opened 2024-06-10
Deprecated adUnit-level App object Methods.


jsligh:
Deprecate the adunit-level setAppContent, getAppContent, clearAppContent, addAppContentData, removeAppContentData, clearAppContentData, addUserData, getUserData, removeUserData, clearUserData, addContextData, addExtData, updateContextData, updateExtData, removeContextData, removeExtData, clearContextData, clearExtData methods. The app object is not at the adunit level -- this design that will not work when we get to supporting multiple imps in a single auction. There are already equivalents of these functions at the global level.

Originally proposed in #977

Issue #1007, opened 2024-06-10
Remove or deprecate userData


jsligh:
Not useful as implemented.  Originally proposed in #977

Issue #1006, opened 2024-06-10
Remove rubicon and appnexus constants


jsligh:
There are 200+ bidders in Prebid Server. It's time to stop favoring the two founding companies.  Originally proposed in #977

Issue #954, opened 2024-01-31
Support Tracking Domain 


YuriyVelichkoPI:
## Background

According to the [changes](https://developer.apple.com/app-store/app-privacy-details/) in Apple's privacy approach, some SDKs should register the Tracking Domain in the [privacy manifest file](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files). 

For now, the Prebid SDK is not on the [list](https://developer.apple.com/support/third-party-SDK-requirements/) of such SDKs. And it doesn't use a centralized server endpoint that can be caught by the [Points of Interest instrument](https://developer.apple.com/documentation/xcode/detecting-when-your-app-contacts-domains-that-may-be-profiling-users).

However, that time may come, or publishers may want to register the PBS endpoint in the privacy manifest proactively. For this case, the SDK should support both tracking and nontracking domains. 

## Objectives

- Introduce function `Prebid.setCustomPrebidServer(url: String, trackingURL: String)`
- If publishers provided `trackingURL` use it in when a user [allowed the app to track](https://developer.apple.com/app-store/user-privacy-and-data-use/). Otherwise - use the `URL` for the bid requests.

bretg:
I'd like to propose that we switch the order here... let's assume the main Prebid Server can track, and offer an additional "non-tracking" version of the Prebid Server URL.

```
Prebid.setCustomPrebidServer(url: String, nonTrackingURL: String)
```

Here's what I think the proposal is:

1. sniff the user's permission settings
2. if tracking is allowed, use the first parameter as the PBS URL
3. if tracking is not allowed, use the second parameter as the PBS URL

I've opened https://github.com/prebid/prebid-server/issues/3753 to track building a feature in Prebid Server that can recognize when it's invoked in "privacy mode", where it will anonymize key aspects as if the LMT flag were set.

bretg:
Discussed with Prebid legal counsel. In short, this general approach is approved.

1. Prebid recommends to app developers that they disclose the "worst case" in their privacy manifest - i.e. define the 5 'CollectedDataType' fields.
2. Define the "main" Prebid Server domain in the NSPrivacyTrackingDomains field.
3. If the user declines to be tracked, we should assume that iOS will block that domain. That's fine.
[...]

PR #1109, opened 2025-02-25
Deprecate methods and properties for 3.0

mdanylov-sigma:
#1009
#1007
#1006 
#954

### DIFF
--- a/PrebidMobile/AdUnits/AdUnit.swift
+++ b/PrebidMobile/AdUnits/AdUnit.swift
@@ -305,26 +305,26 @@ public class AdUnit: NSObject, DispatcherDelegate {
     
     /// This method obtains the ext data keyword & value for adunit targeting.
     /// If the key already exists the value will be appended to the list. No duplicates will be added
-    @available(*, deprecated, message: "This method is deprecated. In the upcoming major release, the method will be removed.")
+    @available(*, deprecated, message: "This method is deprecated. In the upcoming major release, the method will be removed. Please, use Targeting.addAppExtData method instead.")
     public func addExtData(key: String, value: String) {
         adUnitConfig.addExtData(key: key, value: value)
     }
     
     /// This method obtains the ext data keyword & values for adunit targeting
     /// The values if the key already exist will be replaced with the new set of values
-    @available(*, deprecated, message: "This method is deprecated. In the upcoming major release, the method will be removed.")
+    @available(*, deprecated, message: "This method is deprecated. In the upcoming major release, the method will be removed. Please, use Targeting.updateAppExtData method instead.")
     public func updateExtData(key: String, value: Set<String>) {
         adUnitConfig.updateExtData(key: key, value: value)
     }
     
     /// This method allows to remove specific ext data keyword & values set from adunit targeting
-    @available(*, deprecated, message: "This method is deprecated. In the upcoming major release, the method will be removed.")
+    @available(*, deprecated, message: "This method is deprecated. In the upcoming major release, the method will be removed. Please, use Targeting.removeAppExtData method instead.")
     public func removeExtData(forKey: String) {
         adUnitConfig.removeExtData(for: forKey)
     }
     
     /// This method allows to remove all ext data set from adunit targeting
-    @available(*, deprecated, message: "This method is deprecated. In the upcoming major release, the method will be removed.")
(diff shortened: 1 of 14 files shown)


PrebidMobile/AdUnits/AdUnit.swift
```diff
@@ -305,22 +305,26 @@ public class AdUnit: NSObject, DispatcherDelegate {
     
     /// This method obtains the ext data keyword & value for adunit targeting.
     /// If the key already exists the value will be appended to the list. No duplicates will be added
+    @available(*, deprecated, message: "This method is deprecated. In the upcoming major release, the method will be removed.")
```
OlenaPostindustria, line 308: We need to clearly indicate in the message which alternative method should be used in place of the deprecated one. This could be achieved either through a global-level method from the `Targeting` class or by using the `setImpORTBConfig`/`setGlobalORTBConfig` methods.

OlenaPostindustria (changes requested):

PrebidMobile/Host.swift
```diff
@@ -19,9 +19,11 @@ import UIKit
 @objc public enum PrebidHost: Int {
     
     /// URL [https://ib.adnxs.com/openrtb2/prebid](URL)
+    @available(*, deprecated, message: "This property is deprecated. In the upcoming major release, the property will be removed.")
```
OlenaPostindustria, line 22: Maybe we should also deprecate [Prebid.prebidServerHost](https://github.com/prebid/prebid-mobile-ios/blob/9947afd4de926df4608329d8ac208daba7e5f073/PrebidMobile/ConfigurationAndTargeting/Prebid.swift#L92) ? I don't think we'll need it anymore.

PrebidMobile/PrebidMobileRendering/Prebid/Integrations/GAM/InterstitialRenderingAdUnit.swift
```diff
@@ -355,17 +364,20 @@ public class InterstitialRenderingAdUnit: NSObject, BaseInterstitialAdUnitProtoc
     
     /// Adds user data objects.
     /// - Parameter userDataObjects: The user data objects to be added.
+    @available(*, deprecated, message: "This method is deprecated. In the upcoming major release, the method will be removed.")
```
OlenaPostindustria, line 367: What should be used instead of `addUserData` ?

PrebidMobile/AdUnits/AdUnit.swift
```diff
@@ -305,22 +305,26 @@ public class AdUnit: NSObject, DispatcherDelegate {
     
     /// This method obtains the ext data keyword & value for adunit targeting.
     /// If the key already exists the value will be appended to the list. No duplicates will be added
+    @available(*, deprecated, message: "This method is deprecated. In the upcoming major release, the method will be removed. Please, use Targeting.addAppExtData method instead.")
```
OlenaPostindustria, line 308: Are we replacing what goes into `imp[].ext.data`(`addExtData`) with what goes into `app.ext.data`(`Targeting.addAppExtData`)?

PrebidMobile/ConfigurationAndTargeting/Prebid.swift
```diff
@@ -89,6 +89,7 @@ public class Prebid: NSObject {
     // MARK: - Public Properties (Prebid)
     
     /// The host for the Prebid Server.
+    @available(*, deprecated, message: "This property is deprecated. In the upcoming major release, the property will be removed.")
     public var prebidServerHost: PrebidHost = .Custom {
```
YuriyVelichkoPI, line 93: Please review this comment
https://github.com/prebid/prebid-mobile-ios/pull/1112#issuecomment-2709741499

YuriyVelichkoPI: We should not only deprecate the hardcoded values but also provide a new interface for SDK initialization.

YuriyVelichkoPI (changes requested):

PrebidMobile/ConfigurationAndTargeting/Prebid.swift
mdanylov-sigma: @YuriyVelichkoPI just updated initialization interface, let me know if it looks good and may be replicated for Android.

YuriyVelichkoPI (dismissed):
LGTM

OlenaPostindustria (approved):
LGTM

Issue #833, opened 2023-04-24
[DOC] Sync the SDK Config docs on iOS and Android


YuriyVelichkoPI:
## Objectives 

// TODO

PR #1112, opened 2025-02-27
Tracking URL

mdanylov-sigma:
#954 

- Extends `setCustomPrebidServer` method with `nonTrackingURL` parameter.
- Returns the PBS URL based on ATT status.

### DIFF
--- a/PrebidMobile/Host.swift
+++ b/PrebidMobile/Host.swift
@@ -109,7 +109,7 @@ public class Host: NSObject {
     }
     
     public func getHostURL() throws -> String {
-        guard let customHostURL = customHostURL else {
+        guard let customHostURL else {
             throw ErrorCode.prebidServerURLInvalid("")
         }
         return customHostURL.absoluteString
--- a/PrebidMobile/PrebidMobileRendering/AdTypes/AdView/PBMAbstractCreative.m
+++ b/PrebidMobile/PrebidMobileRendering/AdTypes/AdView/PBMAbstractCreative.m
@@ -173,7 +173,7 @@
         self.viewabilityTracker = [[PBMCreativeViewabilityTracker alloc] initWithCreative:self];
     }
     
-    PBMORTBBidExtSkadn * skadnInfo = self.transaction.skadnInfo;
+    PBMORTBBidExtSkadn * skadnInfo = self.transaction.bid.skadn;
     
     BOOL showSKOverlay = !self.creativeModel.hasCompanionAd &&
                          self.creativeModel.adConfiguration.isInterstitialAd &&
--- a/PrebidMobileTests/RenderingTests/Tests/PrebidTest.swift
+++ b/PrebidMobileTests/RenderingTests/Tests/PrebidTest.swift
@@ -179,7 +179,6 @@ class PrebidTest: XCTestCase {
         try host.setHostURL(customTrackingHost, nonTrackingURLString: customNonTrackingHost)
         
         //then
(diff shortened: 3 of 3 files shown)


PrebidMobile/Host.swift
```diff
@@ -41,22 +42,46 @@ import UIKit
 @objcMembers
 public class Host: NSObject {
 
-    private var customHostURL: URL?
+    private var customHostURL: URL? {
+        get {
+            guard #available(iOS 14.0, *) else {
+                return deviceManager.advertisingTrackingEnabled() ? trackingURL : nonTrackingURL
```
OlenaPostindustria, line 48: I noticed that `advertisingTrackingEnabled` currently utilizes the deprecated property `ASIdentifierManager.sharedManager.isAdvertisingTrackingEnabled`. Apple suggests using `AppTrackingTransparency's ATTrackingManager` class. We need to align with current guidelines to ensure everything works as expected.

Reference - https://developer.apple.com/documentation/adsupport/asidentifiermanager/isadvertisingtrackingenabled.

OlenaPostindustria (changes requested):

PrebidMobile/Host.swift
mdanylov-sigma: Since the minimum deployment target is still iOS 11.0, we can't rely solely on `ATTrackingManager`, which is available starting from iOS 14. That's why `ASIdentifierManager` is used to handle all cases.  Are we going to bump the minimum deployment target for 3.0 ?

OlenaPostindustria: No, sorry, accidentally missed the line with availability. Can mark it as resolved.

YuriyVelichkoPI:
I propose to implement this change on top of changes for https://github.com/prebid/prebid-mobile-ios/issues/1006

The idea is the following: 
- Deprecate constants and provide a new initializeSDK method which will accept the host URL as a parameter (both platforms)
- Introduce additional initialize SDK method with `nonTrackingURL` parameter. 

After that, we will have a simple and clear initializtion interface. 

The word `Custom` should be excluded from the initializtion interface since publisher will have to set the URL for any server without favorites. 

If you think it make sense, please revisit the [PR](https://github.com/prebid/prebid-mobile-ios/pull/1109/files#diff-f3b58ac92a63da83acd6962f17b0f062d1e72d78d16f48cdbdf2bce2a7d7947dR93) and make sure that we not only deprecate the usage of old constants but also provide an alternative initialization method where publishers will have to set the server URL directly. Also make sure that current initialization methods are marked as depricated.

YuriyVelichkoPI:
@mdanylov-sigma, since the #833 is merged, please rebase the branch and finalize changes according to the previous comment.

mdanylov-sigma:
> @mdanylov-sigma, since the #833 is merged, please rebase the branch and finalize changes according to the previous comment.

Sorry, I made mistake and implemented changes related to the `Host` class in that PR, that's why there are no significant changes in this PR.
You can review those changes [here](https://github.com/prebid/prebid-mobile-ios/blob/c5793a10d4a5079f5e6dc30bbf6362cbdf940790/PrebidMobile/Host.swift).

YuriyVelichkoPI:
The adding of nonTrackingURL was added in this PR: #1109

YuriyVelichkoPI (approved):
LGTM

OlenaPostindustria (approved):
LGTM

PR #1039, opened 2024-08-21
Add SKOverlay Support in Rendering API

jsligh:
Fixes #853 and #997

PR #1114, opened 2025-03-12
Fix build issues after recent merges

jclou:
- Fixed a build issue merged in from #1039 after changes were made to a property in #1075
- Fixed mismatch between Podfile and Podfile.lock after merging #1075


OlenaPostindustria (approved):
LGTM

PR #1115, opened 2025-03-13
Prebid response object parsing tests

jclou:
As discussed in the Prebid slack, this PR is a first step towards making the JSON parsing code more type safe by utilizing the Swift language's type safety. With these tests in place, we can begin to migrate some data classes and the corresponding parsing code to Swift with confidence that we aren't accidentally breaking something in the process.

I also added a custom file header to the project to insert the apache license comment instead of the default header with the author's name.


PrebidMobile.xcodeproj/xcshareddata/IDETemplateMacros.plist
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>FILEHEADER</key>
+    <string>
+// Copyright 2018-2025 Prebid.org, Inc.
```
OlenaPostindustria, line 7: I suggest using the `___YEAR___` macro here to avoid being strictly tied to the 2025 year.
Reference - https://help.apple.com/xcode/mac/9.0/index.html?localePath=en.lproj#/dev7fe737ce0

PrebidMobileTests/ResponseParsingTests.swift
```diff
@@ -0,0 +1,449 @@
...
+    @objc(compareArray::)
+    func compare(_ entities: [PBMORTBAbstract]?, _ array: [[String : Any]]) -> Bool {
+        guard let entities else {
+            return false
+        }
+        
+        return entities.map { $0.toJsonDictionary() as NSDictionary } == array.map { $0 as NSDictionary }
+    }
+    
+//    func compareJson(lhs: [String : Any], rhs: [String : Any]) {
```
OlenaPostindustria, line 437: Do we still need the commented code, or can it be removed?

OlenaPostindustria (changes requested):

PR #1124, opened 2025-03-31
Playable endcard rendering

mdanylov-sigma:
Modifies WKUserScript injection to render playable endcards after video ad by applying `mraidScript` for main frame only.
If we pass `forMainFrameOnly:NO` the provided test creatives fail to render and we see a black screen instead of playable content.


OlenaPostindustria (commented):
Could you please specify which creative was failing?

OlenaPostindustria (approved):
LGTM

PR #1131, opened 2025-04-24
[SPM Migration] PBMError, BidRequester

jclou:
- `PBMError` to Swift implementation
- `PBMBidRequesterProtocol` to Swift
- `PBMBidRequester` to Swift protocol


PrebidMobile/Factory.swift
```diff
@@ -0,0 +1,34 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+    
+
+import Foundation
+
+@objc(PBMFactory) @_spi(PBMInternal) public
+class Factory: NSObject {
```
YuriyVelichkoPI, line 20: Nitpic: Can you use a more descriptive name for this class, like `BidRequesterFactory` or another name if this class is intended to create not only `BidRequester` objects? 

Can we also use more canonical function names for the factory objects with a `create` prefix?

To get the following code in the result: 
 `BidRequesterFactory.createBidRequester(...)`

YuriyVelichkoPI (approved):
LGTM 

with one nitpick 

also, @mdanylov-sigma shoud review

mdanylov-sigma (dismissed):
LGTM!
Agreed with Yuriy regarding naming.  
@YuriyVelichkoPI  Should we create a separate branch for SPM integration to merge all subsequent PRs there instead of master ?

YuriyVelichkoPI:
> LGTM! Agreed with Yuriy regarding naming. @YuriyVelichkoPI Should we create a separate branch for SPM integration to merge all subsequent PRs there instead of master ?

Discussing the topic in the Prebid Slack we discussed a migration approach in detail and agreed to merge into `main` gradually instead of managing the huge branch with all needed changes, at least there were no objections: 

https://prebid.slack.com/archives/C7TLBPKUM/p1745009188966939?thread_ts=1744301713.647569&cid=C7TLBPKUM

If tests are passing and examples in Internal Test App work well we should be safe to move in this way. The full migration can take several patch ot even minor releases.

PrebidMobile/Factory.swift
jclou: This factory will get used for a bunch of other types following this PR, so I think I'll keep this one as-is for now.

I chose these names for factory methods to try to keep it short and closer to the normal initializer names, hoping it would minimize the number of formatting changes we'll need to make to keep lines from wrapping onto the next line. I actually considered typealiasing `Factory` to `F` to keep the Swift changes to just two characters. 

But there's also value in keeping things easy to understand for others not involved with this SPM migration, so I'll go ahead with adding `create` to the method names.

YuriyVelichkoPI:
@mdanylov-sigma it looks like we need to upgrade the supported GAM version 
tests are not passing

mdanylov-sigma:
> @mdanylov-sigma it looks like we need to upgrade the supported GAM version tests are not passing

@YuriyVelichkoPI Just updated, could you approve PR #1132 please?

YuriyVelichkoPI:
@jclou , please rebase your branch - it should sove the errors in tests

YuriyVelichkoPI (approved):
LGTM

mdanylov-sigma (approved):
LGTM

PR #1148, opened 2025-05-26
Adding skip status check flag for init

mdanylov-sigma:
Align implementation with Android, read more [#857](https://github.com/prebid/prebid-mobile-android/issues/857)

### DIFF
--- a/PrebidMobile/ConfigurationAndTargeting/Prebid.swift
+++ b/PrebidMobile/ConfigurationAndTargeting/Prebid.swift
@@ -132,6 +132,12 @@ public class Prebid: NSObject {
     /// If true, the sdk will add `includebidderkeys` flag inside the targeting object described in [PBS Documentation](https://docs.prebid.org/prebid-server/endpoints/openrtb2/pbs-endpoint-auction.html#targeting)
     public var includeBidderKeys = false
     
+    /**
+     * If true, the SDK will not check the PBS status during initialization. This will save initialization time
+     * if the PBS endpoint is always live and handled client side
+     */
+    public var shouldDisableStatusCheck: Bool = false
+    
     // MARK: - Public Methods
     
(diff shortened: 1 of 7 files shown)


PrebidMobile/ConfigurationAndTargeting/PrebidSDKInitializer.swift
```diff
@@ -29,9 +29,9 @@ class PrebidSDKInitializer {
         
         PrebidJSLibraryManager.shared.downloadLibraries()
         
-        serverStatusRequester.requestStatus { completion?($0, $1) }
-        
         Prebid.registerPluginRenderer(PrebidRenderer())
+
+        serverStatusRequester.requestStatus { completion?($0, $1) }
```
YuriyVelichkoPI, line 34: Is it necessary to change the order of initialization steps?

PrebidMobile/ConfigurationAndTargeting/Prebid.swift
```diff
@@ -132,6 +132,12 @@ public class Prebid: NSObject {
     /// If true, the sdk will add `includebidderkeys` flag inside the targeting object described in [PBS Documentation](https://docs.prebid.org/prebid-server/endpoints/openrtb2/pbs-endpoint-auction.html#targeting)
     public var includeBidderKeys = false
     
+    /**
+     * If true, the SDK will not check the PBS status during initialization. This will save initialization time
```
YuriyVelichkoPI, line 136: Does it actually save time? Server status check is an async operation that just drops a console log message if the status endpoint doesn't respond with an OK status.

So if this change saves initialization time, the comment is ok. If it doesn't, it's better to remove the second sentence.

OlenaPostindustria (approved):
LGTM

PR #1150, opened 2025-05-28
[SPM Migration] Interstitial loader

mdanylov-sigma:


PrebidMobile/PrebidMobileRendering/Prebid/Integrations/GAM/AdLoading/PBMInterstitialAdLoader.m
```diff
@@ -27,14 +25,14 @@
 #import <PrebidMobile/PrebidMobile-Swift.h>
 #endif
 
-@interface PBMInterstitialAdLoader () <InterstitialControllerLoadingDelegate, InterstitialEventLoadingDelegate>
+@interface PBMInterstitialAdLoader_Objc: NSObject <PBMInterstitialAdLoader>
```
YuriyVelichkoPI, line 28: Can you migrate this class as well? Or will it be migrated in another scope?

PR #1157, opened 2025-06-03
[SPM Migration] BannerAdLoader.

mdanylov-sigma:


YuriyVelichkoPI:
@mdanylov-sigma, do you want to fix it in the scope of a separate issue, since it is not a regression? If yes, please create a respective follow-up issue, and we'll merge the current changes.

PR #1158, opened 2025-06-03
[SPM Migration] AutoRefreshManager

mdanylov-sigma:


YuriyVelichkoPI (commented):
I'd put this migration on hold until a precise manual check of all possible cases. 

The refresh looks simple, but if it works incorrectly, it leads to extra bid requests or vice versa, omitted bid requests. 

The most fragile part here is the transition of the app between the background and foreground, where refresh should be paused and resumed, respectively. 

We need to check not only the code and UI/UX but also the traffic in many cases as well. I'm also adding @OlenaPostindustria  as a reviewer, but not sure if she will have time in the next days. 

Is this a blocker for any further steps?

YuriyVelichkoPI:
Hi @mdanylov-sigma! I'm going to review this migration. Please fix the conflicts.

YuriyVelichkoPI (approved):
LGTM with one nitpic. The PR can be merged once the log message is fixed. 

In addition, while reviewing the behavior on the current branch, I noticed several inconsistencies in how publishers are expected to use auto-refresh. To address this, I’ve created an [issue](https://github.com/prebid/prebid-mobile-ios/issues/1163) (epic) to revise the documentation, API, and overall implementation of auto-refresh in Prebid Mobile.

YuriyVelichkoPI (approved):
LGTM

PR #1164, opened 2025-06-16
[SPM Migration] PBMORTBBidRequest

mdanylov-sigma:


PrebidMobile/PrebidMobileRendering/Prebid/PBMCore/ORTB/PBMORTBBid.swift
```diff
@@ -0,0 +1,245 @@
...
+/// A SeatBid object contains one or more `Bid` objects, each of which relates to a specific impression in the bid
+/// request via the `impid` attribute and constitutes an offer to buy that impression for a given `price`.
+class PBMORTBBid<ExtType: PBMJsonCodable>: PBMJsonCodable, PBMORTBExtensible {
+    /// [Required]
+    /// Bidder generated bid ID to assist with logging/tracking.
+    var bidID: String
+    
+    /// [Required]
+    /// ID of the Imp object in the related bid request.
+    var impid: String
```
YuriyVelichkoPI, line 30: The general question - what will happen if live OpenRTB don't have this or another field which is non optional? 

I'd not really on OpenRTB notation in implementation and expect that any field can be absent in the response or request. It will make solution more general and robust.

PrebidMobileTests/RenderingTests/Tests/Prebid/PBMORTBTest.swift
```diff
@@ -23,11 +23,10 @@ class PBMORTBTest: XCTestCase {
     // MARK: - ORTB Response
     
     func testBidToJsonString() {
-        let bid = PBMORTBBid<NSDictionary>()
```
YuriyVelichkoPI, line 26: Unfortunately, we can't modify the support of OpenRTB protocol in this way. Because it is just a protocol and there actually no guarantees that some required field will be missed  in the real responses. 

Yes the probability is very low, but it still exists. Please make all properties of OpenRTB optional. And resubmit the PR. 

You can also remove PBM prefix for all new Swift classes, if it not break another PR related to parsing.

YuriyVelichkoPI (changes requested):

YuriyVelichkoPI (approved):
LGTM

PR #1171, opened 2025-06-25
[SPM Migration] OpenMeasurementSession

mdanylov-sigma:


YuriyVelichkoPI (approved):
LGTM

I didn't QA all integration cases but OM events for InApp banner appear as expected 

image

PR #1172, opened 2025-06-26
[SPM Migration] OMSDKVersionProvider, ViewExposure and UIView Extensions

mdanylov-sigma:


PrebidMobile/Utils/UIView+Extensions.swift
```diff
@@ -46,4 +46,100 @@ extension UIView {
         getSubview(view: self)
         return all
     }
+    
+    @objc(PBMAddFillSuperviewConstraints)
```
jclou, line 50: I don't know for sure, but I would guess that these util functions were intended for internal use and could be made `@_spi(PBMInternal)`

jclou (approved):

OlenaPostindustria (approved):
LGTM!

PR #1179, opened 2025-07-07
SPM package

jclou:
- Created a SPM package
  - Swift and Objc source files split into different directory structures.
  - SPM does not support recursive header search, so I moved all headers into the `PrivateObjcHeaders` directory. Not ideal, but I felt this was the most maintainable solution.
- Replaced OMSDK static xcframework with the dynamic xcframework.
  - SPM doesn't like the way OMSDK builds their static xcframeworks (a lib without a corresponding modulemap). It's possible that they could update their build process, but they haven't fixed this in years so I wouldn't count on it.
  - I updated the non-SPM targets to also use the dynamic framework so that we wouldn't need to check in and manage two different versions.
- Added SPM app target to project. Currently this is just an empty app that pulls in the SPM package locally to verify that it builds correctly.
  - Note that the XCode packages UI doesn't like local package paths ending in a relative path component and will automatically rewrite the path. In this case it rewrites `..` to `../../prebid-mobile-ios`. This path works fine locally, but not so well with our ci/cd setup. I had to manually edit the path in the .pbxproj to get it to work.
  - Added a script to the travis config to build this project.
- Fixed broken macro invocation in objc demo target (caused by the macros header no longer being publicly accessible)

XCode doesn't handle big project-level changes very well, so you might need to clean or delete derived data to get builds working after checking out this change.


Package.swift
```diff
@@ -1,30 +1,42 @@
...
     name: "PrebidMobile",
-
     platforms: [
-        .iOS(.v10)
+        .iOS(.v12),
     ],
-    
     products: [
-        
         .library(
```
mdanylov-sigma, line 13: Could you also add products for EventHandlers (PrebidMobileAdMobAdapters, PrebidMobileGAMEventHandlers and PrebidMobileMAXAdapters) ?

PrebidMobile/include/README.md
```diff
@@ -0,0 +1,5 @@
+#  Objc Public Headers
```
mdanylov-sigma, line 1: Is this README relevant to Obj-C Headers ?

Package.swift
```diff
@@ -1,30 +1,42 @@
...
             name: "PrebidMobile",
-            targets: ["PrebidMobile"])
+            targets: ["PrebidMobile", "PrebidMobileObjc"]
```
mdanylov-sigma, line 15: Could you rename targets to avoid imports described on the screenshot ?
Screenshot 2025-07-08 at 19 02 47

jclou: How about `__PrebidMobileInternal`? I reckon a good amount of iOS devs would recognize the double underscore to mean that this should be considered internal, since Apple uses this pattern often in their APIs. But otherwise the "Internal" suffix should tip people off. If someone does happen to see this name somewhere, the association with Prebid is clear.

On my machine it doesn't show up in autocomplete even if I try to import it.

Screenshot 2025-07-08 at 11 53 44 AM

Screenshot 2025-07-08 at 11 50 46 AM

Package.swift
jclou: I added them. It turns out that xcode will auto-generate schemes for SPM products and since they share the same name with the existing schemes it breaks the build script. Locally the build script works fine, but when it runs on circleci, xcodebuild will more often than not select the SPM scheme and fail.

To fix this I added the `Lib-` prefix to our existing schemes... lmk if that is a problem. An alternate workaround could be to remove the SPM project from the workspace, either completely or as part of the build script when running in the ci/cd environment.

mdanylov-sigma: I'm a bit overwhelmed by the number of schemes in the project at the moment, but we still need to validate the SPM integration through CI/CD, so unfortunately we can’t rely on the proposed alternate workaround.

Also, I’m wondering whether it would make sense to move the event handler packages into separate repositories. Based on the current setup, we’re pulling in additional dependencies (GAM, MAX, etc.) even for the PrebidMobile core, which might be confusing for app developers. I'd like to confirm this approach with @YuriyVelichkoPI  as well.

Screenshot 2025-07-09 at 20 34 35

Screenshot 2025-07-09 at 20 27 25

jclou: I believe those dependencies will only show up if you take those event handler/adapter dependencies and projects that only take `PrebidMobile` won't see any extra dependencies.

Also we would still be able to test SPM through ci/cd. If we remove it from the .xcworkspace completely then we would just need to change the script to use `SPMTestApp.xcodeproj` directly. 

The other proposed solution would involve adding a param to the invocation of `buildPrebidMobile.sh` in the circleci config. If that param is passed then we delete or rename `SPMTestApp.xcodeproj` so that the problematic generated schemes don't interfere with the build. Local invocations of `buildPrebidMobile.sh` would skip this step, and also `buildPrebidSPM.sh` can stay the same.

YuriyVelichkoPI: I tried to create a new project and added the dependency from the branch for this PR:

Screenshot 2025-07-10 at 11 14 45

Then I selected only Prebid Mobile for adding to the target:

Screenshot 2025-07-10 at 11 15 54

And got all other dependencies in the project (even if they are not added in the compile time, they still add mess to the project)
Screenshot 2025-07-10 at 11 16 07

What looks even messier to me is the content of the PrebidMobile dependency. 

Screenshot 2025-07-10 at 11 16 20

It displays the full content of the repository, which doesn't look clean. 

So, yes - creating a separate "data sources" for all Prebid Mobile dependencies looks reasonable. Most probably, it will be a separate step out of the scope of this PR. 

As an alternative, we can consider creating separate branches that contain only the necessary codebase for the specific dependency.

YuriyVelichkoPI: Discussed this shortly on the Backlog Review call and confirmed that we can do the cleanup work later, and it is not a blocker for this PR.

jclou: I see, [SE-0226](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0226-package-manager-target-based-dep-resolution.md) is only partially implemented so far. Dependencies that aren't used won't be compiled into the a build, but they will unfortunately still be downloaded and displayed in xcode's side bar.

Using branches to workaround this issue would not scale well as we would need a separate branch for each permutation, and again for each release version. We could move the adapters/event handlers into their own packages, but it would add extra complexity to the dev/testing/release processes.

IMO this is just how SPM currently works and most devs aren't really concerned with what shows up in xc's packages sidebar. I think the same can be said for how the entire contents of the repo shows up in the package's content.

YuriyVelichkoPI:
@mdanylov-sigma could you check the following on your side. 

I added a simple ad loading code to SPMTests app. 

```
class ViewController: UIViewController, BannerViewDelegate {
    
    var banner: BannerView?
    
    func bannerViewPresentationController() -> UIViewController? {
        self
    }
    
    func bannerView(_ bannerView: BannerView, didReceiveAdWithAdSize adSize: CGSize) {
        
    }
    
    func bannerView(_ bannerView: BannerView, didFailToReceiveAdWith error: any Error) {
        print(error.localizedDescription)
    }
    

    override func viewDidLoad() {
        super.viewDidLoad()
        // Do any additional setup after loading the view.
        
        GAMUtils.shared.initializeGAM()
        
        //Set up SDK.
        try? Prebid.initializeSDK(serverURL: "https://prebid-server-test-j.prebid.org/openrtb2/auction") { status, error in
            switch status {
            case .succeeded:
                print("Prebid successfully initialized")
                Prebid.shared.prebidServerAccountId = "123asdasd"
            case .failed:
                if let error = error {
                    print("An error occurred during Prebid initialization: \(error.localizedDescription)")
                }
            case .serverStatusWarning:
                if let error = error {
                    print("Prebid Server status checking failed: \(error.localizedDescription)")
                }
            case .serverStatusSkipped:
                print("Prebid status check skipped")
            default:
                break
            }
        }

        
        print(Prebid.shared.version)
        print(AdMobConstants.PrebidAdMobRewardedAdapterVersion)
        print(GAMUtils.shared)
        print(MAXConstants.PrebidMAXAdapterVersion)
        
        let adSize = CGSize(width: 320, height: 50)
        banner = BannerView(frame: CGRect(origin: .zero, size: adSize),
                                configID: "123dsfsdf",
                                adSize: adSize)
        
        banner?.delegate = self
        
        banner?.loadAd()
    }

}`
```

And it doesn't work for me. The SDK randomly returns the following errors on attempting to load the ad: 

- `Prebid server does not recognize Account Id` 
- `Prebid server does not recognize Config Id`

The minor issue here is irrelevant error messages. The server doesn't do anything here. SDK tries to validate the config and fails. Within the scope of another task, we should revise the text of the messages. 

However, the main issue for me is that the SDK doesn't validate the IDs randomly and doesn't make the bid request at all. In debug mode, I placed a breakpoint in `PBMBidRequester_Objc` and observed that the following validation sometimes passes and sometimes fails. Without any changes in the app code, of course:

```
  if ([self isInvalidID:self.sdkConfiguration.prebidServerAccountId]) {
      return [PBMError prebidInvalidAccountId];
  }
 ```
 
 Please try to reproduce or advise me on what I'm doing wrong.

YuriyVelichkoPI (approved):
LGTM

mdanylov-sigma (approved):
LGTM!

YuriyVelichkoPI:
I also created a follow-up task to prepare the SDK for the release after this PS is merged: 
https://github.com/prebid/prebid-mobile-ios/issues/1183

mdanylov-sigma:
> I also created a follow-up task to prepare the SDK for the release after this PS is merged: #1183

@YuriyVelichkoPI thanks! Take a look on that

YuriyVelichkoPI:
And one more follow-up for the future enhancements of distribution via SPM:
https://github.com/prebid/prebid-mobile-ios/issues/1184

PR #1191, opened 2025-07-30
Unit test modifications

mdanylov-sigma:


mdanylov-sigma:
Updated the info about the new pipelines and triggers.

> Additionally, I have noticed that the PR check time remains quite time-consuming, taking approximately 25 minutes. It looks like it runs all unit tests. Do you want to keep it or do you have plans to deduce the overall time eventually?

I assumed the long duration was due to queueing and that the reported 25 minutes didn’t reflect actual runtime. 
Thanks for pointing it out - I’ll investigate why the suite still takes so long even after we cut the unit tests by about half.

mdanylov-sigma:
> Does it make sense to add building the SPM Demo app?

The SPM Demo app pulls from the `master`, so it doesn’t reflect the state of the open PR. That's why I skipped it and built only the frameworks. Does it make sense ?

YuriyVelichkoPI (approved):
LGTM

PR #1192, opened 2025-07-31
Remove UIAlertController setting overrides

OlenaTeqBlaze:
## Motivation

This pull request resolves an orientation issue introduced after integrating PrebidMobile into the project. Specifically, we observed that the app would unexpectedly switch from landscape to portrait mode when presenting an alert. 
The root cause appears to be outdated logic related to overriding `UIAlertController` settings, which is no longer necessary and has therefore been removed as part of this fix.


mdanylov-sigma (approved):
LGTM! The PR is also related to #1055

mdanylov-sigma:
@OlenaTeqBlaze could you please the PR rebase on master ? (I've just bumped GAM version to pass the failing check)

PR #1198, opened 2025-08-14
Configurable Location Decimal Precision for User and Device Geo

ankit-thanekar007:
Adds Implementation for https://github.com/prebid/prebid-mobile-ios/issues/1205

### DIFF
--- a/PrebidMobile/Objc/PrebidMobileRendering/Networking/Parameters/PBMGeoLocationParameterBuilder.m
+++ b/PrebidMobile/Objc/PrebidMobileRendering/Networking/Parameters/PBMGeoLocationParameterBuilder.m
@@ -58,11 +58,8 @@
     }
     
     if (self.locationManager.coordinatesAreValid) {
-        CLLocationCoordinate2D coordinates = self.locationManager.coordinates;
-        
-        // Rounds with the precision defined in Targeting or returns the original coordinates if precision is 0.
-        coordinates = [[Utils shared] roundWithCoordinates:coordinates precision:[[Targeting shared] locationPrecision]];
-        
+        // Rounds with the precision defined in Targeting, or returns the original coordinates if precision is nil.
+        CLLocationCoordinate2D coordinates = [[Utils shared] roundWithCoordinates:self.locationManager.coordinates precision:[[Targeting shared] locationPrecision]];
         bidRequest.device.geo.type = @(PrebidConstants.LOCATION_SOURCE_GPS);
         bidRequest.device.geo.lat = @(coordinates.latitude);
         bidRequest.device.geo.lon = @(coordinates.longitude);
--- a/PrebidMobile/Objc/PrebidMobileRendering/Networking/Parameters/PBMGeoLocationParameterBuilder.m
+++ b/PrebidMobile/Objc/PrebidMobileRendering/Networking/Parameters/PBMGeoLocationParameterBuilder.m
@@ -59,6 +59,10 @@
     
     if (self.locationManager.coordinatesAreValid) {
         CLLocationCoordinate2D coordinates = self.locationManager.coordinates;
+        
+        // Rounds with the precision defined in Targeting or returns the original coordinates if precision is 0.
+        coordinates = [[Utils shared] roundWithCoordinates:coordinates precision:[[Targeting shared] locationPrecision]];
+        
         bidRequest.device.geo.type = @(PrebidConstants.LOCATION_SOURCE_GPS);
         bidRequest.device.geo.lat = @(coordinates.latitude);
         bidRequest.device.geo.lon = @(coordinates.longitude);
--- a/PrebidMobile/Objc/PrebidMobileRendering/Networking/Parameters/PBMParameterBuilderService.m
+++ b/PrebidMobile/Objc/PrebidMobileRendering/Networking/Parameters/PBMParameterBuilderService.m
@@ -139,8 +139,10 @@
     }
         
     NSValue * const coordObj = targeting.coordinate;
-    if (coordObj) {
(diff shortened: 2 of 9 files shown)


PrebidMobile/Swift/ConfigurationAndTargeting/Targeting.swift
```diff
@@ -189,6 +189,10 @@ public class Targeting: NSObject {
     /// See CoreLocation framework documentation.
     public var coordinate: NSValue?
     
+    /// Number of decimal places to use when rounding latitude/longitude for device geolocation.
+    /// A value of 0 means no fractional precision (whole degrees only).
+    /// Example: precision 2 ≈ ~1 km accuracy, precision 4 ≈ ~11 m accuracy.
+    public var locationPrecision: UInt = 0
```
YuriyVelichkoPI, line 195: Using 0 as a Precision value can be confusing. It can mean - no decimal digits. We should use optional here or some other value that does't make sense like a Precision vealue. I prefere optional. But need to be sure that property will be avaliable in the ObjC code.

PrebidMobile/Objc/PrebidMobileRendering/Networking/Parameters/PBMGeoLocationParameterBuilder.m
```diff
@@ -59,6 +59,10 @@ - (void)buildBidRequest:(PBMORTBBidRequest *)bidRequest {
     
     if (self.locationManager.coordinatesAreValid) {
         CLLocationCoordinate2D coordinates = self.locationManager.coordinates;
+        
+        // Rounds with the precision defined in Targeting or returns the original coordinates if precision is 0.
+        coordinates = [[Utils shared] roundWithCoordinates:coordinates precision:[[Targeting shared] locationPrecision]];
```
YuriyVelichkoPI, line 64: Please use `roundWithCoordinates` right in the line of initialization of `coordinates` (line 61).

PrebidMobile/Objc/PrebidMobileRendering/Networking/Parameters/PBMParameterBuilderService.m
```diff
@@ -139,8 +139,10 @@ + (nonnull PBMORTBBidRequest *)createORTBBidRequestWithTargeting:(nonnull Target
     }
         
     NSValue * const coordObj = targeting.coordinate;
-    if (coordObj) {
-        const CLLocationCoordinate2D coord2d = coordObj.MKCoordinateValue;
+    if (coordObj && CLLocationCoordinate2DIsValid(coordObj.MKCoordinateValue)) {
```
YuriyVelichkoPI, line 142: Adding `CLLocationCoordinate2DIsValid(coordObj.MKCoordinateValue)` into the if condition can change the current behavior and lead to regression. Can we move 

```const CLLocationCoordinate2D coord2d = coordObj.MKCoordinateValue;```

back to the if's scope?

File: PrebidMobile/Swift/AdUnits/Utils.swift [hunk 4]
```diff
@@ -417,6 +418,18 @@ public class Utils: NSObject {
 
           return nil
       }
+    
+    @objc public func round(coordinates : CLLocationCoordinate2D, precision: UInt) -> CLLocationCoordinate2D {
+        guard precision > 0 else {
```
YuriyVelichkoPI, line 423: precision == 0 should be valid parameter for rouding (see the comment above)

PrebidMobile/Swift/AdUnits/Utils.swift
```diff
@@ -417,6 +418,18 @@ public class Utils: NSObject {
 
           return nil
       }
+    
+    @objc public func round(coordinates : CLLocationCoordinate2D, precision: UInt) -> CLLocationCoordinate2D {
```
YuriyVelichkoPI, line 422: Please add more Unit tests for this method. Use as many variants as you can imagine, including corner cases and nonsense values, like precision = 1000000000 or precision = -1 (If you'll have to use another data for ObjC compatibility)

PrebidMobile/Swift/AdUnits/Utils.swift
```diff
@@ -417,6 +418,27 @@ public class Utils: NSObject {
...
           return nil
       }
+    
+    @objc public func round(coordinates : CLLocationCoordinate2D, precision: NSNumber?) -> CLLocationCoordinate2D {
+        guard let precision, CLLocationCoordinate2DIsValid(coordinates) else {
+            return coordinates
+        }
+        
+        let multiplier = pow(10.0, precision.doubleValue)
+        guard !multiplier.isZero && multiplier.isFinite else {
```
YuriyVelichkoPI, line 428: This check doesn't support the negative value in precision properly. 

In the test above you use Int.min wich probably leads to `multiplier.isFinite == true`, but if you use just, for instance -2 the method will work and return the unexpected result. See the attached screenshot. 

Screenshot 2025-08-26 at 10 49 10


So this method should check that `precision >= 0` and if it is not - log an error and return the provided coordinates. 

And by the way. Requesting the Unit Tests, I meant just checking the result of the execution of the method, but not the full-featured tests with generating the request. Please add classical unt tests for the round() method like: 

```
      let coordinate1 = CLLocationCoordinate2D(latitude: 32.6533, longitude: -161.4343)
       let coordinate2 = CLLocationCoordinate2D(latitude: 32.65, longitude: -161.43)

       XCTAssertEqual(coordinate2, round(coordinates: coordinate1, precision: 2))
 ```
 
 With different possible values.

YuriyVelichkoPI (approved):
LGTM

PR #1201, opened 2025-08-15
Update playbackend default value from 2 → 1 for interstitial video ads

### LINKED ISSUE #1199: Suggestion: Should VIDEO_INTERSTITIAL_PLAYBACK_END value be changed from 2 → 1 for interstitial videos?
## Summary
Discussion / suggestion: Change the value of `VIDEO_INTERSTITIAL_PLAYBACK_END` from 2 → 1 for video ads.

## Current Behavior
- Currently, `VIDEO_INTERSTITIAL_PLAYBACK_END` is sent as `2`.  
- All videos are interstitials, so "On Leaving Viewport" **never occurs**.  
- Video end always happens as "On Video Completion or when Terminated by User". 

## Problem
- Sending `VIDEO_INTERSTITIAL_PLAYBACK_END= 2` may cause DSPs / advertisers to **undervalue the ad request**, leading to lower bids than the actual inventory performance would justify.    

## Suggestion / Question
- Should `VIDEO_INTERSTITIAL_PLAYBACK_END` be set to `1` instead, to reflect the actual behavior?  
- I have already prepared a PR for this change; if the solution looks reasonable, I would appreciate it if you could take a look.

## Additional Notes
- This issue is for the Android implementation; a similar issue has been created for iOS.
- Feedback or review from maintainers on the PR would be greatly appreciated.

ZenyB:
## Summary
Change the default value of the `playbackend` from 2 → 1 for interstitial video ads.

## Current Behavior
- The `playBackend` property is set to `2` by default.
- All videos are interstitials, so "On Leaving Viewport" never occurs.
- Video end always happens as "On Video Completion or when Terminated by User".
- Sending `playBackend = 2` may cause DSPs / advertisers to undervalue the ad request, leading to lower bids than the actual inventory performance would justify.

## Change
- Update default value of `playBackend` to `1` to reflect actual behavior.

## Impact
- This change is backward-compatible and only adjusts the default setting to better reflect real playback behavior, without impacting other features.

## Related Issue
- Closes #1199

### DIFF
--- a/PrebidMobile/Objc/PrebidMobileRendering/Prebid/PBMCore/PBMPrebidParameterBuilder.m
+++ b/PrebidMobile/Objc/PrebidMobileRendering/Prebid/PBMCore/PBMPrebidParameterBuilder.m
@@ -197,7 +197,7 @@
                 PBMORTBVideo * const nextVideo = nextImp.video;
                 
                 if (!self.adConfiguration.adConfiguration.isOriginalAPI) {
-                    nextVideo.playbackend = @(2);
+                    nextVideo.playbackend = @(1);
                     nextVideo.pos = @(7);
                     nextVideo.protocols = @[@(2),@(5)];
                     nextVideo.mimes = PrebidConstants.SUPPORTED_VIDEO_MIME_TYPES;
--- a/PrebidMobileTests/RenderingTests/Tests/PBMORTBAbstractTest.swift
+++ b/PrebidMobileTests/RenderingTests/Tests/PBMORTBAbstractTest.swift
@@ -337,9 +337,9 @@ class PBMORTBAbstractTest : XCTestCase {
         pbmORTBVideo.protocols = [2, 5]
         pbmORTBVideo.pos = 7
         pbmORTBVideo.delivery = [3]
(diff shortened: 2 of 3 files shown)


PrebidMobile/Objc/PrebidMobileRendering/Prebid/PBMCore/PBMPrebidParameterBuilder.m
```diff
@@ -197,7 +197,7 @@ - (void)buildBidRequest:(nonnull PBMORTBBidRequest *)bidRequest {
                 PBMORTBVideo * const nextVideo = nextImp.video;
                 
                 if (!self.adConfiguration.adConfiguration.isOriginalAPI) {
-                    nextVideo.playbackend = @(2);
+                    nextVideo.playbackend = @(1);
```
YuriyVelichkoPI, line 200: The `playbackend` should be changed only for interstitial video. But this code changes it for Oustream video as well. 

Once  the `playbackend` is changed only for interstitial video - please update the unit tests respectively.

YuriyVelichkoPI (changes requested):

Issue #1197, opened 2025-08-13
Add BannerVideoListener Equivalent for Video Playback Events in Prebid Mobile iOS


savan-verse:
**Is your feature request related to a problem? Please describe.**
In the Prebid Mobile Android SDK, there is a BannerVideoListener interface that provides callbacks for important video playback events (e.g., video start, pause, resume, completion) when rendering banner video creatives.
In Prebid Mobile iOS, there is no equivalent API. When using the Rendering API, video playback is handled internally by PBMVideoView with AVPlayer, but these playback events are not exposed to the implementor application.
This makes it difficult to track user engagement with video ads, measure playback progress, or trigger application-specific actions during playback.

**Describe the solution you'd like**
Introduce a protocol (e.g., BannerVideoListener or VideoPlaybackDelegate) in Prebid Mobile iOS that works similarly to the Android BannerVideoListener.

Example methods:
```
func videoDidStart()
func videoDidPause()
func videoDidResume()
func videoDidComplete()
func videoDidMute()
func videoDidUnmute()
```

The SDK should call these delegate methods whenever the corresponding video playback events occur in PBMVideoView or similar internal classes.
This will align iOS with the Android SDK feature set and allow consistent cross-platform event handling.

**Describe alternatives you've considered**
- Traversing the view hierarchy to find AVPlayerLayer inside PBMVideoView and manually adding observers to AVPlayer. This approach is fragile and depends on internal implementation details that may change between SDK versions.
- Method swizzling on internal classes to capture AVPlayer creation and attach event observers. This is not recommended for production use because it relies on private APIs.

**Additional context**
Adding this feature will help publishers and SDK integrators measure and react to user interactions with video ads in banner placements, improve analytics, and maintain parity between iOS and Android implementations.
Android reference: [BannerVideoListener](https://github.com/prebid/prebid-mobile-android/blob/master/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/api/rendering/listeners/BannerVideoListener.java)

> P.S. If this functionality already exists in the iOS SDK, please point me to the relevant implementation or documentation.

PR #1202, opened 2025-08-19
BannerView delegate to manage video playback 

mdanylov-sigma:
#1197

### DIFF
--- a/InternalTestApp/PrebidMobileDemoRendering/ViewControllers/Adapters/Prebid/OpenX/PrebidBannerController.swift
+++ b/InternalTestApp/PrebidMobileDemoRendering/ViewControllers/Adapters/Prebid/OpenX/PrebidBannerController.swift
@@ -37,6 +37,12 @@ class PrebidBannerController: NSObject, AdaptedController, PrebidConfigurableBan
     private let adViewDidDismissScreenButton = EventReportContainer()
     private let adViewWillLeaveApplicationButton = EventReportContainer()
     
+    private let videoPlaybackDidPauseButton = EventReportContainer()
+    private let videoPlaybackDidResumeButton = EventReportContainer()
+    private let videoPlaybackWasMutedButton = EventReportContainer()
+    private let videoPlaybackWasUnmutedButton = EventReportContainer()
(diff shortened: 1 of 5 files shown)


InternalTestApp/PrebidMobileDemoRendering/ViewControllers/Adapters/Prebid/OpenX/PrebidBannerController.swift
```diff
@@ -159,6 +160,21 @@ class PrebidBannerController: NSObject, AdaptedController, PrebidConfigurableBan
...
+    func videoPlaybackDidPause(_ banner: PrebidMobile.BannerView) {}
+    
+    func videoPlaybackDidResume(_ banner: PrebidMobile.BannerView) {}
+    
+    func videoPlaybackWasMuted(_ banner: PrebidMobile.BannerView) {}
+    
+    func videoPlaybackWasUnmuted(_ banner: PrebidMobile.BannerView) {}
+    
+    func videoPlaybackDidComplete(_ banner: PrebidMobile.BannerView) {
+        print("BannerView video playback is finished.")
```
YuriyVelichkoPI, line 175: Please not only log the message but also add the respecive UI elements as for other integration cases: 
image

The reason is that in the ideal world, we would have to add autotests for this functionality. The UI tests should track and count the firing of respective events. That is why we should have a visual presentation of these delegate methods.

YuriyVelichkoPI (changes requested):

InternalTestApp/PrebidMobileDemoRendering/ViewControllers/Adapters/Prebid/OpenX/PrebidBannerController.swift
mdanylov-sigma: > Please not only log the message but also add the respecive UI elements as for other integration cases: 
> 
> image
> 
> 
> 
> The reason is that in the ideal world, we would have to add autotests for this functionality. The UI tests should track and count the firing of respective events. That is why we should have a visual presentation of these delegate methods. 
> 
> 

 Got it, will update

Issue #1194, opened 2025-08-06
Need for slot-specific targeting key-values


Umreg:
I want to have a method for setting custom key-values per ad unit.
So that the publisher can pass slot specific information into the bid stream. 

[It was removed ](https://github.com/prebid/prebid-mobile-ios/issues/977)without any replacement.
For key-values, the Arbitrary ORTB should be used:

Methods available:
- `setGlobalOrtbConfig`
- `impOrtbConfig`

`setGlobalOrtbConfig` ==> needs to be cleared and applied, forcing queuing
`impOrtbConfig` ==> Open RTB doesn't support "keywords" being passed here

Relates to:
https://github.com/prebid/prebid-mobile-android/issues/868

PR #1203, opened 2025-08-19
Content ORTB config

mdanylov-sigma:
#1194

### DIFF
--- a/.github/workflows/PR_checks.yml
+++ b/.github/workflows/PR_checks.yml
@@ -17,13 +17,13 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  XCODE_VERSION: "16.1.0"
+  XCODE_VERSION: "16.4.0"
 
 jobs:
   build:
     name: Build PrebidMobile Frameworks
     if: ${{ !github.event.pull_request.draft }}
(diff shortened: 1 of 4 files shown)


PrebidMobile/Swift/AdUnits/AdUnit.swift
```diff
@@ -319,6 +319,18 @@ public class AdUnit: NSObject, DispatcherDelegate {
         adUnitConfig.impORTBConfig
     }
     
+    /// Sets the content OpenRTB configuration string for the ad unit.
+    ///
+    /// - Parameter ortbConfig: The content OpenRTB configuration string to set. Can be `nil` to clear the configuration.
+    public func setContentORTBConfig(_ ortbConfig: String?) {
```
YuriyVelichkoPI, line 325: I'd suggest introducing the full-featured `setGlobalOrtbConfig` on the ad unit level. The `Targeting.setGlobalOrtbConfig` method should continue to work without changes. But if the publisher sets the unit level `setGlobalOrtbConfig`, it should take precedence over `Targeting.setGlobalOrtbConfig` .

We can also consider "merging" of AdUnit and the Targeting level global config. 

Thoughts?

Issue #1196, opened 2025-08-07
Bug: device.w/h should be in physical pixels in request


OlenaPostindustria:
## Describe the bug
The Prebid SDK is currently sending `device.w` and `device.h` values in device-independent pixels (DIPS). However, according to the OpenRTB specification, these fields should represent physical pixel dimensions.

## Reproduce

1. Run PrebidDemoSwift.
2. Run any test case.
3. See the device dimensions measured in DIPS.

<img width="799" height="825" alt="Image" src="https://github.com/user-attachments/assets/99ab8acc-9d64-4369-ad16-f28bd300cde2" />

PR #1204, opened 2025-08-19
Device screen size in pixels in a request

mdanylov-sigma:
#1196 

Part of [prebid/prebid-server#4501](https://github.com/prebid/prebid-server/issues/4501)

Sends physical screen pixels in `device.w`/`device.h` and opts in to server-side pxratio conversion by always including `ext.prebid.sdk.usepxratio: true` in bid requests. This allows Prebid Server to correctly convert physical pixels to DIPs when generating interstitial ad sizes for high-density screens.

## Changes

- `PBMDeviceInfoParameterBuilder`: switched from `screenSize` to `screenSizeInPixels` so `device.w`/`device.h` carry physical pixels (multiplied by `UIScreen.main.scale`); `device.pxratio` was already set to `UIScreen.main.scale`
- `DeviceAccessManager`: added `screenSizeInPixels()` returning screen dimensions in physical pixels
- `PBMORTBBidRequestExtPrebid`: always emits `ext.prebid.sdk.usepxratio: true`; `sdk` object is now included in every request (previously only when renderers were present); renderers are still appended when set

## Testing

- `MockDeviceAccessManager.screenSizeInPixels()` updated to match real implementation rounding (`Int(($0 * scale).rounded())`)
- `DeviceInfoParameterBuilderTest.testDeviceSize` updated to assert against `mockDeviceAccessManager.screenSizeInPixels()` dynamically rather than hardcoded values
- `PBMORTBAbstractTest` expected JSON strings updated to include `"usepxratio":true` in the `sdk` object

### DIFF
--- a/PrebidMobile/Objc/PrebidMobileRendering/Networking/Parameters/PBMDeviceInfoParameterBuilder.m
+++ b/PrebidMobile/Objc/PrebidMobileRendering/Networking/Parameters/PBMDeviceInfoParameterBuilder.m
@@ -74,7 +74,7 @@
         return;
     }
     
-    CGSize screenSize = self.deviceAccessManager.screenSize;
+    CGSize screenSize = self.deviceAccessManager.screenSizeInPixels;
     
     bidRequest.device.w = @(screenSize.width);
     bidRequest.device.h = @(screenSize.height);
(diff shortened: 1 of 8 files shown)


OlenaPostindustria (changes requested):

YuriyVelichkoPI:
The issue related to calculation of sizes for interstitial ads on PBS: https://github.com/prebid/prebid-server/issues/4501

Issue #835, opened 2023-05-01
Dr. Prebid does not build in 2.1.0


ourcraig:
**Describe the bug**
Dr. Prebid does not build in `2.1.0`. When trying to build and run in iOS Simulator, Dr. Prebid throws various errors concerning undeclared identifiers e.g.

* `Use of undeclared identifier 'GenderFemale'; did you mean 'PBMGenderFemale'?` in `PBVPrebidSDKValdiator -setPrebidTargetingParams`
* `Use of undeclared identifier 'kGADAdSizeFluid'; did you mean 'GADAdSizeFluid'?` in `PBVLineItemsSetupValidator - startTest`

When all undeclared identifiers are fixed based on Xcode suggestions, a new error appears during build:
* `Exception	NSException *	"The Google Mobile Ads SDK was initialized without an application ID. Google AdMob publishers, follow instructions at https://googlemobileadssdk.page.link/admob-ios-update-plist to set a valid application ID. Google Ad Manager publishers, follow instructions at https://googlemobileadssdk.page.link/ad-manager-ios-update-plist."	0x0000600002da4fc0`

**To Reproduce**
1. Download latest stable release from https://github.com/prebid/prebid-mobile-ios/releases/tag/2.1.0
2. Extract `prebid-mobile-ios-2.1.0.zip`
3. Run `pod install` in `prebid-mobile-ios-2.1.0` directory
4. Open `PrebidMobile.xcworkspace`
5. Set Scheme to `Dr.Prebid`
6. Set Destination to anything under `iOS Simulator`
7. `Run` and build will fail

**Expected behavior**
I expect the Dr. Prebid application to build and run successfully on iOS Simulator

**Screenshots**
![Screen Shot 2023-05-01 at 5 50 10 pm](https://user-images.githubusercontent.com/12490985/235425643-1d006274-d144-4102-a49c-cfb8e3ee9777.png)
![Screen Shot 2023-05-01 at 5 45 02 pm](https://user-images.githubusercontent.com/12490985/235425666-a316b959-8b75-4f3b-b9c7-61a60c5ae040.png)

**Desktop (please complete the following information):**
 - OS: macOS Monterey `12.6`

**Smartphone (please complete the following information):**
 - Device: iOS Simulator
 - Version 14.2 (986.5)

YuriyVelichkoPI:
Hi @ourcraig ! This app is extremely outdated and it hasn't been supported for years. 

[...]

PR #1210, opened 2025-08-27
Removes Dr.Prebid application

mdanylov-sigma:
Accordingly to comment in #835 removes outdated Dr.Prebid application


YuriyVelichkoPI:
The current app's codebase is completely outdated. 

If the community needs to restore it, we would rewrite it anyway. 

If somebody needs the dr.Prebid functionality on Mobile - please open a respective issue.

Issue #1111, opened 2025-02-26
AdPosition Enum un-matched definitions between Android and iOS


AbrahamArmasCordero:
**Describe the bug**
Android has more Definitions for the Same Enum than iOS or Android has too many definitions Undefined in iOS does not exist

| Android | iOS|
|--------|--------|
| UNDEFINED(-1)| HERE IS THE BUG |
| UNKNOWN(0)| case undefined = 0 |
| ABOVE_THE_FOLD(1) | case aboveTheFold = 1 |
| LOCKED(2) | case locked = 2 |
| BELOW_THE_FOLD(3) | case belowTheFold = 3 | 
| HEADER(4) | case header = 4 | 
| FOOTER(5) | case footer = 5 | 
| SIDEBAR(6) | case sidebar = 6 | 
| FULLSCREEN(7) | case fullScreen = 7| 

also id 0 in android is "Unknown" and in ios is "Undefined" 

**To Reproduce**
go to android file where this is defined  https://github.com/prebid/prebid-mobile-android/blob/master/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/models/AdPosition.java#L36

and to iOS file where this is defined https://github.com/prebid/prebid-mobile-ios/blob/master/PrebidMobile/PrebidMobileRendering/AdPosition.swift#L33

**Expected behavior**
Both should have the same amount of values. Undefined in iOS should be id = -1 and add the Unknown value or if the problem is in android having too many values we should remove the appropiate value since idk if undefined it's just an option to null on the SDK

Using prebid 2.4.0 in both

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #1216, opened 2025-09-23
Ad position values

mdanylov-sigma:
Aligns the ad [position values with Android implementation](https://github.com/prebid/prebid-mobile-android/pull/886)
#1111


YuriyVelichkoPI (approved):
LGTM

Issue #1217, opened 2025-10-10
Bug: From version 2.5.0, Prebid SDK doesn't replace ${AUCTION_PRICE} inside the nativeAdMarkup with the actual winning bid price


vietta:
**Describe the bug**
From version 2.5.0 (after this [commit](https://github.com/prebid/prebid-mobile-ios/commit/96c8a5aab3617e55c3986ca6f86cf3f109deb60a)), Prebid SDK doesn't replace ${AUCTION_PRICE} inside the nativeAdMarkup with the actual winning bid price.
As a result, native ads produce click URLs that still contain ${AUCTION_PRICE} instead of a numeric value (e.g. 0.814)

**To Reproduce**
1. Integrate Prebid Mobile iOS SDK (version after commit 96c8a5aab3617e55c3986ca6f86cf3f109deb60a)
2. Load a native ad (e.g. from AppNexus / Xandr adapter)
3. Inspect the returned bid.nativeAdMarkup
4. Observe that the link.url still contains ${AUCTION_PRICE} (not replaced)

AppNexus / Xandr link.url example:
```
https://sin3-ib.adnxs.com/click2?e=...&s=...&bcr=...&pp=${AUCTION_PRICE}&bn=92572&test=1&clickenc=https%3A%2F%2Fr1-usc1.zemanta.com%2F...
```

**Expected behavior**
The ${AUCTION_PRICE} macro inside nativeAdMarkup should be replaced by the actual bid price as it was before the referenced commit.
For example:
"pp=${AUCTION_PRICE}" → "pp=0.814"

**Additional context**
Root cause:
Before commit 96c8a5a, the SDK replaced ${AUCTION_PRICE} inside rawBid.adm before creating the nativeAdMarkup.
```
rawBid.adm = macrosHelper.replaceMacros(in: rawBid.adm)
nativeAdMarkup = NativeAdMarkup(jsonString: rawBid.adm)
```

After commit 96c8a5a, the macro replacement happens only for bid.adm (not rawBid.adm):
`let bid = Bid(bid: rawBid)`
But nativeAdMarkup is still created from rawBid.adm, which never gets the replaced price value.

Suggestion (proposed fix): Use the processed bid.adm instead of rawBid.adm when generating nativeAdMarkup
```
// Before
nativeAdMarkup = NativeAdMarkup(jsonString: rawBid.adm)

// After (proposed)
nativeAdMarkup = NativeAdMarkup(jsonString: bid.adm)
```

[...]

PR #1218, opened 2025-10-11
Replace ${AUCTION_PRICE} with the actual winning bid price inside the nativeAdMarkup

vietta:
#1217


YuriyVelichkoPI (changes requested):
The fix looks totally correct. 

Please add the appropriate unit test for the `NativeAd.create()` function. The test should verify that `ad.nativeAdMarkup` contains the correct price value.

Issue #1195, opened 2025-08-06
App Crashed: com.apple.NSURLSession-delegate


thomas-ibm-weather:
**Describe the bug**
App crash when we upgrade to 3.1.0 in Swift Package Manager from Cocopod 3.0.0.
The crash seems cause by data race 

**To Reproduce**

**Expected behavior**

**Screenshots**

<img width="1105" height="823" alt="Image" src="https://github.com/user-attachments/assets/2b46fe53-a032-4eee-a3dc-fc00710cda35" />

**Smartphone (please complete the following information):**
 - Device: iPhone 16 Pro Max, iPad (9th generation), iPhone 16 Pro, iPhone 11 Pro
 - OS: 18.5.0, 18.6.0, 26.0.0
 - Version: 3.1.0

**Additional context**

vpathania1:
Hello, this issue is still happening and we are on version 3.1.0.

PR #1221, opened 2025-11-10
Fix crash from duplicate network callbacks in PBMBidRequester (#1195)

postindustria-code:
Potentially addresses #1195 

## Root Cause:
Network layer occasionally invokes the callback multiple times (redirects/retries).  First call reads `self.completion`, sets it to `nil`, and invokes it.  Second call reads nil completion and crashes at offset 0x10 trying to invoke the previously deallocated completion block.

## The Fix:
Wrapped completion access in @synchronized(self) with nil check.  If completion is `nil`, log warning and return early. Prevents crash while ensuring callback only executes once.

### Crash Signature (Verified):
- `EXC_BAD_ACCESS KERN_INVALID_ADDRESS at 0x0000000000000010`
- Location: `PBMBidRequester.m:154 (makeRequestWithCompletion callback)`

###  Testing:
Added two tests simulating duplicate callbacks (sequential + concurrent).  All existing tests pass.

### DIFF
--- a/PrebidMobile/Objc/PrebidMobileRendering/Prebid/PBMCore/PBMBidRequester.m
+++ b/PrebidMobile/Objc/PrebidMobileRendering/Prebid/PBMCore/PBMBidRequester.m
@@ -97,23 +97,34 @@
                  callback:^(PrebidServerResponse * _Nonnull serverResponse) {
         @strongify(self);
         if (!self) { return; }
-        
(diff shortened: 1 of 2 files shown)


mdanylov-sigma (approved):
LGTM! @YuriyVelichkoPI should we create a release with this fix asap ?

Issue #1227, opened 2025-12-09
Exposure checks failing: Apple internal views detected as "obstructing"


MattMur:
I'm guessing this might be something relatively new with iOS 26's new Glass UI....I'm seeing an issue where my prebid ad view is reporting viewability exposure checks incorrectly. It looks like Apple internal views FloatingBarContainerView and FloatingBarHostingView are being overlayed on top of my view heirarchy when using a UINavigationController, then, during the viewability polling in the method [testForObstructing](https://github.com/prebid/prebid-mobile-ios/blob/0b0dc9962eac145f1e8ee7c692c735cd274eaacd/PrebidMobile/Objc/PrebidMobileRendering/ExtensionsAndWrappers/Exposure/PBMViewExposureChecker.m#L214-L220) it is detecting these views as being obstructions and setting exposure to zero, even when the ad is 100% in view. 

In the view hierarchy debugger, you can see the two Apple internal view's overlayed on top:
<img width="400" height="400" alt="Image" src="https://github.com/user-attachments/assets/3594649c-8fee-46c2-990c-e95f724af429" />

Has anyone else experienced this issue? Are there any workarounds? Ideally I would expect Prebid to be smart enough check that a view is transparent before considering it as an obstruction.

YuriyVelichkoPI:
The viewability tracking approach of Prebid SDK is based on the Open Measurement approach. It assumes that only parts of the ad which are on the screen and not covered by any other views considered as viewable. 

If there are any **known** views which covers the ad on purpose - they should be registred as friendly obstructions. Including the transparent overlaps. 

I should admit that prebid SDK doesn't provide API to register the friendly obstructions like Google IMA does: https://developers.google.com/interactive-media-ads/docs/sdks/android/client-side/omsdk#:~:text=This%20transparent%20overlay%20is%20used,controls%20must%20not%20be%20registered.&text=The%20following%20sample%20code%20demonstrates,The%20string%20can%20be%20nil. 

Prebid SDK registers only own obstructions (playback and interstitial controls). And encourage publishers to be sure that the ad view is not covered by any other view in the app layout. 

Until now we haven't received any complains about this policy. But of course, it can be revisited. 

So in the scope of this task we can support new iOS version by registering the new know iOS' views as friendly obstructions. 

But we still shouldn’t ignore transparent overlapping views when determining ad visibility. Their opacity can change at any time due to external logic, and the ad may stop being visible.

[...]

PR #1228, opened 2025-12-14
Internal view obstructions

mdanylov-sigma:
#1227 

On recent iOS versions with the new `glass` UI, system UI components like tab bar, navigation bar are rendered as overlay views that sit above the application content. During viewability polling in `PBMViewExposureChecker`, these system views were detected as obstructions.

The exposure checker now filters internal UI components out.

Alongside the MRAID examples in the demo app were fixed.


YuriyVelichkoPI (approved):
LGTM

PR #1231, opened 2025-12-19
Fix: Enforce size constraints for GAM banner rendering

MattMur:
This PR fixes an issue I'm seeing with GAM when using BannerView where if an incorrect width/height is set on the parent view, the ad will fail to render. By enforcing ad size on the GAM banner itself inside GAMBannerEventHandler, and setting BannerView to fill to its parent view, the ad will always center-in-view and render. 
For Prebid rendering, this change shouldn't affect anything, since PBMWebView already sets the width and height for ads coming from Prebid server.
Also fixes warnings about ambiguous layout constraints in BannerView's child view by simply adding X and Y constraints. iOS Autolayout can make accurate guesses most of the time but it's better just to have them set, and with 750 priority, so the developer can override them if needed.

Screenshot 2025-12-18 at 5 58 16 PM

### DIFF
--- a/EventHandlers/PrebidMobileGAMEventHandlers/Sources/GAMBannerEventHandler.swift
+++ b/EventHandlers/PrebidMobileGAMEventHandlers/Sources/GAMBannerEventHandler.swift
@@ -242,6 +242,10 @@ public class GAMBannerEventHandler :
         
         if let banner = banner,
            let adSize = dfpAdSize {
+            // Enforce size constraints on banner. Ad will not render without correct constraints.
+            banner.banner.widthAnchor.constraint(equalToConstant: adSize.width).isActive = true
+            banner.banner.heightAnchor.constraint(equalToConstant: adSize.height).isActive = true
+            
             loadingDelegate?.adServerDidWin(banner.banner, adSize: adSize)
         }
     }
--- a/EventHandlers/PrebidMobileGAMEventHandlers/Sources/GAMBannerEventHandler.swift
+++ b/EventHandlers/PrebidMobileGAMEventHandlers/Sources/GAMBannerEventHandler.swift
@@ -206,6 +206,7 @@ public class GAMBannerEventHandler :
             
             if let banner = banner,
                let adSize = dfpAdSize {
+                setSizeConstraints(banner.banner, forAdSize: adSize)
                 loadingDelegate?.adServerDidWin(banner.banner, adSize: adSize)
             }
         }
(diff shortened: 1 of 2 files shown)


YuriyVelichkoPI:
Hi @MattMur ! Thanks for reporting! 

Please, provide more detailed description which allows to reproduce the behaviour. Or even better, add the use case to the internal test app. 

The statement "...  if an incorrect width/height is set on the parent view, the ad will fail to render." looks confusing to me, because why we should support scenario when the size of the parent view was set incorrectly? It's better to force publishers to set the correct size and layout properties. 

But I understand that maybe I miss something due to the numerous approaches of creating the app layout and in any case is better to have the exact use case that this change fixes.

YuriyVelichkoPI:
Does the provided change support multisize (adaptive) banner?

MattMur:
Hi @YuriyVelichkoPI thanks for reviewing this. I've been away for a bit but getting back into this. Yes, the idea of the change is to support multisize banners. The developer can set the BannerView to dynamically fill the parent view, meanwhile the inner view will be fixed to whatever bid size won, and then centered in view.

EventHandlers/PrebidMobileGAMEventHandlers/Sources/GAMBannerEventHandler.swift
```diff
@@ -242,6 +242,10 @@ public class GAMBannerEventHandler :
         
         if let banner = banner,
            let adSize = dfpAdSize {
+            // Enforce size constraints on banner. Ad will not render without correct constraints.
```
YuriyVelichkoPI: Should the same changes be applied in the `primaryAdReceived` method?

YuriyVelichkoPI (approved):
LGTM

EventHandlers/PrebidMobileGAMEventHandlers/Sources/GAMBannerEventHandler.swift
YuriyVelichkoPI: Thanks! 

Aside note: currently we are working on prioritization of tasks for next major release. If you are a Prebid member - please join the Prebid Mobile PMC calls and leave your feedback about what we should focus on. 

If you are not Prebid member - feel frees to share your thoughts in GitHub or email me and I'll forward the feedback to the committee.

MattMur: Appreciate that! We're not members as of yet, but will keep that in mind for the future.

PR #1235, opened 2026-01-22
Add thread-safe access to custom headers

ankit-thanekar007:
Problem : 

This PR fixes a thread safety issue that caused production crashes in some of our Apps. The crash occurred when `PrebidServerConnection.createRequest()` accessed `customHeaders` from multiple dispatch queues concurrently.  

Currently the headers on our side can update dynamically. This can be triggered because of some privacy flags being updated by the user on desktop, or some other use cases. 

Implement thread-safe custom headers to prevent race conditions when accessing headers from multiple threads. Uses a serial dispatch queue to synchronize all custom header operations.

Changes:
- Add customHeaderQueue dispatch queue for synchronization
- Update addCustomHeader() to use sync for thread-safe writes
- Update clearCustomHeaders() to use sync for thread-safe clearing
- Add getCustomHeaders() method for safe external access
- Update PrebidServerConnection to use getCustomHeaders() getter

### DIFF
--- a/PrebidMobile/Swift/ConfigurationAndTargeting/Prebid.swift
+++ b/PrebidMobile/Swift/ConfigurationAndTargeting/Prebid.swift
@@ -45,15 +45,14 @@ public class Prebid: NSObject {
     /// ORTB: bidRequest.test
     public var pbsDebug = false
 
-    /// Dispatch queue for thread-safe access to custom headers
-    let customHeaderQueue : DispatchQueue = DispatchQueue(label: "com.prebid.customHeaderQ")
-
-    /// Private backing storage for custom HTTP headers
-    private var _customHeaders: [String: String] = [:]
-
     /// Custom HTTP headers to be sent with requests.
+    ///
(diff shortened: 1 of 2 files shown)


PrebidMobile/Swift/ConfigurationAndTargeting/Prebid.swift
```diff
@@ -176,14 +189,26 @@ public class Prebid: NSObject {
...
-        customHeaders.removeAll()
+        customHeaderQueue.sync {
+            self._customHeaders.removeAll()
+        }
     }
-    
+
+    /// Returns a copy of the current custom HTTP headers.
+    /// - Returns: A dictionary containing all custom headers.
+    public func getCustomHeaders() -> [String : String] {
```
mdanylov-sigma: @ankit-thanekar007 do we really need the second getter for customHeaders? I guess we can use only `customHeaders` property, looks like the new method is redundant, what do you think ?

PR #1253, opened 2026-03-20
Fix burl auction macro replacement

berkbtasdemir:
## Summary

- Fix `burl` (billing URL) not having OpenRTB auction macros (e.g., `${AUCTION_PRICE}`) resolved before being used for impression tracking
- Changed `burl` from a computed property (returning raw `bid.burl`) to a stored property initialized with `ORTBMacrosHelper.replaceMacros(in:)`, consistent with how `adm` and `nurl` are already handled
- Ensures billing URLs fire with the correct resolved bid price for accurate revenue attribution


mdanylov-sigma (approved):
LGTM

Issue #1255, opened 2026-04-20
[SKAdNetwork] itunesitem, sourceapp and fidelity.timestamp parsed as integers, violating IAB OpenRTB spec (expected strings)


berkbtasdemir:
# [SKAdNetwork] `itunesitem`, `sourceapp` and `fidelity.timestamp` parsed as integers, violating IAB OpenRTB spec (expected strings)

## Summary

The Prebid iOS SDK parses `skadn.itunesitem` and `skadn.sourceapp` from bid responses as **integers**, but the IAB OpenRTB defines both fields as **strings**.

<img width="391" height="119" alt="Image" src="https://github.com/user-attachments/assets/0601ebf3-95aa-4d46-a6b6-833c3707162c" />

<img width="391" alt="Image" src="https://github.com/user-attachments/assets/a4f98549-ba16-482f-b15e-17c29c0e6380" />

This causes App Store linked ads to silently fail and the click does not open the App Store when a demand partner returns those fields as strings.

## IAB Spec (Expected Behavior)

| Field | IAB-specified type |
|---|---|
| `skadn.itunesitem` | `string` |
| `skadn.sourceapp` | `string` |
| `fidelity.timestamp` | `string` |

Reference: [IAB OpenRTB SKAdNetwork — Bid Response](https://github.com/InteractiveAdvertisingBureau/openrtb/blob/main/extensions/community_extensions/skadnetwork.md#bid-response)

## Real-World Impact

We are receiving bids from **Pubmatic** where `itunesitem` and `sourceapp` are returned as strings which is correct per the IAB spec. The SDK fails to parse these fields, breaking all App Store linked ad clicks on iOS.

## Example Spec-Compliant Bid Response (triggers the bug)

```json
{
  "skadn": {
    "fidelities": [
        {
		"fidelity": 0,
		"nonce": "11dced75-b961-44b2-b3d6-f85699aaa331",
		"signature": "MDUCGANaHwKPDqXt4Yv+PICATDd5NpCt9zrD2AIZANw5cyex+fAN0DKRbGVZdIFQ6x7Rht4Pvg==",
		"timestamp": "1776714872203"
	}
    ]
    "version": "2.2",
    "network": "cDkw7geQsH.skadnetwork",
    "sourcedentifier": "20",
    "itunesitem": "123456789",
    "sourceapp": "880047117",
  }
}
```

## Suggested Fix

[...]

PR #1256, opened 2026-04-21
Fix SKAdNetwork click flow for bid servers returning itunesitem, sourceapp, and timestamp as strings

berkbtasdemir:
## Summary

Fixes silent SKAdNetwork attribution failure when bid servers return `itunesitem`, `sourceapp`,
and `timestamp` as JSON strings instead of integers.

Per the [IAB OpenRTB SKAdNetwork spec](https://github.com/InteractiveAdvertisingBureau/openrtb/blob/main/extensions/community_extensions/skadnetwork.md), these fields are defined as **strings** in the bid response. However, the SDK was parsing them with a strict `as? NSNumber` cast causing `SkadnParametersManager` to receive `nil` values and silently skip the entire `SKStoreProductViewController` flow. No error, no log, no App Store deeplink.

Apple's StoreKit API requires these values as `NSNumber`, so the SDK must bridge the type mismatch.

## Changes

- Added `numberOrString(_:)` helper to `JSONObject` that tries `String → Int64 → NSNumber` first (spec-compliant path), then falls back to a raw `NSNumber` for servers already sending the correct type.
- Applied the helper to `itunesitem` and `sourceapp` in `ORTBBidExtSkadn`.
- Applied the helper to `timestamp` in `ORTBSkadnFidelity`

## References
- Upstream issue: prebid/prebid-mobile-ios#1255
- IAB OpenRTB SKAdNetwork spec: https://github.com/InteractiveAdvertisingBureau/openrtb/blob/main/extensions/community_extensions/skadnetwork.md
- Apple StoreKit ad network install validation keys: https://developer.apple.com/documentation/storekit/ad-network-install-validation-keys

### DIFF
--- a/PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCore/ORTB/Prebid/ORTBBidExtSkadn.swift
+++ b/PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCore/ORTB/Prebid/ORTBBidExtSkadn.swift
@@ -69,8 +69,8 @@ open class ORTBBidExtSkadn: NSObject, PBMJsonCodable {
         version             = json[.version]
         network             = json[.network]
         campaign            = json[.campaign]
-        itunesitem          = json[.itunesitem]
-        sourceapp           = json[.sourceapp]
+        itunesitem          = json.numberOrString(.itunesitem)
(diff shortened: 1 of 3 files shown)


mdanylov-sigma (approved):
LGTM!

Issue #1072, opened 2024-12-11
Support pass parameters to GMA SDK in integration scenarios with rendering API. 


YuriyVelichkoPI:
## Motivation 

Here is a description of the problem from the Slack message: 

> In the original API, there was the option to attach publisherProvidedID and customTargeting to the GAMRequest. Is there an equivalent to this in the rendering API for iOS?

The integration scenario: https://docs.prebid.org/prebid-mobile/modules/rendering/ios-sdk-integration-gam.html

## Objectives 

Need to design and implement a solution that allows configuring the GMARequest object  in the given integration scenario. 

The issue is relevant for both platforms iOS and Android

Issue #745, opened 2022-12-12
Support GMA SDK 9.14.0


OlenaPostindustria:
## Motivation
New GMA SDK version was released.

## Objectives
- support the latest GMA SDK in the Original API;
- support the latest GMA SDK in the AdMob Adapters;
- support the latest GMA SDK in the GAMEventHandlers.

PR #1266, opened 2026-05-15
Custom targeting for GAM rendering API

mdanylov-sigma:
https://github.com/prebid/prebid-mobile-ios/issues/1072
[#745 ](https://github.com/prebid/prebid-mobile-android/issues/745)


EventHandlers/PrebidMobileGAMEventHandlers/Sources/GAMRewardedEventHandler.swift
```diff
@@ -107,21 +115,15 @@ public class GAMRewardedAdEventHandler :
...
-            
-            if let responseTargeting = bidResponse.targetingInfo {
-                targeting.merge(responseTargeting) { $1 }
-            }
-            
-            if !targeting.isEmpty {
-                request.customTargeting = targeting
-            }
-                        
+        if bidResponse != nil {
```
OlenaPostindustria: why do we check if `bidResponse` is not nil in this case? the flow should continue normally regardless

OlenaPostindustria (changes requested):

OlenaPostindustria:
There is a potential issue with the handling of `customTargeting` in `EventHandlers`. In the GMA SDK, this property is defined as `NSDictionary<NSString *, id>`, but within `EventHandlers` it is cast to `[String: String]`. Since the original dictionary may contain non-string values, this cast can lead to data loss.

Reference: https://github.com/prebid/prebid-mobile-ios/blob/4f96890f9f8b38db6abd92044a928a709d5458d2/EventHandlers/PrebidMobileGAMEventHandlers/Sources/Boxes/GAMRequestWrapper.swift#L51

.gitignore
```diff
@@ -120,3 +120,4 @@ Pods/
 CLAUDE.md
 .claude/
 
+EventHandlers/Package.resolved
```
OlenaPostindustria, line 123: `Package.resolved` is committed to the branch

OlenaPostindustria (dismissed):
LGTM with minor comment

OlenaPostindustria (approved):

PR #1267, opened 2026-05-19
Ad expiration based on `bid.exp`

### LINKED ISSUE #850: Expire ads respectively to the `seatbid.bid.exp`
## Background

The [OpenRTB protocol](https://www.iab.com/wp-content/uploads/2016/03/OpenRTB-API-Specification-Version-2-5-FINAL.pdf) defines the `bid.exp` property that: 

> Advisory as to the number of seconds the bidder is willing to wait between the auction and the actual impression.

Since the In-App Native API and Rendering API, in fact, cache the bid until it isn't displayed, the SDK should utilize the `bid.exp` field to expire the cache or the respective Ad Object - BannerView, Interstitial controller, etc. 

### In-App Native

It looks like the `CacheManager` for In-App Native ads utilizes the value of `bid.exp` to invalidate the cached bid. However, there are no unit tests not for `Bid` class not for the `CacheManager.shared.save` with the expireInterval. 

So for the scope of the Native Ads need to add unit tests for the `Bid.exp` property. And for  
the `CacheManager.shared.save` method. 

Also, we need to add tests to verify the invocation of `NativeAd.cacheExpired` delegate method because it is essential to notify the app properly and not break this behavior in the future. 

### Rendering API

* Parse the `bid.exp` and make it a part of the ad model (most probably `CreativeModel`).
* Once the bid is received and the model is created, if `bid.exp` is not empty - create an async call after the `bid.exp` time to notify the SDK and the application that the ad is expired and can't be displayed anymore. 
* Introduce the delegate method the `adDidExpire` in Rendering API - banner, interstitial, and rewarded. 
* Once the expiration callback is triggered and the ad is still alive (in the memory):
    * Mark the ad as expired
    * Call the `adDidExpire` delegate method
    * **UPD Requirements** 
        * ~If the ad is not displayed on the screen - destroy it.~
        * ~For interstitial ad units, the `isReady` property should return `false` for expired ad~
        * ~The `show` method for interstitial ad units should do nothing~ 
       * For the banner - follow the current ad unit config 
           * refresh the unit
           * ~act like the ad is failed to load~. 
           * do nothing if ad unit is not refreshable
* Don't add the  API property to signal `$.imp[n].exp` in the request. Let's keep it on #877.
* Add unit tests
* Add integration cases and tests into Internal Test App

mdanylov-sigma:
Closes #850 

Implemented `bid.exp` expiration handling for Native and Rendering API flows:

- Added Native coverage for `Bid.exp`, `CacheManager.save(..., expireInterval:)`, and `NativeAd.cacheExpired` delegate behavior.
- Added Rendering API expiration support through the ad model and ad lifecycle.
- Expired banner/interstitial/rewarded ads now notify delegates and prevent invalid display.
- Interstitial readiness/show behavior now reflects expiration state.
- Banner expiration follows refresh-or-expire behavior.
- Added focused unit coverage for the new expiration flows.
- Split reusable test helpers into separate helper files.


PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/BannerView.swift
```diff
@@ -398,6 +398,22 @@ public class BannerView:
...
+    private func reportAdExpired() {
+        DispatchQueue.main.async { [weak self] in
+            guard let self else { return }
+            
+            self.delegate?.bannerViewDidExpire?(self)
+            self.deployedView?.removeFromSuperview()
+            self.deployedView = nil
+            
+            if self.adUnitConfig.refreshInterval > 0, !self.isRefreshStopped {
+                self.adLoadFlowController?.refresh()
```
OlenaPostindustria, line 419: If auto refresh is in flight, the manual refresh can potentially create a conflict here. Worth stopping it beforehand

PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCacheRenderers/DisplayView.swift
```diff
@@ -178,6 +192,30 @@ class DisplayView: UIView, PrebidMobileDisplayViewProtocol, AdViewManagerDelegat
...
+            self?.expireAd()
+        }
+        expirationWorkItem = workItem
+        DispatchQueue.main.asyncAfter(deadline: .now() + exp.doubleValue, execute: workItem)
+    }
+    
+    private func expireAd() {
+        guard !isExpired else { return }
+        
+        isExpired = true
```
OlenaPostindustria, line 214: let's stick to the same ordering inside `expireAd` in `DisplayView` and `InterstitialController`

PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/BannerView.swift
```diff
@@ -398,6 +398,22 @@ public class BannerView:
         }
     }
     
+    private func reportAdExpired() {
+        DispatchQueue.main.async { [weak self] in
+            guard let self else { return }
+            
+            self.delegate?.bannerViewDidExpire?(self)
```
OlenaPostindustria: I think it's worth to change the order - first, clear `deployedView`, then call `bannerViewDidExpire`

OlenaPostindustria (changes requested):

OlenaPostindustria:
@mdanylov-sigma please, rebase the branch

PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCacheRenderers/InterstitialController.swift
```diff
@@ -127,12 +138,17 @@ public class InterstitialController:
     }
     
     func reportSuccess() {
+        scheduleExpirationIfNeeded()
```
OlenaPostindustria, line 138: do we extend the bid's lifetime by the load time meaning the countdown starts when the ad is loaded rather than from the auction timestamp? ORTB defines `exp` as  `Advisory as to the number of seconds the bidder is willing to wait between the auction and the actual impression.`

PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCacheRenderers/InterstitialController.swift
```diff
@@ -142,6 +158,30 @@ public class InterstitialController:
...
+    }
+    
+    private func expireAd() {
+        guard !isExpired else { return }
+        
+        isExpired = true
+        if !isDisplayed {
+            adViewManager = nil
+        }
+        loadingDelegate?.interstitialControllerDidExpire?(self)
```
OlenaPostindustria, line 177: what if the interstitial ad was already displayed? should we fire the `didExpire` callback?

PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCacheRenderers/InterstitialController.swift
YuriyVelichkoPI: we can omit complex calculations here for now. 

exp values should be - "minutes" 
bid traveling from the ssp/dsp - several seconds  

so inaccuracy shouldn't be critical. 

Moreover, we don't know the actual auction timestamp to make an accurate calculation

PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCacheRenderers/InterstitialController.swift
YuriyVelichkoPI: no - there is no reason to fire the didEpireCallback if impression is already tracked.

YuriyVelichkoPI:
@mdanylov-sigma Please review the task description. 

‼️ I've soften the requirements for Rendering API. 

Let's not destroy or invalidate bids in Rendering API. It can be a braking behaviour change unexpected for publishers. Let's just inform the app that the ad is expired. And that's all in the scope of this ticket. 

The `exp` field is a recommendation, and each bidder can manage it differently. If we destroy the bid - we can lose revenue. 

Sorry for the changes sin the last second, but they look reasonable.

PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCacheRenderers/InterstitialController.swift
```diff
@@ -142,6 +154,27 @@ public class InterstitialController:
...
+    
+    private func scheduleExpirationIfNeeded() {
+        guard expirationWorkItem == nil,
+              let exp = expirationInterval ?? bid.exp,
+              exp.doubleValue > 0 else {
+            return
+        }
+        
+        let workItem = DispatchWorkItem { [weak self] in
+            self?.expireAd()
```
OlenaPostindustria, line 167: if the ad was already displayed, there is no need to call expire ad

PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/InterstitialAdUnitDelegate.swift
```diff
@@ -40,4 +40,7 @@ public protocol InterstitialAdUnitDelegate: NSObjectProtocol {
 
     /// Called when user clicked the ad
     @objc optional func interstitialDidClickAd(_ interstitial: InterstitialRenderingAdUnit)
+    
+    /// Called when the loaded ad expired and can no longer be displayed.
```
OlenaPostindustria: this comment should probably also be updated

PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/RewardedAdUnitDelegate.swift
```diff
@@ -40,4 +40,7 @@ import Foundation
 
     /// Called when user clicked the ad
     @objc optional func rewardedAdDidClickAd(_ rewardedAd: RewardedAdUnit)
+    
+    /// Called when the loaded ad expired and can no longer be displayed.
```
OlenaPostindustria: comment needs to be updated

PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCacheRenderers/DisplayView.swift
```diff
@@ -55,9 +58,13 @@ class DisplayView: UIView, PrebidMobileDisplayViewProtocol, AdViewManagerDelegat
     public required init?(coder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
+    
+    deinit {
+        expirationWorkItem?.cancel()
```
OlenaPostindustria, line 64: we should also cancel timer once the impression happens, according to the ORTB

OlenaPostindustria (changes requested):

OlenaPostindustria:
should we also implement this feature inside the adapters?

YuriyVelichkoPI:
> should we also implement this feature inside the adapters?

@OlenaPostindustria  Could you clarify the question? 

Since there is no respective callbacks in Adpaters' interfaces, I'm not sure what exactly should we do in adapters on expiration event.

OlenaPostindustria:
@YuriyVelichkoPI For example, the AppLovin SDK provides an `MAAdapterError` that could theoretically be used when `adDidExpire` is triggered. However, since we agreed that `adDidExpire` is intended only to notify the publisher that the ad has expired, using it to report an error would not be appropriate.
Screenshot 2026-07-23 at 19 14 57

YuriyVelichkoPI:
Let's keep the adpateers as is then.

OlenaPostindustria:
@mdanylov-sigma could you please rebase the branch?

OlenaPostindustria (dismissed):
LGTM

PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCacheRenderers/InterstitialController.swift
thanhhaidev: Hi @YuriyVelichkoPI,

I have one question: Why don't we schedule it when the bid response is received? I think that's the point in time that's closest to when the actual action occurs.

YuriyVelichkoPI: @thanhhaidev you mean that currently we schedule a expiration timer _after_ the ad is rendered and rendering consumes time, so we should schedule a timer _before_ this, right? 

if so - yes it make sense to me. 

it shouldn't affect timings a lot, since even if rendering takes seconds the expiration time as usual not lower then 5 minutes, and this number is `Advisory` from bidders. But from technical perspective, to start the timer as close to the auction as possible makes perfect sense.

PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/BannerView.swift
```diff
@@ -406,6 +406,21 @@ public class BannerView:
         }
     }
     
+    private func reportAdExpired() {
+        DispatchQueue.main.async { [weak self] in
+            guard let self else { return }
+            
+            self.autoRefreshManager?.cancelRefreshTimer()
+            self.deployedView?.removeFromSuperview()
+            self.deployedView = nil
```
YuriyVelichkoPI, line 421: Per the amended requirements in #850, banner expiration should "refresh the unit" if refreshable, and otherwise **do nothing**. Here, `deployedView` is unconditionally removed from its superview (and `autoRefreshManager`'s timer cancelled) regardless of `refreshInterval`/`isRefreshStopped`, and only the *refresh* itself is gated by that check below.

For a non-refreshable banner, this silently blanks the ad slot with no replacement — which isn't "do nothing." Please gate the `deployedView` teardown (and refresh-timer cancellation) on the same refreshable condition used for `adLoadFlowController?.refresh()`, so a non-refreshable banner keeps showing its last (expired) creative untouched, only notifying the delegate via `bannerViewDidExpire`.

YuriyVelichkoPI (changes requested):
Reviewed against the requirements in #850. Overall the `bid.exp` expiration wiring (DisplayView/InterstitialController timer, delegate propagation through BannerView/BaseInterstitialAdUnit/RewardedAdUnit) is solid and well covered by tests.

One behavioral issue worth fixing before merge — see inline comment on `BannerView.reportAdExpired()`.

YuriyVelichkoPI (approved):
LGTM

Issue #1220, opened 2025-11-03
DisplayViewLoadingDelegate fires displayViewDidLoadAd with empty view


MattMur:
**Describe the bug**
I'm writing a custom renderer, trying to leverage as much of the existing architecture as possible. However currently the DisplayView fires the `DisplayViewLoadingDelegate` protocol `displayViewDidLoadAd` with an empty view (DisplayView has no ad, or subviews). This is counter-intuitive and feels like a bug considering that if the ad truly did load, then it should not be empty.

It seems the source of the problem initiates from the method [onTransactionIsReady:](https://github.com/prebid/prebid-mobile-ios/blob/12d6284a9c763a8f9a0d9b0b9c2b2686824eecf1/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMAdViewManager.m#L345C1-L367C2) where both `setupCreative:` and `adViewManagerDelegate adLoaded:` are both dispatched async to the main thread. However later down the line in the method  `show:` at the point where the actual creative view gets injected into the DisplayView, it for some reason [again dispatches to the main thread.](https://github.com/prebid/prebid-mobile-ios/blob/12d6284a9c763a8f9a0d9b0b9c2b2686824eecf1/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMAdViewManager.m#L113-L120) The problem is that by moving the creative insertion to the top of the call stack with `dispatch_async`, it preempts `displayViewDidLoadAd` and causes it to be called first, before the method `show` has actually finished loading the ad.

**To Reproduce**
1. Set breakpoint on 'BannerAdLoader' line 77
2. Run PrebidDemoSwift with In-App Display Banner rendering
3. See that displayView is empty, not loaded

**Expected behavior**
Parameter `displayView: UIView` subviews should not be empty. `displayView` should contain a PBMWebView.

**Smartphone (please complete the following information):**
 - Device: iPhone 17
 - OS: 26

**Solutions**
The best option to me would be to simply postpone the call to [[self.adViewManagerDelegate adLoaded:[transaction getAdDetails]]](https://github.com/prebid/prebid-mobile-ios/blob/12d6284a9c763a8f9a0d9b0b9c2b2686824eecf1/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMAdViewManager.m#L365C1) and instead call it **after the creative has been injected** on line 120.
[...]

PR #1269, opened 2026-05-19
Fix load callbacks firing before creative attachment

mdanylov-sigma:
#1220 

# Summary

Fixed the banner/custom renderer load callback ordering so the load callback is not fired before the creative view is attached.

- Updated `PBMAdViewManager.m` so banner creative insertion runs immediately when already on the main thread. If called off-main, it still dispatches to the main queue. This prevents `adLoaded` / `displayViewDidLoadAd` from firing before the creative view has been added to the display view.
- Added a regression test in `AdViewManagerTest.swift` to verify that the display view already contains the creative when `adLoaded` fires.

### DIFF
--- a/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMAdViewManager.m
+++ b/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMAdViewManager.m
@@ -32,6 +32,8 @@
 @property (nonatomic, nullable, readonly) id<PBMTransaction> currentTransaction; // computed
 @property (nonatomic, assign) BOOL videoInterstitialDidClose;
 
+- (void)displayCreativeView:(UIView *)creativeView rootViewController:(UIViewController *)viewController;
+
 @end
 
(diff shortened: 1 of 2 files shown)


PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMAdViewManager.m
```diff
@@ -32,6 +32,8 @@ @interface PBMAdViewManager_Objc: NSObject <PBMAdViewManager>
 @property (nonatomic, nullable, readonly) id<PBMTransaction> currentTransaction; // computed
 @property (nonatomic, assign) BOOL videoInterstitialDidClose;
 
+- (void)displayCreativeView:(UIView *)creativeView rootViewController:(UIViewController *)viewController;
```
OlenaPostindustria: this declaration seems to be unnecessary since the method is private

OlenaPostindustria (dismissed):
LGTM! left minor suggestion

OlenaPostindustria (approved):

PR #1272, opened 2026-06-01
Fix clickthrough handling in PBMWebView to ensure iframes are able to load content

MattMur:
I found an issue where iframes loading within an PBMWebView after an `mraid.expand()` call could cause the iframe's navigation to trigger `receivedClickthroughLink` and cause the iframe content to clickout to Safari.

### Steps
1. Ad content loads
2. User clicks the ad, `[self wasRecentlyTapped]` is now true
3. mraid.expand() is called
4. Ad content loads within the expanded PBMWebView, 
5. Within the ad content there may be iframes (social widgets in our case) that load
6. iframe navigation will trigger `webView: decidePolicyForNavigationAction:` which will call `[self.delegate webView:self receivedClickthroughLink:url]` and trigger an inadvertent clickout.

### Fix

To prevent this we can first check two things first in order to determine if the URL is meant to be clicked out. 
First, Is the link coming from the main frame. OR is navigationType == WKNavigationTypeLinkActivated. If either are true, then we can proceed with click through logic. Otherwise we can simply allow the content to load. This should allow iframes to load content without without interfering with clickout logic, while also still enforcing auto-click protection on the main frame.

### DIFF
--- a/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMWebView.m
+++ b/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMWebView.m
@@ -326,19 +326,25 @@ static NSString * const KeyPathOutputVolume = @"outputVolume";
     }
     
     //Prevent malicious auto-clicking
-    if ([self wasRecentlyTapped]) {
-        //Open clickthrough
-        @weakify(self);
-        dispatch_async(dispatch_get_main_queue(), ^{
-            @strongify(self);
-            if (!self) { return; }
-            [self.delegate webView:self receivedClickthroughLink:url];
-        });
+    BOOL isMainFrame = navigationAction.targetFrame != nil && navigationAction.targetFrame.isMainFrame;
+    BOOL isLinkActivated = navigationAction.navigationType == WKNavigationTypeLinkActivated;
+    if (isMainFrame || isLinkActivated) {
+        if ([self wasRecentlyTapped]) {
+            //Open clickthrough
+            @weakify(self);
+            dispatch_async(dispatch_get_main_queue(), ^{
+                @strongify(self);
+                if (!self) { return; }
+                [self.delegate webView:self receivedClickthroughLink:url];
+            });
+        } else {
+            PBMLogWarn(@"User has not recently tapped. Auto-click suppression is preventing navigation to: %@", url);
+        }
+        decisionHandler(WKNavigationActionPolicyCancel);
     } else {
-        PBMLogWarn(@"User has not recently tapped. Auto-click suppression is preventing navigation to: %@", url);
+        // Allow iframe navigations that aren't user-initiated link taps
+        decisionHandler(WKNavigationActionPolicyAllow);
     }
-    
-    decisionHandler(WKNavigationActionPolicyCancel);
 }
 
 - (void)webView:(WKWebView *)webView didFinishNavigation:(WKNavigation *)navigation {


mdanylov-sigma (changes requested):
The bug makes sense: iframe navigation can be incorrectly treated as clickout after `mraid.expand()`.

But the fix as written seems broader than the bug. It allows any navigation that is not main-frame and not `WKNavigationTypeLinkActivated` to load unconditionally, bypassing `wasRecentlyTapped` and the `receivedClickthroughLink:` delegate entirely. 
Since this is click-fraud-sensitive code, I think we should tighten this before merging.

A few concerns:

- `navigationType == WKNavigationTypeOther` alone is not a safe signal. Programmatic redirects, `window.location` changes, and meta-refresh navigations can also report as `.other`, so “iframe + other” can still leave a bypass.
- Same-origin-only may be too strict, since ad creatives often load cross-origin iframe content, trackers, and CDN resources.
- `targetFrame == nil` should not be treated as iframe navigation. Right now `isMainFrame` is computed as `targetFrame != nil && targetFrame.isMainFrame`, so a nil `targetFrame` falls into the allow branch. That can represent popup/new-window navigation and seems like an unintended side effect.

Can we narrow the allow branch to real subframe content navigations only, e.g. require:

- `targetFrame != nil`
- `!targetFrame.isMainFrame`
- `navigationType != WKNavigationTypeLinkActivated`
- URL scheme is limited to expected web/content schemes, if that is sufficient for the iframe cases being fixed

Main-frame, link-activated, popup/new-window (`targetFrame == nil`), and unexpected-scheme navigations should remain click-gated or cancelled.

Please also it would be nice to have regression coverage for `decidePolicyForNavigationAction:`. At minimum, the tests proving that `targetFrame == nil` and unexpected-scheme subframe navigations do not bypass `wasRecentlyTapped` suppression.

YuriyVelichkoPI:
@MattMur, could you please review @mdanylov-sigma’s comment? Let us know whether you think the requested changes should be addressed in a separate issue, while this one remains scoped to the specific case described.

MattMur:
Hi @mdanylov-sigma  thank you for reviewing this. I finally found time to come back to this and address the valid concerns pointed out. I created a new method `isSafeSubframeNavigationWithTargetFrame` that I think addresses each  concern. Also as requested I've added new tests to ensure various use cases are caught.

mdanylov-sigma (commented):
Thanks for the update, @MattMur. The latest commit follows the shape I asked for in June: popups (nil `targetFrame`), main-frame navigations, `.linkActivated` and unexpected schemes all still need a recent tap, and there is a pure helper with tests. I merged it into current master (no conflicts) and ran it on the iOS 26.5 simulator. Two problems still need fixing, and there is one trade-off to decide on.

### Blockers

**1. Four of the new Swift tests crash**

These crash with SIGTRAP when the test finishes:
- `testDecidePolicyForNavigationActionAllowsSubframeContentWithoutTap`
- `testDecidePolicyForNavigationActionRejectsMainFrame`
- `testDecidePolicyForNavigationActionRejectsSubframeLinkActivated`
- `testDecidePolicyForNavigationActionRejectsSubframeWithUnexpectedScheme`

```
CoreFoundation     CFRetain
WebKit             -[WKFrameInfo dealloc]
PrebidMobileTests  outlined destroy of WKFrameInfo?
PrebidMobileTests  @objc MockWKNavigationAction.__ivar_destroyer
WebKit             -[WKNavigationAction dealloc]
PrebidMobileTests  PBMWebViewTest.testDecidePolicyForNavigationActionRejectsMainFrame()
```

[`MockWKFrameInfo`](https://github.com/prebid/prebid-mobile-ios/blob/c4828384fae49ef983d02928ea60c02881796a29/PrebidMobileTests/RenderingTests/Tests/PBMWebViewTest.swift#L25) subclasses `WKFrameInfo`, but an instance that WebKit didn't create lacks WebKit's internal storage, so `-[WKFrameInfo dealloc]` crashes when the mock is released. The nil-frame test passes because it never creates one. Keeping the mocks alive (for example in a static array) makes all five pass. A cleaner option is to drop `MockWKFrameInfo`, cover the frame/type/scheme cases through `isSafeSubframeNavigationWithTargetFrame:...` (the ObjC helper tests already pass), and keep only nil-frame integration tests. Seen on Xcode 26.5; I haven't checked CI's Xcode 16.4.

PR CI won't catch this: [`PrebidMobilePRTests.xctestplan`](https://github.com/prebid/prebid-mobile-ios/blob/2befc557e5922511b21b161fe2a5c2f26847cc7d/PrebidMobileTests/PrebidMobilePRTests.xctestplan#L454-L458) skips `PBMWebViewObjCTest` and `PBMWebViewTest` as whole classes, so none of the new tests run on PRs. It would first fail on a `bump-to` full run.

**2. Iframes that don't use http(s) still click out after a tap**

I checked with a `WKWebView` probe on macOS WebKit. Iframes added after the ad loads reach `decidePolicy` as subframe navigations with type `.other` for `about:blank` (including the common "create an iframe, then `document.write`" pattern used by ad-tech, verification and social scripts), `about:srcdoc`, `data:` and `blob:`. The [http/https-only list](https://github.com/prebid/prebid-mobile-ios/blob/c4828384fae49ef983d02928ea60c02881796a29/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMWebView.m#L357-L369) sends all of them to the tap check. Calling `decidePolicy` right after a tap on the iOS simulator:

| Navigation after a tap | Policy | Clickthrough |
|---|---|---|
| subframe, `.other`, `https` | allow | none |
| subframe, `.other`, `about:blank` | cancel | `about:blank` |
| subframe, `.other`, `about:srcdoc` | cancel | `about:srcdoc` |
| subframe, `.other`, `data:` | cancel | `data:text/html,...` |
| subframe, `.linkActivated`, `https` | cancel | fires |
| nil frame (popup), `https` | cancel | fires |
| main frame, `https` | cancel | fires |

So the iframe never loads, and a clickthrough fires anyway. On the default path `PBMSafariVCOpener` [rejects non-http(s) URLs](https://github.com/prebid/prebid-mobile-ios/blob/2befc557e5922511b21b161fe2a5c2f26847cc7d/PrebidMobile/Objc/PrebidMobileRendering/Prebid/PBMCore/PBMSafariVCOpener.m#L97), but two earlier checks in `handleClickthrough` [ignore the URL](https://github.com/prebid/prebid-mobile-ios/blob/2befc557e5922511b21b161fe2a5c2f26847cc7d/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMAbstractCreative.m#L255-L270): an SKAdN bid opens the App Store sheet (`SKStoreProductViewController`), and a `clickHandlerOverride` counts a click (`creativeWasClicked` plus click tracking). That is the same kind of unintended clickout this PR is fixing.

Suggested fix: also let subframes load `about:blank`, `about:srcdoc`, `data:` and `blob:` without a tap. They load inside the iframe and can't open anything outside the app. Custom and deep-link schemes should still need a tap.

### Needs a decision

**JS-driven clicks inside iframes are now dropped.** A button inside an iframe with `onclick="location.href=clickTag"` also shows up as a subframe navigation with type `.other` (confirmed with the same probe). With this change the landing page loads inside the iframe, with no clickthrough and no click tracking; on master it opened a clickthrough. Anchor taps, `window.open` and `top.location` still need a tap. This comes directly from the "not `.linkActivated`" rule I suggested, so it's not a mistake on your side, but it is the "`.other` isn't a reliable signal" problem in reverse: real clicks get lost instead of fake ones getting through. If you think that's acceptable for real creatives, please at least document it in the helper comment and add a test for it.

### Suggestions

1. **Test the reported scenario.** None of the new tests call `recordTapEvent`, so none reproduces the actual bug (tap, then an iframe loads, then a clickout). Please add "tap + subframe https → `.allow`, no clickthrough", plus versions of the nil-frame and custom-scheme tests that tap first and check the clickthrough *does* fire, not just that the navigation is cancelled.
2. **Run the new tests on PRs.** Removing the class-level `"PBMWebViewTest"` and `"PBMWebViewObjCTest"` entries from `PrebidMobilePRTests.xctestplan` would enable only the new tests, since the existing methods are already skipped one by one.
3. **Update the PR description.** It still describes the first commit's "main frame OR link activated" logic.

### Nits

- In `isSafeSubframeNavigationWithTargetFrame:(BOOL)hasTargetFrame`, the name says "frame" but the value is a BOOL, and `isMainFrame` depends on it, so the impossible pair `(NO, YES)` is accepted. Rename it, or pass the `WKFrameInfo *`.
- The colons in the multi-line ObjC calls and declarations aren't aligned.
- The ObjC helper tests and the Swift tests cover the same cases. With the Swift migration underway, keeping only the Swift ones seems better.
- "see SDK-iframe-clickout fix" in the Swift test comment refers to a ticket outside this repo.

<details><summary>How I checked</summary>

- Merged this PR into master `2befc557` (no conflicts) and built `PrebidMobileTests` on Xcode 26.5 / iOS 26.5 simulator. No new warnings.
- New ObjC helper tests: 5/5 pass. New Swift tests: 1 passes, 4 crash (all 5 pass once the mocks are kept alive).
- Which navigations WebKit reports: a macOS `WKWebView` loading iframes added after load, plus clicks on an anchor, a `location.href` button and a `top.location` button inside a `srcdoc` iframe.
- Policy and clickthrough after a tap: direct `decidePolicy` calls on the iOS simulator (table above).
- Not run: `buildPrebidMobile.sh`, SwiftLint.

</details>

MattMur:
Hi @mdanylov-sigma that's interesting, I'm running Xcode 26.2 and they all pass for me. I'll remove them though if they're not needed. 

The **JS-driven clicks inside iframes are now dropped** issue is a tricky one. For now, I'm going to simply limit the scope of this PR to when `mraidState = .expanded`. This narrows the JS-driven-click issue to the expanded state only so that JS-driven clicks and other iframe navigations still work as they did previously. Added a comment to document this potential issue which I hope will suffice.

mdanylov-sigma (commented):
Thanks @MattMur, the two new commits address most of my last review:
- The crashing `MockWKFrameInfo` tests are gone. The crash was inside WebKit's `dealloc`, so it may well depend on the SDK version, and 26.2 passing is plausible.
- `[WKWebView handlesURLScheme:]` is a good allowlist. On both macOS and the iOS 26.5 simulator it returns YES only for schemes WebKit renders itself (http(s), about, data, blob, file, javascript, ftp, ws(s), applewebdata) and NO for `mailto`, `tel`, `sms`, `itms-*`, `maps`, `facetime`, `mraid`, custom schemes and a nil scheme.
- Limiting the bypass to `mraidState == expanded` is a reasonable way to contain the JS-click limitation, and the doc comment explains it. Outside the expanded state the behaviour is now exactly master's. A follow-up issue for the expanded-state limitation would keep it tracked.

I merged `d555a1dd` into master `2befc557` and ran it on the iOS 26.5 simulator. Two things are left.

### Blockers

**1. CI can't build the tests**

[`PBMWebViewTest.swift:543`](https://github.com/prebid/prebid-mobile-ios/blob/d555a1dd87471d7f12f75cc4af785d8dfc7824f1/PrebidMobileTests/RenderingTests/Tests/PBMWebViewTest.swift#L543) calls `webView.tapdownGestureRecognizer`, which master renamed to `tapGestureRecognizer` in #1294. The branch is 29 commits behind master and CI builds the merge, so **Run Quick Unit Tests for Frameworks** stops at:

```
PBMWebViewTest.swift:543:40: error: value of type 'PBMWebView' has no member 'tapdownGestureRecognizer'
** TEST BUILD FAILED **
```

Merging master and switching to `tapGestureRecognizer` fixes it. With that one change the merged tree builds with no new warnings, all 9 new ObjC helper tests pass, both new Swift popup tests pass, and the existing `decidePolicy` tests still pass.

**2. Please confirm the one-part expand flow is covered**

The steps in the PR description (tap, `mraid.expand()`, iframe clicks out) match a one-part expand, where the same web view is reused and keeps the recorded tap. In that flow the web view's `mraidState` doesn't become `expanded` right away:
- the modal push completion only sets `delayedMraidState = expanded` ([PBMMRAIDController.m:365-369](https://github.com/prebid/prebid-mobile-ios/blob/2befc557e5922511b21b161fe2a5c2f26847cc7d/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMMRAIDController.m#L365-L369)),
- which is applied to the web view in `webView:exposureChange:` ([PBMMRAIDController.m:210-214](https://github.com/prebid/prebid-mobile-ios/blob/2befc557e5922511b21b161fe2a5c2f26847cc7d/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMMRAIDController.m#L210-L214)), on the next exposure check (polled every 200 ms, [PBMWebView.m:839](https://github.com/prebid/prebid-mobile-ios/blob/2befc557e5922511b21b161fe2a5c2f26847cc7d/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMWebView.m#L839)).

Until then `isExpanded` is `NO`, so an iframe that loads within 5 s of the tap still clicks out. Iframes created in a `stateChange` listener are fine, because `changeToMRAIDState:` sets the native state before firing the JS event ([PBMWebView.m:642-645](https://github.com/prebid/prebid-mobile-ios/blob/2befc557e5922511b21b161fe2a5c2f26847cc7d/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMWebView.m#L642-L645)). Iframes created straight after calling `mraid.expand()` are not. Two-part expand (`mraid.expand(url)`) is covered from the start, since the new web view sets `expanded` when MRAID is injected ([PBMWebView.m:499](https://github.com/prebid/prebid-mobile-ios/blob/2befc557e5922511b21b161fe2a5c2f26847cc7d/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMWebView.m#L499)), but that web view has no recorded tap anyway.

Could you check when the social widgets load in your creative? If they load right after `mraid.expand()` rather than after the `stateChange` event, this change won't fix your original case yet.

### Decisions after a tap

Direct `decidePolicy` calls on the merged build (iOS 26.5 simulator):

| State | Navigation | Result |
|---|---|---|
| expanded | subframe `.other`: `https`, `about:blank`, `about:srcdoc`, `data:` | allowed, no clickthrough |
| expanded | subframe `.other`: `itms-apps:` | clickthrough |
| expanded | subframe `.linkActivated`, popup (nil frame), main frame | clickthrough |
| expanded, no tap | subframe `.other`: `https` | allowed |
| default | subframe `.other`: `https` or `about:blank` | clickthrough (same as master) |

### Suggestions

1. **Run the new tests on PRs.** `PrebidMobilePRTests.xctestplan` still skips `PBMWebViewTest` and `PBMWebViewObjCTest` as whole classes ([lines 454-458](https://github.com/prebid/prebid-mobile-ios/blob/2befc557e5922511b21b161fe2a5c2f26847cc7d/PrebidMobileTests/PrebidMobilePRTests.xctestplan#L454-L458)), so CI only compiles them.
2. **Update the PR description.** It still describes the first commit; the expanded-state scope and the `handlesURLScheme:` check are worth mentioning.
3. **Consider `isEqual:` for the state check** ([PBMWebView.m:331](https://github.com/prebid/prebid-mobile-ios/blob/d555a1dd87471d7f12f75cc4af785d8dfc7824f1/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMWebView.m#L331)). `==` works today because only the six shared `MRAIDState` instances exist, but the class defines value equality and `PBMMRAIDController` uses both styles, so `[self.mraidState isEqual:PBMMRAIDState.expanded]` is sturdier.

### Nits

- The [doc comment](https://github.com/prebid/prebid-mobile-ios/blob/d555a1dd87471d7f12f75cc4af785d8dfc7824f1/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMWebView.m#L357-L364) says "an iframe loading content inside the button"; it's the landing page loading inside the iframe.
- The comment at [PBMWebViewTest.swift:505-506](https://github.com/prebid/prebid-mobile-ios/blob/d555a1dd87471d7f12f75cc4af785d8dfc7824f1/PrebidMobileTests/RenderingTests/Tests/PBMWebViewTest.swift#L505-L506) belonged to the deleted `AllowsSubframeContentWithoutTap` test.
- `testDecidePolicyForNavigationActionAcceptsPopupWithNilTargetFrame` expects `.cancel` plus a clickthrough, so "Accepts" reads backwards; something like `...OpensClickthroughForPopupInDefaultState` would be clearer.
- Still open from last time: the `isSafeSubframeNavigationWithTargetFrame:(BOOL)hasTargetFrame` naming and the unaligned ObjC colons.

<details><summary>How I checked</summary>

- Merged `d555a1dd` into master `2befc557` (no conflicts), renamed the one `tapdownGestureRecognizer` call, and built `PrebidMobileTests` on Xcode 26.5 / iOS 26.5 simulator.
- `handlesURLScheme:` results: a small ObjC program run on macOS and inside the iOS 26.5 simulator.
- Decision table: direct `decidePolicy` calls with mocked navigation actions (frame mocks kept alive to avoid the `WKFrameInfo` dealloc crash).
- The expand timing comes from reading `PBMMRAIDController` and `PBMWebView` on master; I did not run a full MRAID expand end to end.
- `testEvaluateJSNil` fails locally for environmental reasons and is outside the PR plan. Not run: `buildPrebidMobile.sh`, SwiftLint.

</details>

PR #1275, opened 2026-07-04
Resolves crash related to nullable advertisingIdentifier

### LINKED ISSUE #1273: Crash (EXC_BREAKPOINT) in DeviceAccessManager.advertisingIdentifier() when ASIdentifierManager returns nil
**Describe the bug**

DeviceAccessManager.advertisingIdentifier() force-bridges ASIdentifierManager.shared().advertisingIdentifier into a non-optional Swift UUID:

    // PrebidMobile/Swift/PrebidMobileRendering/Utilities/DeviceAccessManager.swift:85
    @objc public func advertisingIdentifier() -> String {
        ASIdentifierManager.shared().advertisingIdentifier.uuidString
    }

Apple annotates -[ASIdentifierManager advertisingIdentifier] as nonnull, so it bridges to a non-optional UUID in Swift. But in rare device states the ObjC property returns nil (the identifier is fetched over XPC from a system daemon that can be unavailable — e.g. very early in launch before the AdSupport service is ready, just after boot / before first unlock, under memory pressure / XPC failure, or under certain MDM / Screen Time restrictions). When it returns nil, the implicit NSUUID? -> UUID bridge calls UUID._unconditionallyBridgeFromObjectiveC, which traps -> EXC_BREAKPOINT and the app crashes.

Note: normal ATT denial / notDetermined does NOT cause this — that returns the all-zeros UUID (00000000-0000-0000-0000-000000000000), which bridges fine. The crash is specifically the nil case, where the OS returns nil despite the nonnull annotation.

This is hit whenever a bid request is built with Targeting.shared.isAllowedAccessDeviceData == true (PBMDeviceInfoParameterBuilder.m:88 reads deviceAccessManager.advertisingIdentifier), so any fetchDemand() call can crash.

**To Reproduce**
1. Force ASIdentifierManager.shared().advertisingIdentifier to return nil (e.g. swizzle the getter for testing).
2. Ensure Targeting.shared.isAllowedAccessDeviceData returns true.
3. Trigger a bid request via BannerAdUnit.fetchDemand(...).
4. See error - app crashes with EXC_BREAKPOINT in DeviceAccessManager.advertisingIdentifier() at DeviceAccessManager.swift:85.

**Expected behavior**

The SDK should tolerate a nil advertising identifier (fall back to an empty string / omit the ifa param) instead of trapping. Suggested fix - treat the value as optional:

    @objc public func advertisingIdentifier() -> String {
        ASIdentifierManager.shared().advertisingIdentifier?.uuidString ?? ""
    }

**Screenshots**

Symbolicated crash stack (abridged):

    Crashed: com.apple.main-thread
    0  Foundation      UUID._unconditionallyBridgeFromObjectiveC + 196
    1  PrebidMobile    DeviceAccessManager.advertisingIdentifier()  (DeviceAccessManager.swift:85)
    2  PrebidMobile    -[PBMDeviceInfoParameterBuilder buildBidRequest:]  (PBMDeviceInfoParameterBuilder.m:88)
    3  PrebidMobile    +[PBMParameterBuilderService buildParamsDictWithAdConfiguration:...]
    5  PrebidMobile    -[PBMBidRequester_Objc getRTBRequest]  (PBMBidRequester.m:180)
    6  PrebidMobile    -[PBMBidRequester_Objc makeRequestWithCompletion:]  (PBMBidRequester.m:75)
    9  PrebidMobile    UserAgentService.fetchUserAgent(completion:)  (UserAgentService.swift:37)
    11 PrebidMobile    -[PBMBidRequester_Objc requestBidsWithCompletion:]  (PBMBidRequester.m:55)
    12 PrebidMobile    AdUnit.baseFetchDemand(adObject:completion:)  (AdUnit.swift:182)
    13 PrebidMobile    AdUnit.fetchDemand(completionBidInfo:)  (AdUnit.swift:113)
    14 <app>           BannerAdUnit.fetchDemand { ... }

 - OS: N/A (iOS SDK issue, not desktop)
 - Browser: N/A
 - Version: N/A

 - Device: Multiple (47 distinct devices affected in production)
 - OS: iOS (observed on iOS 18.7)
 - Browser: N/A (native app, in-app ad rendering)
 - Version: PrebidMobile 3.3.1 (same unguarded code also present on master)

**Additional context**

Low but persistent volume in our production Crashlytics - 50 occurrences across 47 devices (0.6% of crashing users). The root cause is entirely inside the SDK: apps cannot prevent ASIdentifierManager from returning nil (it comes from a system daemon being momentarily unavailable), so this can only be fixed upstream by making the bridge nil-safe.

mdanylov-sigma:
## Summary

Fixes a crash (`EXC_BREAKPOINT` in `UUID._unconditionallyBridgeFromObjectiveC`) that occurs when `ASIdentifierManager.shared().advertisingIdentifier` returns `nil` at the Objective-C runtime level, despite Apple's header annotating the property as `nonnull`.

Closes #1273.

## Root cause

`-[ASIdentifierManager advertisingIdentifier]` is declared `nonnull` in Apple's headers, so Swift imports it as a non-optional `UUID`. In rare device states (XPC/daemon unavailable, early boot, MDM/Screen Time restrictions, memory pressure) the method returns `nil` anyway. When Swift accesses the property, the compiler-generated bridging thunk calls `UUID._unconditionallyBridgeFromObjectiveC` to force the ObjC `NSUUID?` into a Swift `UUID`, and that call traps on `nil`.

## Why not the fix suggested in the issue

The issue suggests:

    @objc public func advertisingIdentifier() -> String {
        ASIdentifierManager.shared().advertisingIdentifier?.uuidString ?? ""
    }

This doesn't actually fix the crash, and doesn't compile as written:

- Because the property is imported as **non-optional** `UUID` (due to the `nonnull` annotation), `?.` can't be applied to it - Swift will reject optional-chaining on a non-optional type.
- More importantly, the trap doesn't happen at the point of consuming the value - it happens **inside the property access itself**, in the synthesized ObjC→Swift bridging thunk. By the time you could apply `??`, the crash has already occurred. Wrapping the *result* in optional-handling can't help if the *access* is what crashes.

In short: this is a case of an API lying about its nullability, and the fix has to happen below the level where Swift's automatic bridging kicks in - not after it.

## Fix implemented

`DeviceAccessManager.advertisingIdentifier()` now bypasses the synthesized property getter entirely and invokes the underlying ObjC selector manually via `NSObject.perform(_:)`:

    @objc public func advertisingIdentifier() -> String {
        let advertisingIdentifier = ASIdentifierManager.shared()
            .perform(#selector(getter: ASIdentifierManager.advertisingIdentifier))?
            .takeUnretainedValue() as? NSUUID
        return advertisingIdentifier?.uuidString ?? ""
    }

`perform(_:)` sends the message directly (effectively `objc_msgSend`) and returns `Unmanaged<AnyObject>?`, without going through the compiler's `nonnull`-enforcing bridge. This lets a genuine `nil` return value surface safely as `Unmanaged<AnyObject>?.none`, which we then optionally cast to `NSUUID` and unwrap with `??`. No trap is possible regardless of what the underlying daemon returns.

### Secondary fix: empty `ifa` no longer blocks the `ifv` fallback

`PBMDeviceInfoParameterBuilder.buildBidRequest:` treats an empty-string `ifa` the same as `nil`:

    NSString *ifa = [Targeting.shared isAllowedAccessDeviceData] ? self.deviceAccessManager.advertisingIdentifier : nil;
    if (ifa.length == 0) {
        ifa = nil;
    }

Previously, an empty (but non-nil) `ifa` would fail the `!ifa` check used to decide whether to fall back to `device.extAtts.ifv` (identifierForVendor), so a bid request could go out with neither a usable `ifa` nor an `ifv`. Now an empty identifier is normalized to `nil`, so the existing `ifv` fallback logic engages correctly.

### DIFF
--- a/PrebidMobile/Objc/PrebidMobileRendering/Networking/Parameters/PBMDeviceInfoParameterBuilder.m
+++ b/PrebidMobile/Objc/PrebidMobileRendering/Networking/Parameters/PBMDeviceInfoParameterBuilder.m
@@ -86,6 +86,9 @@
     NSNumber *lmt = @(!self.deviceAccessManager.advertisingTrackingEnabled);
     
     NSString *ifa = [Targeting.shared isAllowedAccessDeviceData] ? self.deviceAccessManager.advertisingIdentifier : nil;
+    if (ifa.length == 0) {
+        ifa = nil;
+    }
     
     bidRequest.device.lmt = lmt;
     bidRequest.device.ifa = ifa;
--- a/PrebidMobile/Swift/PrebidMobileRendering/Utilities/DeviceAccessManager.swift
+++ b/PrebidMobile/Swift/PrebidMobileRendering/Utilities/DeviceAccessManager.swift
@@ -82,7 +82,10 @@ public class DeviceAccessManager: NSObject {
     // MARK: - IDFA & Tracking
 
     @objc public func advertisingIdentifier() -> String {
-        ASIdentifierManager.shared().advertisingIdentifier.uuidString
+        let advertisingIdentifier = ASIdentifierManager.shared()
+            .perform(#selector(getter: ASIdentifierManager.advertisingIdentifier))?
+            .takeUnretainedValue() as? NSUUID
+        return advertisingIdentifier?.uuidString ?? ""
     }
 
     @objc public func advertisingTrackingEnabled() -> Bool {
--- a/PrebidMobileTests/RenderingTests/Mocks/MockDeviceAccessManager.swift
+++ b/PrebidMobileTests/RenderingTests/Mocks/MockDeviceAccessManager.swift
@@ -19,13 +19,14 @@ import CoreGraphics
 @testable import PrebidMobile
 
 fileprivate let advertisingTrackingEnabledDefault = true
+fileprivate let advertisingIdentifierDefault = "abc123"
 fileprivate let defaultUserLanguageCode = "ml"
 
 class MockDeviceAccessManager: DeviceAccessManager {
     
     static let nullUUID = "00000000-0000-0000-0000-000000000000"
     static let mockIdentifierForVendor = "B78D99E3-5BD1-49AF-A669-0D77B464C5B9"
-    static let mockAdvertisingIdentifier = "abc123"
+    static var mockAdvertisingIdentifier = advertisingIdentifierDefault
     static var mockAdvertisingTrackingEnabled = advertisingTrackingEnabledDefault
     @available(iOS 14, *)
     static var mockAppTrackingTransparencyStatus = ATTrackingManager.AuthorizationStatus.notDetermined
@@ -84,6 +85,7 @@ class MockDeviceAccessManager: DeviceAccessManager {
     }
     
     class func reset() {
+        self.mockAdvertisingIdentifier = advertisingIdentifierDefault
         self.mockAdvertisingTrackingEnabled = advertisingTrackingEnabledDefault
         self.mockUserLanguageCode = defaultUserLanguageCode
     }
--- a/PrebidMobileTests/RenderingTests/Tests/PBMDeviceAccessManagerTests.swift
+++ b/PrebidMobileTests/RenderingTests/Tests/PBMDeviceAccessManagerTests.swift
@@ -16,6 +16,8 @@
 import Foundation
 import XCTest
 import UIKit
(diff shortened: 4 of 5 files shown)


OlenaPostindustria (approved):
LGTM! 👍🏻

Issue #950, opened 2024-01-11
Multi-AdUnit Auction and Winner Selection Logic


bretg:
**Is your feature request related to a problem? Please describe.**

I learned two things about Prebid SDK that are not aligned with other Prebid products:

1.  SDK currently supports doing auctions with only one AdUnit at a time. Doing one auction at a time is poor performance and costs everyone time and money.
2. In "Rendering" mode, SDK selecting a winner without the ability to override. No other part of Prebid makes a winning bid decision that cannot be overridden by publisher code. Use cases include: choosing deals over open market bids and [limited bid caching](https://docs.prebid.org/dev-docs/faq.html#does-prebidjs-cache-bids).

**Describe the solution you'd like**

I would like to propose requirements for a future major version of the SDK to align it with other Prebid products:

1. Support the ability to call for multiple AdUnits at the same time.
2. Support the ability for app developers to provide alternate bid-choosing logic. i.e. don't tie the auction, winner selection, and render together into one inseparable interface.

There are many ways to design these features into the SDK. Will leave that task to the people who know mobile SDKs better.

mdanylov-sigma:
This issue's two asks are split across the proposal: multi-impression auctions (one ORTB request, multiple imp[] entries - **Phase 4)** and a publisher-controlled winner-selection override for Rendering mode (`PrebidBidSelecting/bidSelector`, **Phase 5**, already prototyped in #1276). The selector ships independently of the multi-impression work and targets both iOS and Android.

Full spec for the bid-selector part (API shape, design rationale, alternatives considered, open questions): [The API proposal](https://docs.google.com/document/d/1CA6JjmiW0YKfM1qu3SRXjjPwESITzyYnNgnqWV2_52w/edit?usp=sharing), Part I.

PR #1276, opened 2026-07-07
Bid selector for winner override

mdanylov-sigma:
## Summary

- Adds `PrebidBidSelecting`, an `@objc` protocol letting publishers override the SDK's default marker-driven winning-bid selection, resolving part 2 of #950.
- `BidResponse.applyBidSelector(_:)` is called by `PBMBidRequester` right after transforming the server response, before cache filtering, and is re-consulted after `removeBidsWithoutSuccessfulCache()` so the selector's decision survives cache-based filtering.
- When a selector is active, it takes precedence over both the default `hb_pb`/`hb_bidder` marker logic and `Targeting.forceSdkToChooseWinner`. Returning `nil` is final — no bid's keywords are attached and `AdUnit.setUp` reports `.prebidDemandNoBids`, not success.
- Non-winning bids' standard winner markers (`hb_pb`/`hb_bidder`/`hb_cache_id`) are stripped from the merged `targetingInfo` when a selector is active, so the ad server is never targeted with a different bid than the one `winningBid` reports.
- Selector results are validated to be a member of the bids array the SDK provided; a stale/foreign `Bid` is logged and rejected rather than becoming the winner.

### DIFF
--- a/PrebidMobile/Swift/AdUnits/MultiformatAdUnit/PrebidRequest.swift
+++ b/PrebidMobile/Swift/AdUnits/MultiformatAdUnit/PrebidRequest.swift
@@ -28,7 +28,10 @@ public class PrebidRequest: NSObject {
     
     /// A flag that determines whether SKOverlay should be supported for interstitials
     public var supportSKOverlayForInterstitial: Bool = false
-    
+
+    /// A publisher-supplied strategy for choosing the winning bid. See `PrebidBidSelecting`.
+    public var bidSelector: PrebidBidSelecting?
+
     // MARK: - Internal properties
     
     private(set) var bannerParameters: BannerParameters?
--- a/PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCore/BidResponse.swift
+++ b/PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCore/BidResponse.swift
@@ -29,7 +29,10 @@ public class BidResponse: NSObject {
     public private(set) var tmaxrequest: NSNumber?
     
     public private(set) var ext: ORTBBidResponseExt?
-    
+
+    /// The publisher-supplied bid selector currently in effect, if any. Set via `applyBidSelector(_:)`.
+    public private(set) var bidSelector: PrebidBidSelecting?
+
     private(set) var rawResponse: RawBidResponse?
     
     public convenience init(adUnitId: String?, targetingInfo: [String: String]?) {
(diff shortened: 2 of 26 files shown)


PrebidMobile/Swift/AdUnits/MultiformatAdUnit/PrebidRequest.swift
```diff
@@ -28,7 +28,10 @@ public class PrebidRequest: NSObject {
     
     /// A flag that determines whether SKOverlay should be supported for interstitials
     public var supportSKOverlayForInterstitial: Bool = false
-    
+
+    /// A publisher-supplied strategy for choosing the winning bid. See `PrebidBidSelecting`.
+    public var bidSelector: PrebidBidSelecting?
```
OlenaPostindustria, line 33: Should this feature be enabled for original ad units as well?

PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCore/BidResponse.swift
```diff
@@ -103,21 +106,53 @@ public class BidResponse: NSObject {
...
+    /// - Parameter selector: The bid selector to apply, or `nil` to restore the default selection logic.
+    public func applyBidSelector(_ selector: PrebidBidSelecting?) {
+        bidSelector = selector
+        updateWinningBidAndTargetingInfo(from: allBids ?? [])
+    }
+
+    /// Standard winner-marker keys (see `Bid.isWinning`). These specifically identify "the" winner,
+    /// so they must never be attributed to a bid that isn't the (possibly selector-chosen) winner --
+    /// otherwise `targetingInfo` could still point at a different bid than `winningBid` does.
+    private static let winningBidMarkerKeys: Set<String> = ["hb_pb", "hb_bidder", "hb_cache_id"]
```
OlenaPostindustria, line 121: these markers are duplicated throughout the SDK, consider consolidating them into a single source

OlenaPostindustria (changes requested):

YuriyVelichkoPI:
@mdanylov-sigma  Is there a community approved spec or proposal describing the way SDK should provide bid selection logic?

mdanylov-sigma:
> @mdanylov-sigma  Is there a community approved spec or proposal describing the way SDK should provide bid selection logic? 

@YuriyVelichkoPI no, there wasn’t a community approved spec or proposal for this. The PrebidBidSelecting protocol / bidSelector design here is my own proposal for how to satisfy that ask - not something reviewed or signed off on beforehand. My fault.

If this doesn’t match what the maintainers/community have in mind, let's close the PR and working out the actual design together first.

YuriyVelichkoPI:
Let's make PRs as draft. 

Then - create a shared docoment with the implementation proposal (feature specification). Once it ready and approved for we'll return back to implementation according to that spec. Because for now, without approved requirements, it's hard to review the PR.

mdanylov-sigma:
I put together a full proposal: problem statement, API shape (shown as pseudocode), design rationale for each non-obvious decision, alternatives considered, and open questions for maintainers. Posting it for review before this PR goes further.

[Full doc](https://docs.google.com/document/d/1CA6JjmiW0YKfM1qu3SRXjjPwESITzyYnNgnqWV2_52w/edit?usp=sharing), Part I.

Issue #396, opened 2021-07-09
Migration to Swift


YuriyVelichkoOpenX:
## Motivation

The public classes of the Rendering Module were migrated from Objective-C to Swift. All others are on Objective-C still. 

Swift Package Manager does not support mixed frameworks. Therefore we can't provide Prebbid Mobile with rendering through SPM. 

We could split Objective-C and Swift parts to introduce different modules, but it controversies our modularization strategy (#395) in part dedicated to following the principles of layered architecture. So it is better to concentrate efforts on the complete migration.

## Given

In the following doc, you can find the list of classes for migration

https://docs.google.com/spreadsheets/d/1ZZoRkXsIa95RZELCoEsP4IHJ5TDyRsNAlfKSQU2YohI/edit?usp=sharing

## Objective

Feel free to take any Objective-C class, mark the row in the doc with yellow color, and migrate it to Swift. 

Try to follow the priority column. It will help to take the less coupled classes first. 

Mark migrated classes green after migration.

YuriyVelichkoPI:
The implementation of this epic will fix SPM integration

YuriyVelichkoPI:
Posting a phased migration plan drafted with Claude based on a fresh codebase audit. Reopening this issue — the migration is not complete: `PrebidMobile/Objc/` still contains ~119 `.m` and 152 `.h` files (vs 271 already-Swift files).

---

# Migration plan: PrebidMobile iOS SDK → Swift

## 1. Goal & scope

Migrate the entire ObjC layer in `PrebidMobile/Objc/` (~119 `.m` + 152 `.h` files) to Swift in independently mergeable steps. The Swift API layer in `PrebidMobile/Swift/` (271 files) is unchanged in shape — Swift twins land under the same path on the Swift side.

In scope: everything under `PrebidMobile/Objc/` including ORTB models, networking, AdView/MRAID/VAST rendering, OMSDK wrapper, utilities.
Out of scope: `EventHandlers/` adapters (already mostly Swift), tests don't need Swift rewrites unless a class change forces it.

## 2. Constraints

- **Min iOS deployment target stays at 13.0.** Any step that would require a bump must stop and ask first.
- **Drop ObjC consumer support.** The public umbrella `PrebidMobile.h` is empty — all `PBM*` ObjC types are public only to cross the SPM split-target seam (`__PrebidMobileInternal` ↔ `PrebidMobile`). No external app imports them. After migration, types collapse back to `internal`.
[...]

PR #1280, opened 2026-07-21
Swift Migration phases:  0, 1, 2

YuriyVelichkoPI:
# [swift-migration] Phases 0–2 — Setup, ORTB request models, Utilities & extensions

Squashed combination of `swift-migration-phase-0`, `swift-migration-phase-1`, and
`swift-migration-phase-2` into three commits on top of `master`, per the
[migration plan](https://github.com/prebid/prebid-mobile-ios/issues/396#issuecomment-4330667336)

Part of an ObjC → Swift
migration of `PrebidMobile/Objc/` covering ~119 `.m` + 152 `.h` files across 9 phases; this
PR lands the first three.

### Phase 0 — Setup & tooling (no migration)

- **S0.1 — Functional-gap audit & playbook** (`docs/migration/playbook.md`): codified 5 gaps
  before any Swift twins were written — dropping `NSCopying` (no live callsite), empty
  child-dict suppression in `JSONObject`, `isEqual`/`hash` for `PBMORTBFormat` (`NSSet`
  dedup), the `@objc NSObject` requirement for Phase 1–3 twins, and `init?` vs. the old
  broken-instance fallback.
- **S0.2 — Parity harness**: `ORTBParityHelper.swift` with `assertORTBParity<ObjCType, SwiftType>`
  plus baseline fixtures for Phase 1 leaf types.
- **S0.3 — Tooling hardening**: removed dead CircleCI provisioning from all four build/test
  scripts (repo runs exclusively on GitHub Actions `macos-15` now); hardened
  `testPrebidMobile.sh` (pre-deletes the test simulator before recreating it, scoped the
  clean-build step to an explicit scheme/destination); fixed a flaky
  `PrebidEventDelegateTests` crash (`assertForOverFulfill`); added the `/build-sdk` skill and
  updated `/xcodebuild`; started tracking `CLAUDE.md` (removed from `.gitignore`).

### Phase 1 — ORTB bid-request models (25 files)

Ports every ObjC `PBMORTB*` request-side model in
`PrebidMobile/Objc/PrebidMobileRendering/ORTB/` to Swift, in dependency order:

- **S1.1** — leaf models (`ORTBFormat`, `ORTBPublisher`, `ORTBGeo`, `ORTBDeal`,
  `ORTBSourceExtOMID`, `ORTBImpExtSkadn`, `ORTBDeviceExtAtts`)
- **S1.2** — composite request blocks (`ORTBBanner`, `ORTBVideo`, `ORTBPmp`,
  `ORTBImpExtPrebid`, `ORTBImp`)
- **S1.3** — top-level objects (`ORTBApp`, `ORTBAppExt(Prebid)`, `ORTBDevice`,
  `ORTBDeviceExtPrebid(Interstitial)`, `ORTBUser`, `ORTBRegs`, `ORTBSource`,
  `ORTBRendererConfig`)
- **S1.4** — root container (`ORTBBidRequest`, `ORTBBidRequestExtPrebid`); deleted
  `PBMORTBAbstract.m` (headers kept — still imported by Phase 3/4 ObjC parameter builders)

Naming convention established and applied from here on: Swift class drops the `PBM`
prefix (`ORTBFoo`), ObjC callers keep seeing the original name via `@objc(PBMORTBFoo)`.

10 gaps discovered and codified in the playbook (framework-build visibility requiring
`@objc public`, explicit ObjC selector bridge annotations, private-header invisibility to
Swift, empty-array preservation, `JSONObject.dict`'s `private(set)`, `NSCopying` via JSON
round-trip on `ORTBBidRequest`, `NSMutableDictionary` decode from `JSONSerialization`, and
the `PBMORTBAbstract` deletion cascade into test helpers).

### Phase 2 — Utilities & extensions (parallel-safe with Phase 1)

Ports `PrebidMobile/Objc/PrebidMobileRendering/Utilities/` and `ExtensionsAndWrappers/`:

- **S2.1** — standalone utilities: `Functions` (+ `Functions+Testing`), MRAID constant
  classes, `DeepLinkPlus`, `DeviceAccessManagerKeys`, `DownloadDataHelper`,
  `CircularProgressBar{Layer,View}`, `WKScriptMessageHandlerLeakAvoider`. Deferred:
  `PBMDeepLinkPlusHelper` (Phase 4) and `PBMWindowLocker` (Phase 8) — both depend on
  private-header types not yet ported. `PBMMRAIDConstants.m` partially remains (only the
  `NS_TYPED_ENUM` string globals, which have no Swift-bridgeable equivalent).
- **S2.2** — Foundation/UIKit category extensions: `NSDictionary`, `NSMutableDictionary`,
  `NSString`, `NSURL`, `NSException`, `UIView`, `UIWindow` `+PBMExtensions`,
  `TouchDownRecognizer`, and the view-exposure trio (`ViewExposureImpl`,
  `ViewExposureChecker`, `UIView+PBMViewExposure`).
- **S2.3** — `NSTimer` wrapper: `TimerInterface`, `NSTimer+PBMScheduledTimerFactory`,
  `WeakTimerTargetBox`. `NSInvocationOperation` replaced with `NSObject.perform(_:with:)`.

13 additional gaps codified (S2.1-A through S2.3-C) — `@_spi(PBMInternal)` propagation,
`dispatch_time()`/`UIInterfaceOrientationIsPortrait()` replacements, `@objc(name:error:)`
selector labels on throwing methods, `@dynamic` → `@NSManaged`, ObjC protocol
forward-declaration pattern, Foundation-nil string bridging, and header-reduction fallout.

## What changed vs. the original phase branches

The three phase branches (`swift-migration-phase-0/1/2`) were squashed into one commit per
phase and rebased onto current `master`. Two conflicts surfaced during the rebase, both
resolved as a union of intent (nothing from either side was dropped):

- **`.gitignore`** — kept master's new `EventHandlers/Package.resolved` entry; dropped the
  `CLAUDE.md`/`.claude/` ignore block per Phase 0's intent (this migration tracks those
  files).
- **`PrebidMobile.xcodeproj/project.pbxproj`** — kept Phase 1's two new `PBXBuildFile`
  entries (`ORTBBidRequestExtPrebid.swift`, `ORTBDeviceExtPrebidInterstitial.swift`); the
  conflict was purely adjacent-line positioning in the sorted list.

## Test plan

- [x] `./scripts/buildPrebidMobile.sh` — all 4 XCFrameworks build clean (device + simulator)
- [x] `./scripts/testPrebidMobile.sh --latest --quick` — 727 tests, 0 failures
- [x] `./scripts/testPrebidMobile.sh --latest` — full suite (1111 tests), run before merge
- [ ] Reviewer: confirm gap decisions in `docs/migration/playbook.md`


mdanylov-sigma (commented):
Migration playbook (`docs/migration/playbook.md`) is a genuinely useful artifact here - the gap-by-gap decision log made it possible to verify individual Swift ports against the deleted ObjC originals even for a PR this size. Worth carrying the same practice into later phases.

PrebidMobile/Swift/PrebidMobileRendering/Utilities/DownloadDataHelper.swift
```diff
@@ -0,0 +1,60 @@
...
+                completionClosure(nil, PBMError.error(description: "Unable to determine video file size: \(url?.absoluteString ?? "nil")"))
+                return
+            }
+
+            guard Int(contentLength) <= maxSize else {
+                completionClosure(nil, PBMError.error(description: "Cannot preRender video at \(url?.absoluteString ?? "nil"). Size of \(contentLength) bytes is greater than the maximum size for preloading of \(maxSize) bytes."))
+                return
+            }
+
+            self?.downloadData(for: url, completionClosure: completionClosure)
```
mdanylov-sigma: Completion closure may never fire. This callback captures `[weak self]` (line 40), but the ObjC original implicitly captured `self` *strongly* here. The only production caller (`PBMCreativeFactoryJob.m:192`) creates `PBMDownloadDataHelper` as a purely local variable with no other strong reference — in ObjC, the block's own strong self-capture was what kept it alive across the async HEAD→GET round trip. With `weak self`, once the creating method returns, ARC can deallocate the downloader before the HEAD response arrives; when it does, `self` is nil and `self?.downloadData(for:completionClosure:)` on line 57 silently no-ops — the GET never fires and `completionClosure` is never called. Worth confirming whether this explains any of the VAST/video preload timeouts in the suite. Suggest capturing `self` strongly here (or having the caller hold a strong reference for the duration of the download).

PrebidMobile/Swift/PrebidMobileRendering/Utilities/Functions.swift
```diff
@@ -75,8 +75,169 @@ import Foundation
...
+    @objc(dispatchTimeAfterTimeInterval:)
+    public static func dispatchTimeAfterTimeInterval(_ timeInterval: TimeInterval) -> UInt64 {
+        dispatchTimeAfterTimeInterval(timeInterval, startTime: DispatchTime.now().rawValue)
+    }
+
+    @objc(dispatchTimeAfterTimeInterval:startTime:)
+    public static func dispatchTimeAfterTimeInterval(_ timeInterval: TimeInterval,
+                                                     startTime: UInt64) -> UInt64 {
+        // dispatch_time_t is UInt64; use DispatchTime to avoid the deprecated C dispatch_time()
+        let base: DispatchTime = startTime == 0 ? .now() : DispatchTime(uptimeNanoseconds: startTime)
```
mdanylov-sigma: **Possible blocker — unit mismatch between `DispatchTime.rawValue` and `uptimeNanoseconds`.** The single-arg overload above (line 170) passes `DispatchTime.now().rawValue` as `startTime`. `DispatchTime.rawValue` is an opaque `dispatch_time_t` encoding, not a nanoseconds-since-boot count, so reconstructing it here via `DispatchTime(uptimeNanoseconds: startTime)` reinterprets one unit as the other and produces a deadline that isn't `now + timeInterval`. Both production call sites still route through the ObjC bridge (`[PBMFunctions dispatchTimeAfterTimeInterval:...]` in `PBMVideoView.m:501`, the video skip-delay countdown, and `PBMHTMLCreative.m:193`, the HTML display interval), so this looks live. Might be worth threading `DispatchTime`/`dispatch_time_t` through opaquely instead of round-tripping via `uptimeNanoseconds`.

PrebidMobile/Objc/PrebidMobileRendering/Prebid/PBMCore/PBMPrebidParameterBuilder.m
```diff
@@ -13,16 +13,14 @@
  limitations under the License.
  */
 
+#import "PBMConstants.h"
```
mdanylov-sigma, line 18: **Local SwiftPM build fails here.** `PBMFunctions.h` (removed above) was this file's only source of `#import <UIKit/UIKit.h>`, and line 109 still uses `[UIScreen mainScreen]`. Verified by checking out this branch and running `swift build --target __PrebidMobileInternal` for `arm64-apple-ios13.0-simulator`: `error: declaration of 'UIScreen' must be imported from module 'UIKit.UIScreen' before it is required`. The Xcode/CocoaPods framework build masks this (the generated `PrebidMobile-Swift.h`, pulled in via `SwiftImport.h`, happens to bring UIKit along there), but SPM's `__PrebidMobileInternal` target's `@import PrebidMobile;` doesn't re-export it. Adding `#import <UIKit/UIKit.h>` directly fixed it locally.

PrebidMobile/Objc/PrebidMobileRendering/PBMCreativeViewabilityTracker.m
```diff
@@ -13,12 +13,9 @@
  limitations under the License.
  */
 
-#import "PBMViewExposureChecker.h"
 #import "PBMMacros.h"
+#import "PBMScheduledTimerFactory.h"
```
mdanylov-sigma, line 19: Same SwiftPM UIKit-visibility issue as `PBMPrebidParameterBuilder.m` — `UIView+PBMExtensions.h` (removed) was this file's UIKit source, and `UIView` is still used directly below (e.g. `testedView`, `initWithView:`). Needs a direct `#import <UIKit/UIKit.h>` for the local SPM build to compile.

PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMAdViewManager.m
```diff
@@ -14,10 +14,8 @@
  */
 
 #import "PBMAdLoadManagerProtocol.h"
```
mdanylov-sigma, line 18: Same SwiftPM UIKit-visibility issue — `UIView+PBMExtensions.h` / `PBMFunctions+Private.h` (removed) were this file's UIKit source, and `UIView` / `UIInterfaceOrientationLandscapeLeft` (line 99/101) are still used directly. Needs a direct `#import <UIKit/UIKit.h>`.

PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCore/ORTB/Request/ORTBDeal.swift
```diff
@@ -0,0 +1,67 @@
...
+    public override init() {
+        super.init()
+    }
+
+    @objc(initWithJsonDictionary:)
+    public required init(jsonDictionary: [String: Any]) {
+        super.init()
+        let json = JSONObject<Key>(jsonDictionary)
+        id          = json[.id]
+        bidfloor    = json[.bidfloor] ?? 0.0
```
mdanylov-sigma: Decoding a JSON missing `bidfloor`/`bidfloorcur`/`wseat`/`wadomain` now restores the Swift class default (`0.0`, `"USD"`, `[]`, `[]`) instead of matching the ObjC original, which unconditionally overwrote these with whatever was in the dict — `nil` when the key was absent, and `nil` is never re-serialized on encode. Since `jsonDictionary` writes all four back out unguarded (lines 54–58), a deal decoded from a partial JSON now emits keys (e.g. `"bidfloorcur":"USD"`, `"wseat":[]`) that ObjC would have omitted on re-encode.

PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCore/ORTB/Request/ORTBImp.swift
```diff
@@ -0,0 +1,137 @@
...
+        super.init()
+        let json = JSONObject<Key>(jsonDictionary)
+        impID            = json[.id]
+        banner           = json[.banner]
+        video            = json[.video]
+        native           = json[.native]
+        pmp              = json[.pmp] ?? ORTBPmp()
+        displaymanager   = json[.displaymanager]
+        displaymanagerver = json[.displaymanagerver]
+        instl            = json[.instl] ?? 0
```
mdanylov-sigma: Same pattern as `ORTBDeal` — `instl`/`clickbrowser`/`secure` fall back to the class default (`0`/`1`/`0`) via `?? default` on decode, where ObjC's unconditional overwrite would leave them `nil` (and un-serialized) when the JSON key is missing. All three are written back unguarded in `jsonDictionary` (lines 86–88).

PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCore/ORTB/Request/ORTBBidRequest.swift
```diff
@@ -0,0 +1,109 @@
...
+        app       = json[.app]    ?? ORTBApp()
+        device    = json[.device] ?? ORTBDevice()
+        user      = json[.user]   ?? ORTBUser()
+        regs      = json[.regs]   ?? ORTBRegs()
+        source    = json[.source] ?? ORTBSource()
+        extPrebid = ORTBBidRequestExtPrebid(
+            jsonDictionary: (jsonDictionary["ext"] as? [String: Any])?["prebid"] as? [String: Any] ?? [:]
+        )
+
+        if let impDicts = jsonDictionary["imp"] as? [[String: Any]], !impDicts.isEmpty {
```
mdanylov-sigma: Inverse of the `ORTBDeal`/`ORTBImp` issue: `imp` defaults to `[ORTBImp()]` (line 24), but this only overwrites it `if !impDicts.isEmpty`. Decoding `{}` or `{"imp":[]}` leaves the one-element default in place, whereas ObjC's `initWithJsonDictionary:` unconditionally set `_imp` to whatever it built — an empty array in that case. Net effect: decoding an empty/missing `imp` and re-encoding produces a phantom impression that ObjC wouldn't have. Not currently exercised by a production decode path, but worth fixing before one exists.

PrebidMobileTests/ORTBParityHelper.swift
```diff
@@ -0,0 +1,100 @@
...
+// MARK: - Parity assertion
+
+/// Verifies that a Swift ORTB type produces expected JSON output for a given fixture.
+/// Used during Phase 1–3 of the ObjC → Swift migration to confirm JSON parity.
+///
+/// Usage:
+/// ```swift
+/// assertORTBParity(jsonString: ORTBFixtures.format, swiftType: ORTBFormat.self)
+/// ```
+func assertORTBParity<T: PBMJsonCodable>(
```
mdanylov-sigma, line 35: `assertORTBParity` only checks Swift-internal `encode(decode(json)) == encode(decode(decode(json)))` self-consistency against fully-populated `ORTBFixtures`, not a comparison against captured ObjC output (which is no longer possible once the ObjC types are deleted). It wouldn't have caught the `ORTBDeal`/`ORTBImp`/`ORTBBidRequest` decode-default issues above, since those only manifest on partial/incomplete JSON and the fixtures here are all fully populated. The name/doc-comment reads as a stronger guarantee than what's actually verified — might be worth softening the claim or adding partial-JSON fixtures.

PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCore/ORTB/Request/ORTBFormat.swift
```diff
@@ -0,0 +1,73 @@
...
+        json[.h]      = h
+        json[.wratio] = wratio
+        json[.hratio] = hratio
+        json[.wmin]   = wmin
+        return json.dict
+    }
+
+    // MARK: - NSObject equality (Gap 3 — used in NSSet deduplication)
+
+    public override func isEqual(_ object: Any?) -> Bool {
```
mdanylov-sigma, line 69: Minor semantics change: ObjC's `[self.w isEqual:other.w]` returns `NO` when `w` is nil (messaging nil), so two nil-`w` formats were never equal under the old `isEqual:`. Swift's `w == other.w` on `NSNumber?` treats `nil == nil` as `true`. Low risk since `w`/`h` are effectively always populated for real banner sizes, but this is the one type the playbook calls out as needing custom equality (for `NSSet` dedup in `PBMPrebidParameterBuilder.m`), so flagging in case it should be a documented decision.

PrebidMobileTests/RenderingTests/Tests/UtilitiesExtensionsTests/TestPBMFunctions.swift
```diff
@@ -279,11 +279,11 @@ class TestPBMFunctions: XCTestCase {
     
         var dict: JsonDictionary?
         do {
-            dict = try PBMFunctions.dictionaryFromData(data)
+            dict = try Functions.dictionaryFromData(data)
             XCTFail("Test method should throw exception")
         }
         catch {
-            XCTAssert(error.localizedDescription.contains("Could not convert json data to jsonObject:"))
+            XCTAssert(!error.localizedDescription.isEmpty)
```
mdanylov-sigma: This and the equivalent assertion at line 302 were weakened from checking specific error text (`"Could not convert json data to jsonObject:"` / `"Could not cast jsonObject to JsonDictionary:"`) to `!error.localizedDescription.isEmpty`, which now passes for virtually any thrown error. Understandable since `Functions.swift`'s `dictionaryFromData` throws different wording than the deleted ObjC method, but it'd be worth tightening this to match whatever it actually throws now rather than dropping the specificity entirely.

PrebidMobileTests/PrebidEventDelegateTests.swift
```diff
@@ -34,6 +34,8 @@ class PrebidEventDelegateTests: XCTestCase {
     
     func test_eventDelegate_isCalled() {
         let exp = expectation(description: "Expect PrebidEventDelegate to be called")
+        // PBMBidRequester also fires this delegate; other tests' in-flight responses can land here.
+        exp.assertForOverFulfill = false
```
mdanylov-sigma: The comment here ("other tests' in-flight responses can land here") describes a real test-isolation bug — a shared/global delegate receiving a response from a different test. `assertForOverFulfill = false` silences the symptom (the crash) rather than fixing the underlying isolation issue. Fine as a stopgap, but might be worth a tracked follow-up rather than leaving it as the permanent fix.

.gitignore
```diff
@@ -116,8 +116,4 @@ Pods/
 # Visual Studio Code
 .build/
 
-# Claude Code
-CLAUDE.md
-.claude/
-
 EventHandlers/Package.resolved
```
mdanylov-sigma, line 127: This removes `CLAUDE.md` / `.claude/` from `.gitignore` entirely rather than scoping it to what's actually meant to be tracked (`CLAUDE.md`, `agents/`). As written, anything under a contributor's local `.claude/` - `settings.local.json`, `worktrees/`, etc. - is no longer ignored and could get committed accidentally. Consider scoping instead, e.g.:
```
.claude/*
!.claude/skills/
!.claude/skills/**
.claude/settings.local.json
.claude/worktrees/
```

scripts/testPrebidMobile.sh
```diff
@@ -40,18 +40,22 @@ echo -e "\n\n${GREEN}INSTALL PODS${NC}\n\n"
 
 cd ..
 
-export PATH="/Users/distiller/.gem/ruby/2.7.0/bin:$PATH"
-gem install cocoapods
 pod install --repo-update
```
mdanylov-sigma, line 49: `gem install cocoapods` was removed here (and in `buildPrebidMobile.sh`, `testPrebidDemo.sh`, `testPrebidMobileAdapters.sh`) with nothing added in its place. Reasonable if the `macos-15` GitHub Actions image ships CocoaPods already, but a `command -v pod || { echo "CocoaPods not found"; exit 1; }` guard would fail loudly instead of with a confusing `pod: command not found` mid-script if that assumption ever breaks.

scripts/buildPrebidMobile.sh
```diff
@@ -41,8 +41,6 @@ touch "$LOG_FILE_FRAMEWORK"
 
 echo -e "\n\n${GREEN}INSTALL PODS${NC}\n\n"
 
-export PATH="/Users/distiller/.gem/ruby/2.7.0/bin:$PATH"
-gem install cocoapods
 pod install --repo-update
```
mdanylov-sigma, line 50: Same as `testPrebidMobile.sh` — `gem install cocoapods` removed with no replacement guard.

agents/review/SKILL.md
```diff
@@ -0,0 +1,172 @@
...
+Produce a structured review covering correctness, architecture, tests, and
+repo-specific concerns. Adapt depth to the PR type — migration PRs get extra
+migration-checklist items; feature/fix PRs skip migration sections.
+
+## Step 1 — Orient
+
+```bash
+git log --oneline master..HEAD          # commits in this PR
+git diff --stat master..HEAD            # files changed
+git diff master..HEAD                   # full diff (read selectively)
```
mdanylov-sigma: `git diff master..HEAD` / `git log --oneline master..HEAD` use a two-dot range against a local (not `origin/master`) ref. For reviewing a PR branch this is fragile: two-dot compares tip-to-tip directly rather than against the merge-base, so if `master` has moved on, the diff can include master's changes as noise; and a stale local `master` (not fetched) silently produces a wrong diff. `git diff origin/master...HEAD` (three-dot) would be more robust here.

docs/migration/playbook.md
```diff
@@ -0,0 +1,477 @@
...
+
+Files to scan after each migration step:
+- `PrebidMobileTests/RenderingTests/Tests/PBMORTBAbstractTest.swift`
+- `PrebidMobileTests/RenderingTests/Tests/PBMORTBBidRequestTest.swift`
+- `PrebidMobileTests/RenderingTests/Tests/ParameterBuilderTests/PrebidParameterBuilderTest.swift`
+- Any other Swift test file that imports or instantiates the migrated types
+
+Run after renaming: `xcodebuild ... build-for-testing 2>&1 | grep "error:" | grep "has been renamed"` to catch stragglers.
+
+## Known flaky test — do not over-investigate
```
mdanylov-sigma: This "do not over-investigate" policy for `PBMBidRequesterTest.testBanner_300x250` is reasonable as a note about one specific test in this migration, but as a standing instruction (including to any AI agent following this playbook) it reads as blanket permission to dismiss future failures in that test without investigation. Might be worth scoping it more tightly to this PR/phase, or replacing with a link to a tracked flakiness issue.

docs/migration/pr-phase-0.md
```diff
@@ -0,0 +1,27 @@
+## Summary
```
mdanylov-sigma: This file (and `pr-phase-1.md`, `pr-phase-1-review.md`, `pr-phase-2.md`, `pr-phase-2-description.md`, `pr-phase-012.md`) look like drafts used to write the GitHub PR descriptions for the individual phase branches / now-closed PRs. They duplicate content already on GitHub and will go stale as soon as those PRs' descriptions are edited. Consider dropping these and keeping only `playbook.md`, which is the living reference doc.

PrebidMobile/Swift/PrebidMobileRendering/Utilities/Functions.swift
```diff
@@ -75,8 +75,216 @@ import Foundation
...
+
+    @objc public static var deviceMaxSize: CGSize {
+        let screenSize = deviceScreenSize
+        let saInsets   = safeAreaInsets
+        return CGSize(width:  screenSize.width  - saInsets.left - saInsets.right,
+                      height: screenSize.height - statusBarHeight - saInsets.top - saInsets.bottom)
+    }
+
+    @objc public static var safeAreaInsets: UIEdgeInsets {
+        UIApplication.shared.keyWindow?.safeAreaInsets ?? .zero
```
mdanylov-sigma: **Possible correctness issue.** `safeAreaInsets` still reads `UIApplication.shared.keyWindow` directly — the exact crash class this diff just fixed two lines below in `statusBarHeight` (via the new `sharedApplication` accessor at line 143). Per that accessor's own doc comment, `UIApplication.shared` traps when the SDK runs host-less (e.g. the unit-test bundle). `deviceMaxSize` (line 269) calls this `safeAreaInsets` right next to the now-safe `statusBarHeight` (line 271), and is reachable from production ObjC via `PBMMRAIDController.m`/`PBMWebView.m`. Looks like `safeAreaInsets` was simply missed when the fix was applied to its siblings.

PrebidMobile/Swift/PrebidMobileRendering/Utilities/Functions.swift
```diff
@@ -75,8 +75,216 @@ import Foundation
...
+        case dispatchTimeNow:
+            return (DispatchTime.now() + timeInterval).rawValue
+        case dispatchTimeForever:
+            return dispatchTimeForever
+        default:
+            return startTime &+ UInt64(bitPattern: machTicks(fromSeconds: timeInterval))
+        }
+    }
+
+    private static func machTicks(fromSeconds seconds: TimeInterval) -> Int64 {
```
mdanylov-sigma: **Possible correctness issue.** `machTicks(fromSeconds:)` has three edge-case problems, all currently latent since nothing reaches this code path (see the companion comment on the `default:` case below):
1. `Int64(seconds * TimeInterval(NSEC_PER_SEC))` traps for `.nan`/`.infinity`/very large `seconds` — the `.now()` branch instead uses `DispatchTime`'s saturating `+`, so the two branches have inconsistent robustness for the same bad input.
2. On a real device (`mach_timebase_info` = 125/3), `nanoseconds * Int64(timebase.denom)` can itself overflow-trap before the division, for `seconds` ≳ 97 years.
3. Back at the call site (`return startTime &+ UInt64(bitPattern: machTicks(...))`), a **negative** `timeInterval` with a small `startTime` produces a negative `Int64`, whose `UInt64(bitPattern:)` is near `UInt64.max`; adding that via `&+` yields a deadline near `UInt64.max` ("almost forever") instead of the intended "already past." Traced through concrete numbers: `startTime=1_000_000`, `timeInterval=-1.0` → `machTicks = -24_000_000` → result `≈ UInt64.max - 23_000_000`.

Separate, smaller point: `mach_timebase_info(&timebase)` is re-fetched on every call here, but the numer/denom ratio is a hardware constant fixed for the process's lifetime — could be hoisted into a `private static let` computed once instead of re-resolved per call.

PrebidMobile/Swift/PrebidMobileRendering/Utilities/Functions.swift
```diff
@@ -75,8 +75,216 @@ import Foundation
...
+    }
+
+    @objc(toStringJsonDictionary:error:)
+    public static func toStringJsonDictionary(_ jsonDictionary: [String: Any]) throws -> String {
+        try jsonString(from: jsonDictionary)
+    }
+
+    // MARK: - ObjC bridge (PBMFunctions+Private — UI)
+
+    @objc public static var statusBarHeight: CGFloat {
```
mdanylov-sigma, line 296: `statusBarHeight` only checks the new `sharedApplication` accessor, unlike `attemptToOpen` (line 156) which checks `Functions.application ?? sharedApplication`. The `Functions.application` test-injection seam is bypassed here — a test that sets `Functions.application = mockApplication` and reads this static var would silently resolve against the real `UIApplication.shared`/`sharedApplication` instead of the mock. `TestPBMFunctions.swift` only exercises the `statusBarHeight(application:)` overload directly, so this inconsistency isn't caught today.

agents/xcodebuild/SKILL.md
```diff
@@ -0,0 +1,267 @@
...
+| `use of undeclared identifier 'PBMJsonDictionary'` | Lost typedef from deleted PBMFunctions.h | Add `#import "PBMConstants.h"` |
+| `'PBMFoo' has been renamed to 'Foo'` | Swift test using ObjC bridging name | `perl -pi -e 's/PBMFoo\b/Foo/g'` in test files |
+| `inaccessible due to '@_spi'` | Swift test accessing Functions.* without SPI import | Add `@_spi(PBMInternal) @testable import` |
+| `cannot call value of non-function type 'String'` | ObjC method → Swift property; called with `()` | Remove `()` — it's now a property |
+| `missing arguments for parameters 'width', 'height'` | Swift init lacks no-arg form | Add `public override init() { super.init() }` with defaulted properties |
+| `cannot declare conformance to 'NSObjectProtocol'` | Mock class missing NSObject | Add `: NSObject` superclass |
+| `@objc method name provides N argument names, but method has N+1` | `throws` method needs `:error:` in explicit name | Use `@objc(name:error:)` not `@objc(name:)` |
+| `'dispatch_time' has been replaced` | C function unavailable in Swift | `(DispatchTime(uptimeNanoseconds: t) + interval).rawValue` |
```
mdanylov-sigma: This troubleshooting row still maps the `'dispatch_time' has been replaced` compiler error to `(DispatchTime(uptimeNanoseconds: t) + interval).rawValue` — the exact pattern `playbook.md` Gap S2.1-C and `Functions.swift` were just corrected to stop using (explicitly labeled "WRONG — do not use" in the playbook). A future contributor hitting that compiler error will consult this skill doc first and get led straight back into the ~41x-off-on-arm64 bug this PR fixed elsewhere. Worth updating to point at the new `Functions.dispatchTimeAfterTimeInterval` pattern instead.

PrebidMobile/Swift/PrebidMobileRendering/Utilities/Functions.swift
```diff
@@ -75,8 +75,216 @@ import Foundation
...
+    /// (i.e. on arm64 devices; the simulator's 1:1 timebase hides the error).
+    @objc(dispatchTimeAfterTimeInterval:startTime:)
+    public static func dispatchTimeAfterTimeInterval(_ timeInterval: TimeInterval,
+                                                     startTime: UInt64) -> UInt64 {
+        switch startTime {
+        case dispatchTimeNow:
+            return (DispatchTime.now() + timeInterval).rawValue
+        case dispatchTimeForever:
+            return dispatchTimeForever
+        default:
```
mdanylov-sigma, line 227: This `default:` branch — the actual mach-tick conversion this PR added to fix the original bug — has zero test coverage anywhere in the repo. Both real production callers (`PBMHTMLCreative.m:188`, `PBMVideoView.m:499`) only use the 1-arg overload, which always passes `startTime = dispatchTimeNow` and hits the `.now()` case above. The only 2-arg test (`PBMFunctionsPrivateTest.m`) also only passes `DISPATCH_TIME_NOW`. So this branch — and the edge cases flagged in the `machTicks` comment below — ship unverified on simulator or device.

PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCore/ORTB/Request/ORTBBidRequest.swift
```diff
@@ -0,0 +1,111 @@
...
+        regs      = json[.regs]   ?? ORTBRegs()
+        source    = json[.source] ?? ORTBSource()
+        extPrebid = ORTBBidRequestExtPrebid(
+            jsonDictionary: (jsonDictionary["ext"] as? [String: Any])?["prebid"] as? [String: Any] ?? [:]
+        )
+
+        // Assigned unconditionally, matching `initWithJsonDictionary:` — decoding a request
+        // with a missing or empty "imp" must clear the one-element default, not preserve it,
+        // or re-encoding invents a phantom impression. See playbook Gap S2.5-C.
+        let impDicts = jsonDictionary["imp"] as? [Any] ?? []
```
mdanylov-sigma, line 64: Minor: this `compactMap` silently drops non-dictionary elements from a mixed-type `"imp"` array instead of failing/falling back to the default, which diverges a bit from the surrounding comment's claim of matching ObjC's unconditional-assignment semantics exactly (ObjC's cast to `[[String: Any]]` would have failed the whole array on one bad element). Arguably safer as written, but worth a one-line note if intentional — and the new `partialBidRequest`/`emptyImpBidRequest` fixtures only cover absent/empty `imp`, not malformed elements, so there's no test pinning this behavior either way.

PrebidMobile/Swift/PrebidMobileRendering/Utilities/Functions.swift
```diff
@@ -75,8 +75,216 @@ import Foundation
...
+    // MARK: - Shared application
+
+    /// `[UIApplication sharedApplication]` is `nil` whenever the SDK runs outside an
+    /// application process — e.g. in the unit-test bundle, which has no app host.
+    /// Swift imports `UIApplication.shared` as non-optional, so that `nil` becomes an
+    /// invalid reference that segfaults as soon as it is used (notably when boxed into
+    /// a `PBMUIApplicationProtocol` existential). Fetch it through the ObjC runtime
+    /// instead, so the `nil` stays observable — this is the Swift equivalent of the
+    /// `if (!uiApplication)` / `conformsToProtocol:` guards the ObjC original had.
+    static var sharedApplication: PBMUIApplicationProtocol? {
```
mdanylov-sigma: This fix is applied inconsistently: within this same file it covers `attemptToOpen` and `statusBarHeight` but not `safeAreaInsets` (see comment above), and at least 8 other Swift files elsewhere in the SDK (`ExtensionsAndWrappers/UIApplication+Extensions.swift`, `UIWindow+PBMExtensions.swift`, `AutoRefreshManager.swift`, `ViewExposureChecker.swift`, `LocationManager.swift`, etc.) still read `UIApplication.shared` directly and unguarded. The crash class this PR set out to fix is only closed at 2 call sites out of many that share the same risk — might be worth a playbook gap entry (there isn't one for this pattern yet) so the fix is discoverable/applied consistently going forward.

Separate, smaller point: this is a computed `static var` that redoes the `NSSelectorFromString`/`responds(to:)`/`perform(_:)` runtime dispatch on every access, even though the resolved `UIApplication` instance never changes during a process's lifetime — could be cached once after first successful resolution. It's on a rendering-adjacent path (`attemptToOpen` fires on every click-through, `statusBarHeight` on every viewability check).

scripts/testPrebidDemo.sh
```diff
@@ -13,8 +13,11 @@ xcrun simctl create iPhone-16-Pro-PrebidMobile com.apple.CoreSimulator.SimDevice
 cd ..
 echo $PWD
 
-export PATH="/Users/distiller/.gem/ruby/2.7.0/bin:$PATH"
-gem install cocoapods
+if ! command -v pod >/dev/null 2>&1; then
```
mdanylov-sigma, line 19: This is the only one of the four scripts touched by this fix that doesn't have `set -e`/`set -euo pipefail` anywhere in the file (`buildPrebidMobile.sh`, `testPrebidMobile.sh`, `testPrebidMobileAdapters.sh` all do). `pod deintegrate`/`pod install --repo-update`/`pod update` below can fail for reasons other than a missing `pod` binary (network, dependency resolution, stale lockfile); without `set -e` the script continues into `xcodebuild` against a broken/partial Pods checkout instead of failing fast with an actionable message.

scripts/buildPrebidMobile.sh
```diff
@@ -41,8 +41,12 @@ touch "$LOG_FILE_FRAMEWORK"
 
 echo -e "\n\n${GREEN}INSTALL PODS${NC}\n\n"
 
-export PATH="/Users/distiller/.gem/ruby/2.7.0/bin:$PATH"
-gem install cocoapods
+if ! command -v pod >/dev/null 2>&1; then
```
mdanylov-sigma, line 44: This 5-line `command -v pod` guard is now duplicated near-verbatim across all four scripts (`buildPrebidMobile.sh`, `testPrebidDemo.sh`, `testPrebidMobile.sh`, `testPrebidMobileAdapters.sh`), with styling already drifting between them (only this one uses `${RED}`/`${NC}`). Might be worth factoring into one function in a shared script the four source, so future changes don't need four edits.

docs/migration/playbook.md
```diff
@@ -0,0 +1,629 @@
...
+(`PBMPrebidParameterBuilder`) builds every element via `+ortbFormatWithSize:`, which always sets `w`
+and `h`, so all-nil formats never reach the `NSSet`. Reproducing ObjC exactly would need
+`w != nil && h != nil && …`, which makes `isEqual:` non-reflexive and violates the contract
+`NSSet`/`Hashable` rely on.
+
+**Rule:** When porting an `isEqual:` built from `isEqual:` calls on optional properties, decide
+explicitly whether all-nil instances must stay distinct. If the ObjC semantics can't be reproduced
+without breaking reflexivity, keep the Swift semantics and leave a comment at the callsite
+justifying it — don't leave the difference silent.
+
```
mdanylov-sigma, line 575: Per this file's own convention (and AGENTS.md's rule to update the playbook "when a new gap is discovered"), the other three fixes in this round each got a lettered gap entry (S2.5-A/B/D here, S2.5-C above). The new `sharedApplication` workaround for host-less `UIApplication.shared` (`Functions.swift`) is the same kind of newly-discovered gap but has no corresponding entry — `playbook.md` doesn't mention `sharedApplication`/`UIApplication` anywhere. Might be worth a Gap S2.5-E here so the pattern is discoverable for the other ~8 files still reading `UIApplication.shared` directly.

.gitignore
```diff
@@ -116,8 +116,12 @@ Pods/
 # Visual Studio Code
 .build/
 
-# Claude Code
-CLAUDE.md
-.claude/
+# Claude Code — shared agent config (AGENTS.md, CLAUDE.md, agents/) is tracked on purpose;
+# only per-developer machine state is ignored.
+.claude/settings.local.json
```
mdanylov-sigma, line 121: This scoped `.claude/` list (`settings.local.json`, `shell-snapshots/`, `todos/`, `statsig/`, `projects/`) is introduced with the comment "only per-developer machine state is ignored," but `.claude/worktrees/` — created whenever an agent runs with worktree isolation, containing a full nested git checkout — isn't on it and is exactly that kind of state. Confirmed it exists on at least one dev machine and wouldn't be caught by any of these five entries.

docs/migration/pr-phase-012.md
```diff
@@ -0,0 +1,179 @@
...
+- Narrowed the playbook's flaky-test section to the single allowlisted test with an explicit
+  three-point confirmation, replacing the general "re-run once and move on" framing.
+- Collapsed `docs/migration/` to one PR doc: deleted the five superseded drafts
+  (`pr-phase-0/1/1-review/2/2-description.md`), which described branches that no longer exist.
+
+## Test plan
+
+- [x] `./scripts/buildPrebidMobile.sh` — all 4 XCFrameworks build clean (device + simulator)
+- [x] `./scripts/buildPrebidMobilePackage.sh` — SwiftPM build of the working tree clean
+- [x] `./scripts/testPrebidMobile.sh --latest --quick` — 754 tests, 0 failures (727 + the three
```
mdanylov-sigma: Arithmetic doesn't match: "754 tests, 0 failures (727 + the three new `ORTBAbstractTest` methods...)" — but 727 + 3 = 730, not 754. Probably just a typo in one of the three numbers.

PrebidMobileTests/PrebidEventDelegateTests.swift
```diff
@@ -20,31 +20,40 @@ private typealias Callback = (Data?, Data?) -> Void
...
     
-    let mockRequestData = "req".data(using: .utf8)
-    let mockResponseData = "res".data(using: .utf8)
-    
+    // `Prebid.shared.eventDelegate` is process-global, and `PBMBidRequester` fires it too, so
+    // an in-flight response from another test can reach this delegate. Tagging the payloads
+    // per test instance lets the callback ignore foreign invocations instead of relaxing the
+    // expectation with `assertForOverFulfill = false`, which would also mask a genuine
+    // double-call from the code under test.
+    private let tag = UUID().uuidString
```
mdanylov-sigma, line 28: Minor simplification thought: this UUID-tag mechanism (a `tag`, two `lazy var` payloads, two local snapshot lets, and a content-based filter) is five new pieces of state to solve "ignore another test's in-flight callback on the shared `Prebid.shared.eventDelegate`." Since XCTest allocates a fresh `PrebidEventDelegateTests`/`delegate` instance per test method, filtering by delegate identity (bail unless the currently-invoked delegate is still the one this test registered) would get the same isolation without the UUID/tagging convention every future test in this file needs to remember to repeat. Not a bug — just seems like more machinery than needed.

mdanylov-sigma (commented):
Re-review after the fix commits (`4a630a42`, `6895edd0`, `b17d27e7`). The fixes for the earlier findings are thorough and well-documented — good work. This pass (xhigh-effort, 7 independent finder angles + direct verification) surfaced 15 new items, mostly concentrated in `Functions.swift`'s `sharedApplication`/`machTicks` additions. Draft — not yet submitted.

PrebidMobile/Swift/PrebidMobileRendering/Utilities/Functions.swift
YuriyVelichkoPI: Fixed in the pending commit.

`safeAreaInsets` no longer touches `UIApplication.shared`. Both it and the `sharedApplication` accessor now go through a single private `resolvedUIApplication: UIApplication?` that does the ObjC runtime lookup, so there is one place left in the file that can produce an application instance:

```swift
@objc public static var safeAreaInsets: UIEdgeInsets {
    resolvedUIApplication?.keyWindow?.safeAreaInsets ?? .zero
}
```

`sharedApplication` keeps its `PBMUIApplicationProtocol?` type for its existing callers; it just returns `resolvedUIApplication`. It could not be reused directly here because `PBMUIApplicationProtocol` does not declare `keyWindow`.

New test: `TestFunctions.testSafeAreaInsetsWithoutApplicationHost` — asserts `.zero` rather than a crash when running host-less.

PrebidMobile/Swift/PrebidMobileRendering/Utilities/Functions.swift
YuriyVelichkoPI: Fixed in the pending commit — all three edge cases, plus the timebase re-fetch.

```swift
private static let machTimebase: mach_timebase_info_data_t? = {
    var timebase = mach_timebase_info_data_t()
    guard mach_timebase_info(&timebase) == KERN_SUCCESS, timebase.numer != 0 else { return nil }
    return timebase
}()

private static let maxRepresentableSeconds = TimeInterval(Int64.max / Int64(NSEC_PER_SEC))

private static func representableSeconds(_ seconds: TimeInterval) -> TimeInterval {
    guard !seconds.isNaN else { return 0 }
    return clamp(seconds, lowerBound: -maxRepresentableSeconds, upperBound: maxRepresentableSeconds)
}

private static func machTicks(fromSeconds seconds: TimeInterval) -> Int64 {
    let nanoseconds = Int64(representableSeconds(seconds) * TimeInterval(NSEC_PER_SEC))
    guard let timebase = machTimebase else { return nanoseconds }
    let (ticks, overflow) = nanoseconds.multipliedReportingOverflow(by: Int64(timebase.denom))
    guard !overflow else { return nanoseconds < 0 ? .min : .max }
    return ticks / Int64(timebase.numer)
}
```

- **NaN / out-of-range → trap.** `representableSeconds` maps NaN to `0` and clamps the magnitude, so the `Int64(...)` conversion is in range by construction. It is also applied on the `DispatchTime.now()` path, where `DispatchTime`'s own `+` had the same precondition.
- **Product overflow.** `multipliedReportingOverflow` saturates to `.min`/`.max` instead of trapping.
- **Negative → near `UInt64.max`.** The `default:` branch is now sign-branched and clamps at `DISPATCH_TIME_NOW` on underflow instead of wrapping:

```swift
let ticks = machTicks(fromSeconds: timeInterval)
if ticks >= 0 {
    let (deadline, overflow) = startTime.addingReportingOverflow(UInt64(ticks))
    return overflow ? dispatchTimeForever : deadline
}
let elapsed = UInt64(ticks.magnitude)
return elapsed > startTime ? dispatchTimeNow : startTime - elapsed
```
- **Timebase re-fetch.** `numer`/`denom` is a hardware constant for the process lifetime, so it is now a `private static let` resolved once.

One note on the clamp bound: it is derived by integer division (`Int64.max / Int64(NSEC_PER_SEC)`) rather than `TimeInterval(Int64.max) / TimeInterval(NSEC_PER_SEC)`, because the latter rounds up to 2^63 and would let `seconds * NSEC_PER_SEC` land exactly on the trapping boundary.

PrebidMobile/Swift/PrebidMobileRendering/Utilities/Functions.swift
YuriyVelichkoPI: Fixed in the pending commit — `statusBarHeight` now matches `attemptToOpen`:

```swift
@objc public static var statusBarHeight: CGFloat {
    guard let application = Functions.application ?? sharedApplication else { return 0 }
    return statusBarHeight(application: application)
}
```

New test `TestFunctions.testStatusBarHeightUsesInjectedApplication` pins the seam: it sets `Functions.application` to a `MockUIApplication` with a 3x4 status bar frame, asserts `4.0` through the static var (not the overload), then flips `isStatusBarHidden` and asserts `0.0`.

agents/xcodebuild/SKILL.md
YuriyVelichkoPI: Fixed in the pending commit. The row now reads:

| `'dispatch_time' has been replaced` | C function unavailable in Swift | Call `Functions.dispatchTimeAfterTimeInterval(_:startTime:)`. **Never** `DispatchTime(uptimeNanoseconds:)` on a `dispatch_time_t` — off by ~41x on arm64 (playbook Gap S2.1-C) |

Good catch on the direction of the reference — the skill doc was pointing at the pattern the playbook labels "WRONG".

While there I also found the playbook's own Gap S2.1-C reference implementation had gone stale (it still showed the `&+ UInt64(bitPattern:)` form), so that block is refreshed to the current sign-branched code as well.

PrebidMobile/Swift/PrebidMobileRendering/Utilities/Functions.swift
YuriyVelichkoPI: Fixed in the pending commit — 6 new tests in `TestFunctions`, all exercising the 2-arg overload with an explicit non-`DISPATCH_TIME_NOW` start time:

- `testDispatchTimeAfterTimeIntervalWithExplicitStartTime`
- `...WithNegativeInterval`
- `...UnderflowClampsToNow`
- `...WithForeverStartTime`
- `...WithNonFiniteIntervals`
- `...FromNow`

Worth recording why this had zero coverage: `PBMFunctionsPrivateTest` is listed in `PrebidMobilePRTests.xctestplan`'s `skippedTests`, so it does not run on PRs at all. The new tests went into `TestFunctions`, which is not skipped.

The positive-delay assertion converts back out of ticks rather than asserting an implementation-derived number, so it stays honest on both a 1:1 (simulator) and a 125/3 (arm64 device) timebase:

```swift
let ticks = deadline - startTime
let nanoseconds = ticks * UInt64(timebase.numer) / UInt64(timebase.denom)
XCTAssertEqual(Double(nanoseconds), Double(NSEC_PER_SEC), accuracy: Double(NSEC_PER_SEC) / 1000.0)
```

PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCore/ORTB/Request/ORTBBidRequest.swift
YuriyVelichkoPI: Fixed in the pending commit — the divergence is intentional, so it is now stated as such and pinned by a test.

```swift
// Deliberate divergence from ObjC: the cast to `NSArray<NSDictionary *> *` was
// unchecked, so one non-dictionary element corrupted every impression in the
// request. Dropping only the malformed elements keeps the valid ones usable.
// Pinned by `testDecodingRequestWithMalformedImpElements`.
```

New fixture `ORTBFixtures.malformedImpBidRequest`:

```json
{"id":"req-1","imp":[{"id":"imp-1"},"not-an-object",7,null,{"id":"imp-2"}]}
```

and `ORTBAbstractTest.testDecodingRequestWithMalformedImpElements` asserts `imp.count == 2` and `imp.compactMap { $0.impID } == ["imp-1", "imp-2"]`.

PrebidMobile/Swift/PrebidMobileRendering/Utilities/Functions.swift
YuriyVelichkoPI: Split answer — one part fixed, one tracked, one declined.

**Fixed:** `safeAreaInsets` (see the reply on that thread) and `statusBarHeight`'s missed `Functions.application` seam. Within `Functions.swift` there are now no direct `UIApplication.shared` reads left.

**Playbook entry added:** new **Gap S2.5-E — `UIApplication.shared` is non-optional in Swift but nil host-less** in `docs/migration/playbook.md`, with the runtime-lookup snippet and two rules (honour the `Functions.application` seam at call sites; do not memoize the result).

**Tracked, out of scope for this PR:** converting the remaining call sites. I measured the scope rather than carrying the estimate forward — it is **10 files across 15 call sites**, not ~8:

```
rg -c "UIApplication\.shared" --glob '*.swift' PrebidMobile/
```

Those are behavioural changes to files this PR does not otherwise touch, so they belong in the phase that ports them. Recorded in `docs/migration/pr-phase-012.md` under "S2.7 — Deferred, with rationale", with the re-measure command and the per-file counts in the gap entry.

**Declined:** memoizing `sharedApplication`. `mach_timebase_info` is now cached because it is an immutable hardware constant, but the application instance is not the same case: `sharedApplication` is `nil` until `UIApplicationMain` has run, so a cached `nil` from an early call would outlive the condition that produced it, and unlike the timebase it would need mutable global state (with the attendant race). The saving is a runtime `objc_msgSend` on paths that are already doing UIKit work. The accessor carries that reasoning as a doc comment so it does not get "optimized" later:

```swift
/// Deliberately not memoized: `sharedApplication` is `nil` until `UIApplicationMain`
/// has run, so a cached `nil` could outlive the condition that produced it.
```

scripts/testPrebidDemo.sh
YuriyVelichkoPI: Fixed in the pending commit.

```bash
# Set bash script to exit immediately if any commands fail.
set -e
```

Plain `set -e` rather than `set -euo pipefail` for two reasons: it matches the three sibling scripts verbatim, and `set -u` would break this script immediately — line 26 dereferences `"$1"` unguarded, so an argument-less invocation would start failing.

scripts/buildPrebidMobile.sh
YuriyVelichkoPI: Agreed, but out of scope for this PR — tracked.

The duplication is real (four near-verbatim copies, with `${RED}`/`${NC}` styling only in `buildPrebidMobile.sh`). Factoring it out means introducing a new sourced script and changing the entry-point contract of all four, which is a build-infrastructure change unrelated to the Swift migration this PR is doing, and it would be the riskiest thing in the diff for CI.

Recorded in `docs/migration/pr-phase-012.md` under "S2.7 — Deferred, with rationale" so it is not lost.

docs/migration/playbook.md
YuriyVelichkoPI: Fixed in the pending commit — **Gap S2.5-E — `UIApplication.shared` is non-optional in Swift but nil host-less** added to `docs/migration/playbook.md`.

It covers the failure mode (imported non-optional, but `nil` before `UIApplicationMain`, and boxing it into a `PBMUIApplicationProtocol` existential segfaults), the runtime-lookup workaround, and two rules: read `Functions.application ?? sharedApplication` at call sites so the test seam keeps priority, and do not memoize.

The scope note in the entry uses a measured figure rather than the estimate — **10 files across 15 call sites** — together with the re-measure command and the per-file counts, so the entry can be checked off objectively later.

.gitignore
YuriyVelichkoPI: Fixed in the pending commit — `.claude/worktrees/` added to the list.

docs/migration/pr-phase-012.md
YuriyVelichkoPI: Fixed in the pending commit. The `727` was the stale figure; rather than reconstruct the arithmetic I re-ran the subset and used the measured number.

`./scripts/testPrebidMobile.sh --latest --quick` → **763 tests, 0 failures** on the current branch, which includes the 9 tests added in this review round (2 for the `sharedApplication`/seam fixes, 6 for `dispatchTimeAfterTimeInterval`, 1 for the malformed-`imp` decode). The line now states the measured total instead of a derivation.

PrebidMobileTests/PrebidEventDelegateTests.swift
YuriyVelichkoPI: Agreed on the diagnosis, but out of scope for this PR — tracked.

Delegate-identity filtering is the better design; the UUID convention is a per-test discipline that will get forgotten. But this is a test-only refactor of a file that is currently green, with no behavioural bug behind it, so changing it here adds review surface to an already-large migration PR without closing a risk.

Recorded in `docs/migration/pr-phase-012.md` under "S2.7 — Deferred, with rationale".

PrebidMobile/Swift/PrebidMobileRendering/Utilities/Functions.swift
```diff
@@ -75,8 +75,266 @@ import Foundation
...
+        let screenSize = deviceScreenSize
+        let saInsets   = safeAreaInsets
+        return CGSize(width:  screenSize.width  - saInsets.left - saInsets.right,
+                      height: screenSize.height - statusBarHeight - saInsets.top - saInsets.bottom)
+    }
+
+    // `UIApplication.shared` traps when the SDK runs host-less, so resolve it through
+    // the ObjC runtime here too — `deviceMaxSize` reaches this from production code.
+    @objc public static var safeAreaInsets: UIEdgeInsets {
+        resolvedUIApplication?.keyWindow?.safeAreaInsets ?? .zero
```
mdanylov-sigma: **`safeAreaInsets` still doesn't honor the `Functions.application` injection seam.** `attemptToOpen` (line ~166) and `statusBarHeight` (line ~296) both read `Functions.application ?? sharedApplication` per the new playbook rule, but this reads `resolvedUIApplication` directly — it has to, since `PBMUIApplicationProtocol` has no `keyWindow`. Net effect: a test that sets `Functions.application = mockApplication` sees the mock reflected in `statusBarHeight` but not in `safeAreaInsets`/`deviceMaxSize` (which calls both, line ~317-318).

Worth flagging because `docs/migration/playbook.md:607` now explicitly states *"Scope: applied in Functions.swift (attemptToOpen, statusBarHeight, safeAreaInsets)"* right under the "Honour the injection seam" rule — which reads as claiming all three follow it, and this one doesn't. Either extend `PBMUIApplicationProtocol` with `keyWindow` so `Functions.application` can supply it, or narrow that doc line to be explicit about the exception.

PrebidMobileTests/RenderingTests/Tests/UtilitiesExtensionsTests/TestPBMFunctions.swift
```diff
@@ -255,35 +255,127 @@ class TestPBMFunctions: XCTestCase {
...
+
+        XCTAssertEqual(Functions.statusBarHeight, 4.0)
+
+        mockApplication.isStatusBarHidden = true
+        XCTAssertEqual(Functions.statusBarHeight, 0.0)
+    }
+
+    // `safeAreaInsets` and its caller `deviceMaxSize` must stay reachable when the
+    // SDK runs without an application host, where `UIApplication.shared` is nil.
+    func testSafeAreaInsetsWithoutApplicationHost() {
```
mdanylov-sigma, line 325: This test never sets `Functions.application`, so it only proves "doesn't crash / doesn't return NaN" — a `.zero` result (exactly what the gap above produces) satisfies `XCTAssertGreaterThanOrEqual(Functions.safeAreaInsets.top, 0.0)` trivially. It sits right next to `testStatusBarHeightUsesInjectedApplication` (which does assert against an injected mock) and reads like parallel coverage for the same fix, but it can't catch the injection-seam gap noted above either way.

PrebidMobileTests/RenderingTests/Tests/UtilitiesExtensionsTests/TestPBMFunctions.swift
```diff
@@ -255,35 +255,127 @@ class TestPBMFunctions: XCTestCase {
...
+                       startTime)
+        XCTAssertGreaterThan(Functions.dispatchTimeAfterTimeInterval(.infinity, startTime: startTime),
+                             startTime)
+        XCTAssertEqual(Functions.dispatchTimeAfterTimeInterval(-.infinity, startTime: startTime), 0)
+        XCTAssertGreaterThan(Functions.dispatchTimeAfterTimeInterval(.greatestFiniteMagnitude,
+                                                                    startTime: startTime),
+                             startTime)
+
+        // The DISPATCH_TIME_NOW branch shares the same guards.
+        XCTAssertNotEqual(Functions.dispatchTimeAfterTimeInterval(.nan), 0)
```
mdanylov-sigma: `XCTAssertNotEqual(Functions.dispatchTimeAfterTimeInterval(.nan), 0)` and the `.infinity` line below only check the result isn't exactly `0` — satisfied by any real `DispatchTime.now().rawValue` regardless of whether the NaN/infinity clamping is actually correct. Minor, but this is the same weak-assertion shape (`agents/review/SKILL.md`'s own "tests not weakened" checklist item) this PR round fixed elsewhere in this very file.

PrebidMobileTests/RenderingTests/Tests/UtilitiesExtensionsTests/TestPBMFunctions.swift
```diff
@@ -255,35 +255,127 @@ class TestPBMFunctions: XCTestCase {
...
+                       UInt64.max)
+    }
+
+    // Non-finite and out-of-range intervals saturate instead of trapping.
+    func testDispatchTimeAfterTimeIntervalWithNonFiniteIntervals() {
+        let startTime: UInt64 = 1_000_000_000
+
+        XCTAssertEqual(Functions.dispatchTimeAfterTimeInterval(.nan, startTime: startTime),
+                       startTime)
+        XCTAssertGreaterThan(Functions.dispatchTimeAfterTimeInterval(.infinity, startTime: startTime),
```
mdanylov-sigma: `XCTAssertGreaterThan(Functions.dispatchTimeAfterTimeInterval(.infinity, startTime: startTime), startTime)` (and the `.greatestFiniteMagnitude` case below it) only check direction, not magnitude. If a future edit swapped `multipliedReportingOverflow` for a wrapping `&*` in `machTicks`, an `.infinity` input could silently wrap to a small positive tick count and this would still pass. `testDispatchTimeAfterTimeIntervalWithExplicitStartTime` just above pins the actual nanosecond magnitude within a tolerance — might be worth the same rigor here (e.g. assert the deadline is within some tolerance of `UInt64.max` / the saturation ceiling).

docs/migration/playbook.md
```diff
@@ -0,0 +1,687 @@
...
+- **Honour the injection seam.** Call sites read `Functions.application ?? sharedApplication`, in
+  that order — `Functions.application` (`Functions+Testing.swift`) is how tests substitute a mock.
+  Consulting only `sharedApplication` silently ignores the mock and resolves the real singleton.
+- **Do not memoize.** The resolution is `nil` until `UIApplicationMain` has run, so a cached `nil`
+  can outlive the condition that produced it. The runtime lookup is a selector resolution plus one
+  `objc_msgSend` — not worth trading for a staleness hazard and mutable global state.
+
+**Scope:** applied in `Functions.swift` (`attemptToOpen`, `statusBarHeight`, `safeAreaInsets`), which
+no longer reads `UIApplication.shared` directly. **10 other Swift files still do, across 15 call
```
mdanylov-sigma: The literal re-measure command given, `rg -c 'UIApplication\.shared' --glob '*.swift' PrebidMobile/Swift/`, returns 11 files / 17 matches when actually run — not "10 other Swift files... 15 call sites" — because `Functions.swift` itself still matches twice (doc-comment prose at ~line 138/322, not code) and the command doesn't exclude it. The per-file counts for the other 10 files are individually correct and do sum to 15; only the aggregate is stale relative to the literal command.

PrebidMobile/Swift/PrebidMobileRendering/Utilities/Functions.swift
```diff
@@ -75,8 +75,266 @@ import Foundation
...
+
+    // MARK: - ObjC bridge (PBMFunctions+Private — Device Info)
+
+    @objc public static var deviceScreenSize: CGSize {
+        UIScreen.main.bounds.size
+    }
+
+    @objc public static var deviceMaxSize: CGSize {
+        let screenSize = deviceScreenSize
+        let saInsets   = safeAreaInsets
```
mdanylov-sigma: Minor efficiency note: `deviceMaxSize` calls both `safeAreaInsets` and `statusBarHeight`, and each now independently re-resolves the shared application through the ObjC-runtime lookup (`NSSelectorFromString`/`responds(to:)`/`perform(_:)`). Before this round `safeAreaInsets` was a cheap direct `UIApplication.shared` read; now `deviceMaxSize` pays that lookup cost twice per call instead of resolving once and sharing it.

PrebidMobile/Swift/PrebidMobileRendering/Utilities/Functions.swift
```diff
@@ -75,8 +75,266 @@ import Foundation
...
+    /// application process — e.g. in the unit-test bundle, which has no app host.
+    /// Swift imports `UIApplication.shared` as non-optional, so that `nil` becomes an
+    /// invalid reference that segfaults as soon as it is used (notably when boxed into
+    /// a `PBMUIApplicationProtocol` existential). Fetch it through the ObjC runtime
+    /// instead, so the `nil` stays observable — this is the Swift equivalent of the
+    /// `if (!uiApplication)` / `conformsToProtocol:` guards the ObjC original had.
+    ///
+    /// Deliberately not memoized: `sharedApplication` is `nil` until `UIApplicationMain`
+    /// has run, so a cached `nil` could outlive the condition that produced it.
+    private static var resolvedUIApplication: UIApplication? {
```
mdanylov-sigma: Simplification thought: `resolvedUIApplication` (`UIApplication?`) and `sharedApplication` (`PBMUIApplicationProtocol?`, which just calls the former) do the same lookup with duplicated doc comments. Since `UIApplication` already conforms to `PBMUIApplicationProtocol` via extension, a single property returning `UIApplication?` would work for every call site — `attemptToOpen`/`statusBarHeight` via automatic optional upcast, `safeAreaInsets` via `.keyWindow` directly — without the second property.

docs/migration/playbook.md
```diff
@@ -0,0 +1,687 @@
...
+    guard let applicationClass = UIApplication.self as AnyObject as? NSObjectProtocol,
+          applicationClass.responds(to: selector),
+          let application = applicationClass.perform(selector)?.takeUnretainedValue()
+    else { return nil }
+    return application as? UIApplication
+}
+```
+
+Two follow-on rules:
+
```
mdanylov-sigma, line 602: The 15 remaining `UIApplication.shared` call sites are deferred as prose-only follow-up with no lint rule or CI check stopping a new direct read from being added before those files are touched — unlike the analogous S2.5-A fix, which shipped a CI gate specifically because "a markdown-only rule is invisible to the current CI matrix." Might be worth a SwiftLint `custom_rules` regex ban (with the 15 known sites suppressed inline) so this can't silently grow, especially since the doc's own re-measure command is already slightly stale (see comment above).

docs/migration/pr-phase-012.md
```diff
@@ -0,0 +1,233 @@
...
+  commits landed on master after the branch was cut as spurious reversions), and the "tests not
+  weakened" check now names the specific anti-patterns.
+- Corrected the test-plan claim in `AGENTS.md` and the review skill: both `.xctestplan`s select by
+  *exclusion* (`skippedTests`), so new test classes need no registration.
+- Narrowed the playbook's flaky-test section to the single allowlisted test with an explicit
+  three-point confirmation, replacing the general "re-run once and move on" framing.
+- Collapsed `docs/migration/` to one PR doc: deleted the five superseded drafts
+  (`pr-phase-0/1/1-review/2/2-description.md`), which described branches that no longer exist.
+
+## S2.7 — second review round (13 findings)
```
mdanylov-sigma: Nit: this section's `## S2.7 — second review round (13 findings)` heading plus its `Fixed:`/`Deferred, with rationale` subsections don't match the `## S2.5 — Review fixes (...)` / `## S2.6 — Review fixes (round 2)` naming and flat-bullet-list structure the prior two rounds in this same file use — just a consistency nit for future scanning.

agents/xcodebuild/SKILL.md
```diff
@@ -0,0 +1,267 @@
...
+| `'PBMFoo' has been renamed to 'Foo'` | Swift test using ObjC bridging name | `perl -pi -e 's/PBMFoo\b/Foo/g'` in test files |
+| `inaccessible due to '@_spi'` | Swift test accessing Functions.* without SPI import | Add `@_spi(PBMInternal) @testable import` |
+| `cannot call value of non-function type 'String'` | ObjC method → Swift property; called with `()` | Remove `()` — it's now a property |
+| `missing arguments for parameters 'width', 'height'` | Swift init lacks no-arg form | Add `public override init() { super.init() }` with defaulted properties |
+| `cannot declare conformance to 'NSObjectProtocol'` | Mock class missing NSObject | Add `: NSObject` superclass |
+| `@objc method name provides N argument names, but method has N+1` | `throws` method needs `:error:` in explicit name | Use `@objc(name:error:)` not `@objc(name:)` |
+| `'dispatch_time' has been replaced` | C function unavailable in Swift | Call `Functions.dispatchTimeAfterTimeInterval(_:startTime:)`. **Never** `DispatchTime(uptimeNanoseconds:)` on a `dispatch_time_t` — off by ~41x on arm64 (playbook Gap S2.1-C) |
```
mdanylov-sigma: This now correctly points at `Functions.dispatchTimeAfterTimeInterval(_:startTime:)` instead of the broken pattern — good. One caveat worth adding: `Functions` is `@_spi(PBMInternal) public class`, so this works without ceremony from files inside the SDK's own modules but not from a different module (e.g. `Example/PrebidDemo/PrebidDemoObjectiveC/...`, which still has a couple of raw `dispatch_time()` calls). A contributor porting one of those following this row verbatim would hit an accessibility error this row doesn't warn about.

scripts/testPrebidDemo.sh
```diff
@@ -2,6 +2,9 @@ if [ -d "scripts" ]; then
 cd scripts/
 fi
 
+# Set bash script to exit immediately if any commands fail.
+set -e
```
mdanylov-sigma, line 6: This new `set -e` makes the script terminate immediately at the standalone `xcodebuild ... test-without-building` call on failure, before the later `if [[ ${PIPESTATUS[0]} == 0 ]] ... else echo "failed"; exit 1; fi` block is ever reached — so on a real test failure the diagnostic message is now dead code (exit code is still non-zero, so CI gating itself isn't affected). Not a new issue — `testPrebidMobile.sh`/`testPrebidMobileAdapters.sh` already have the identical pattern pre-PR — just noting this makes the fourth script consistent with an already-slightly-broken sibling idiom rather than a clean one.

mdanylov-sigma (approved):
LGTM! 
Thanks for this significant work on AI integration and Swift migration!

PR #1283, opened 2026-07-23
Fix: interstitial ad controller is dismissed when SKStoreProductViewController is presented for SKAdNetwork creatives  

### LINKED ISSUE #1282: Bug: Interstitial ad controller is dismissed when SKStoreProductViewController is presented for SKAdNetwork creatives
# Description 

When testing SKAdN creatives, tapping the ad presents the in-app App Store overlay. At the same time, the underlying interstitial ad view controller is dismissed instead of remaining in the view hierarchy.

# Expected Behavior

The `SKStoreProductViewController` should be presented on top of the interstitial ad, while the underlying ad view controller remains in the view hierarchy.

OlenaPostindustria:
Closes #1282

### DIFF
--- a/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMAbstractCreative.m
+++ b/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMAbstractCreative.m
@@ -483,10 +483,6 @@
 #pragma mark - SKStoreProductViewControllerDelegate
 
 - (void)productViewControllerDidFinish:(SKStoreProductViewController *)viewController {
-    if (viewController.presentingViewController && !viewController.isBeingDismissed) {
-        [viewController dismissViewControllerAnimated:YES completion:nil];
-    }
-
     [self.creativeViewDelegate creativeClickthroughDidClose:self];
 
     if (self.skStoreProductViewControllerExitBlock) {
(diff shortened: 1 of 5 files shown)


PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMAbstractCreative.m
```diff
@@ -480,6 +480,17 @@ - (void)modalManagerDidLeaveApp:(id<PBMModalState>)state {
     PBMLogError(@"Abstract function called");
 }
 
+#pragma mark - SKStoreProductViewControllerDelegate
+
+- (void)productViewControllerDidFinish:(SKStoreProductViewController *)viewController {
```
mdanylov-sigma, line 485: Accordingly to the [StoreKit documentation](https://developer.apple.com/documentation/storekit/skstoreproductviewcontrollerdelegate/productviewcontrollerdidfinish%28_%3A%29?changes=_8%2C_8) the delegate should call the [dismissModalViewControllerAnimated:](https://developer.apple.com/documentation/UIKit/UIViewController/dismissModalViewControllerAnimated:?changes=_8,_8) method on the view controller that originally presented the store screen.

I suppose without `[viewController dismissViewControllerAnimated:YES completion:nil]`, tapping "Cancel" or "Done" on the App Store overlay won't actually dismiss it - the overlay will stay stuck on a screen.

mdanylov-sigma (changes requested):

PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMAbstractCreative.m
OlenaPostindustria: The docs refer to the view controller that presents the `SKStoreProductViewController`, i.e. ad controller. I don't believe we should do it. `[viewController dismissViewControllerAnimated:YES completion:nil]` will do nothing here as `viewController` is `SKStoreProductViewController` and it's already dismissed.

mdanylov-sigma (approved):

PR #1286, opened 2026-07-24
Fix: video resumes when overlay is presented when returning to foreground

### LINKED ISSUE #1285: Bug: Video audio resumes on app foreground when the overlay is present
# Description

When an interstitial video ad is clicked and the App Store overlay or SafariViewController is presented, the app can be sent to the background (for example, when the user opens the App Store). If the user then returns to the app while the overlay is still displayed, the video's audio resumes playing automatically.

The video itself is not visible because it is covered by the overlay, but its audio continues playing in the background.

# Expected Behavior
The video should remain paused while the overlay is displayed. No video playback or audio should resume until the user returns to the ad.

# Root Cause
The video view remains registered for application lifecycle notifications. When the app enters the foreground, the notification handler invokes play() on the video player, causing audio playback to resume even though the video is obscured by the overlay.

OlenaPostindustria:
Closes #1285


mdanylov-sigma (approved):
LGTM

PR #1292, opened 2026-07-29
Fix: outstream video keeps playing when scrolled off-screen 

### LINKED ISSUE #1290: Outstream video keeps playing when scrolled off-screen — visibility-based pause/resume is not implemented (Android parity)
**Describe the bug**
On iOS, an outstream (banner) video ad keeps playing after it is scrolled completely off-screen — it never pauses, and quartile/completion events keep firing for an ad that is not visible. If the user has unmuted the ad, audio also continues from off-screen.

The Prebid Android SDK implements visibility-based playback control (pause when the ad leaves the viewport, resume when it re-enters), so this is an iOS/Android behavioral gap. We confirmed the gap by comparing SDK sources on both platforms (references below).

We raised this through Magnite, who responded: "Confirming this is a bug - a bug fix will need to be filed in GitHub to fix this issue."

**To Reproduce**

Load an outstream video ad (e.g. stored response prebid-demo-video-outstream, custom bidding integration / DisplayView) inside a scrollable view.
While the video is playing, scroll until the ad is completely off-screen.
Observe that playback continues (audio keeps playing if unmuted; VAST quartile/complete trackers keep firing). On Android the video pauses and resumes when scrolled back into view.

**Expected behavior**
Playback should pause when the ad becomes non-viewable and resume when it re-enters the viewport, matching the Prebid Android SDK behavior.

**Screenshots**
Screen recordings available on request (iOS: playback continues off-screen / Android: pauses and resumes).

Device: iPhone 16 (also reproduced on simulator)
OS: iOS 26.5
Prebid Mobile iOS SDK: 3.3.1

**Additional context**
Where the gap is on iOS:
- PBMAbstractCreative.onViewabilityChanged: only performs the one-time first-display action; there is no handling for the ad becoming non-viewable, and the base pause/resume hooks are never driven by visibility:
https://github.com/prebid/prebid-mobile-ios/blob/3.3.1/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMAbstractCreative.m#L442-L447
- PBMVideoCreative.onAdDisplayed stops the viewability tracker entirely after the first display, removing any means of detecting the ad leaving the screen (and the video creative has no onViewabilityChanged override):
https://github.com/prebid/prebid-mobile-ios/blob/3.3.1/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMVideoCreative.m#L110-L114

Android implementation (for comparison) — handleAdWindowFocus() → resume() / handleAdWindowNoFocus() → pause() wired to visibility changes:
https://github.com/prebid/prebid-mobile-android/blob/3.3.1/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/video/VideoCreative.java#L168-L195

DisplayView exposes no public pause/resume API either, so publishers cannot implement this on their side.

Related: https://github.com/prebid/prebid-mobile-ios/issues/1289 - another iOS/Android gap on outstream video

OlenaPostindustria:
Closes #1290

### DIFF
--- a/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMVideoCreative.m
+++ b/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMVideoCreative.m
@@ -39,6 +39,9 @@
 @property (nonatomic, strong) NSData *data;
 @property (nonatomic, strong) PBMRewardedConfig *rewardedConfig;
 
+// Whether playback should follow the viewability of the ad view.
+@property (nonatomic, assign, readonly) BOOL isVisibilityDrivenPlayback;
+
 @end
 
 #pragma mark - Implementation
@@ -50,6 +53,10 @@
 + (NSInteger)maxSizeForPreRenderContent {
     // 25 MiB
     return 25 * 1024 * 1024;
+}
+
+- (BOOL)isVisibilityDrivenPlayback {
+    return !self.creativeModel.adConfiguration.presentAsInterstitial;
 }
 
 #pragma mark - PBMAbstractCreative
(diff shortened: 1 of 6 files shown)


mdanylov-sigma (approved):
LGTM!

PR #1294, opened 2026-07-29
Fix: use system tap gesture recognizer

### LINKED ISSUE #1289: Outstream video ads register a click on touch-down — scroll gestures fire click trackers and open the clickthrough
**Describe the bug**
On iOS, clicks on outstream (banner) video ads are detected with a custom gesture recognizer (PBMTouchDownRecognizer) that succeeds the moment a finger touches the ad view (touchesBegan). It does not wait for touch-up and does not check for movement, so a scroll gesture that merely starts on the ad is treated as a click: click trackers (VAST ClickTracking / creativeWasClicked) fire, and if the creative has a clickthrough URL, the landing page opens while the user was only trying to scroll.

Android uses the OS-standard click semantics (a click registers only when the finger lifts without significant movement; scrolls are intercepted by the parent scroll view), so this is an iOS/Android behavioral gap. It inflates CTR on iOS and degrades UX in scrollable feeds.

We raised this through Magnite, who responded: "This is also the behavior that was implemented like this intentionally, but perhaps incorrect product decision, so worth a GitHub issue to be fixed."

**To Reproduce**

Load an outstream video ad (e.g. stored response prebid-demo-video-outstream, custom bidding integration / DisplayView) inside a scrollable view.
Place a finger on the ad and drag vertically to scroll the feed.
Observe that a click event fires immediately on touch-down (and the clickthrough opens if the creative has a click URL), even though the gesture was a scroll.

**Expected behavior**
A click should be registered only on a standard tap (touch-up without movement), matching UIKit conventions and the Prebid Android SDK behavior. Touches that turn into scroll gestures should not fire click trackers or open the clickthrough.

**Screenshots**
Screen recording available on request (click log firing while scrolling with the finger resting on the ad).

Device: iPhone 16 (also reproduced on simulator)
OS: iOS 26.5
Prebid Mobile iOS SDK: 3.3.1

**Additional context**

- Root cause: PBMTouchDownRecognizer transitions to .ended inside touchesBegan, so any touch-begin is a successful "click":
https://github.com/prebid/prebid-mobile-ios/blob/3.3.1/PrebidMobile/Objc/PrebidMobileRendering/ExtensionsAndWrappers/PBMTouchDownRecognizer.m
- Consumed in PBMVideoView.setupTapRecognizer / recordTapEvent::
https://github.com/prebid/prebid-mobile-ios/blob/3.3.1/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMVideoView.m#L1001-L1021
- Android (correct behavior for comparison) uses standard OnClickListener on the video view:
https://github.com/prebid/prebid-mobile-android/blob/3.3.1/PrebidMobile/PrebidMobile-core/src/main/java/org/prebid/mobile/rendering/video/VideoCreativeView.java
- There is no public API to customize the gesture recognition, so publishers cannot work around this.

Related: https://github.com/prebid/prebid-mobile-ios/issues/1290 - another iOS/Android gap on outstream video

OlenaPostindustria:
Closes #1289


mdanylov-sigma (approved):
LGTM

PR #1301, opened 2026-07-31
Fix: SKAdN response fields according to IAB spec

### LINKED ISSUE #1300: Bug: SKAdNetwork response fields decoded with wrong data types
# Description

`ORTBBidExtSkadn` and `ORTBSkadnFidelity` type several `bid.ext.skadn` fields as `NSNumber / UUID`, but the [IAB SKAdNetwork community extension](https://github.com/InteractiveAdvertisingBureau/openrtb/blob/main/extensions/community_extensions/skadnetwork.md) defines all of them as strings.

# To Reproduce

Parse a spec-compliant bid response:

```json
{
  "version": "2.2",
  "network": "cDkw7geQsH.skadnetwork",
  "campaign": "45",
  "itunesitem": "123456789",
  "sourceapp": "880047117",
  "fidelities": [{
    "fidelity": 0,
    "nonce": "473b1a16-b4ef-43ad-9591-fcf3aefa82a7",
    "timestamp": "1594406342",
    "signature": "MEQCIEQ..."
  }]
}
```

OlenaPostindustria:
Closes #1300

### DIFF
--- a/PrebidMobile/Swift/PrebidMobileRendering/AdTypes/AdView/SKOverlayManager.swift
+++ b/PrebidMobile/Swift/PrebidMobileRendering/AdTypes/AdView/SKOverlayManager.swift
@@ -69,7 +69,7 @@ public class SKOverlayManager: NSObject {
             return nil
         }
         
-        guard let itunesitem = skadnInfo.itunesitem?.stringValue else {
+        guard let itunesitem = skadnInfo.itunesitem else {
             Log.warn("SDK failed to build SKOverlay configuration. `itunesitem` is nil.")
             return nil
         }
--- a/PrebidMobile/Swift/PrebidMobileRendering/AdTypes/AdView/SKOverlayManager.swift
+++ b/PrebidMobile/Swift/PrebidMobileRendering/AdTypes/AdView/SKOverlayManager.swift
@@ -63,7 +63,7 @@ public class SKOverlayManager: NSObject {
     }
     
     @available(iOS 14.0, *)
-    private func buildConfig(with skadnInfo: ORTBBidExtSkadn) -> SKOverlay.AppConfiguration? {
+    func buildConfig(with skadnInfo: ORTBBidExtSkadn) -> SKOverlay.AppConfiguration? {
         guard let skoverlay = skadnInfo.skoverlay else {
             Log.warn("SDK failed to build SKOverlay configuration. `skoverlay` dictionary is nil.")
             return nil
@@ -74,10 +74,15 @@ public class SKOverlayManager: NSObject {
             return nil
         }
         
+        guard let appIdentifier = itunesitem.strictNumberValue?.stringValue else {
+            Log.warn("SDK failed to build SKOverlay configuration. `itunesitem` is not a valid App Store item identifier: \(itunesitem).")
+            return nil
+        }
+        
         let position: SKOverlay.Position = skoverlay.pos == 0 ? .bottom : .bottomRaised
         
         let config = SKOverlay.AppConfiguration(
-            appIdentifier: itunesitem,
+            appIdentifier: appIdentifier,
             position: position
         )
         
--- a/PrebidMobile/Swift/PrebidMobileRendering/AdTypes/AdView/SKOverlayManager.swift
+++ b/PrebidMobile/Swift/PrebidMobileRendering/AdTypes/AdView/SKOverlayManager.swift
@@ -74,15 +74,10 @@ public class SKOverlayManager: NSObject {
             return nil
         }
         
-        guard let appIdentifier = itunesitem.strictNumberValue?.stringValue else {
-            Log.warn("SDK failed to build SKOverlay configuration. `itunesitem` is not a valid App Store item identifier: \(itunesitem).")
-            return nil
-        }
-        
         let position: SKOverlay.Position = skoverlay.pos == 0 ? .bottom : .bottomRaised
         
         let config = SKOverlay.AppConfiguration(
-            appIdentifier: appIdentifier,
+            appIdentifier: itunesitem,
             position: position
         )
         
--- a/PrebidMobile/Swift/PrebidMobileRendering/SkadnParametersManager.swift
+++ b/PrebidMobile/Swift/PrebidMobileRendering/SkadnParametersManager.swift
@@ -89,8 +89,8 @@ public class SkadnParametersManager: NSObject {
             }
             
             if #available(iOS 16.1, *) {
-                if let numberSourceidentifier = skadnInfo.sourceidentifier?.strictNumberValue {
-                    productParams[SKStoreProductParameterAdNetworkSourceIdentifier] = numberSourceidentifier
+                if let sourceidentifier = skadnInfo.sourceidentifier {
+                    productParams[SKStoreProductParameterAdNetworkSourceIdentifier] = sourceidentifier
                 }
             }
             
--- a/PrebidMobile/Swift/PrebidMobileRendering/SkadnParametersManager.swift
+++ b/PrebidMobile/Swift/PrebidMobileRendering/SkadnParametersManager.swift
@@ -30,23 +30,16 @@ public class SkadnParametersManager: NSObject {
         return nil
     }
     
-    /// The ORTB SKAdNetwork extension types these identifiers as strings of digits, so they are
-    /// parsed strictly: `Int64` is locale-independent - unlike `NumberFormatter`, which follows
-    /// `Locale.current` - and rejects both non-numeric and out-of-range values.
-    private static func number(from string: String) -> NSNumber? {
-        Int64(string).map { NSNumber(value: $0) }
-    }
-    
     @available(iOS 14.5, *)
     public static func getSkadnImpression(for skadnInfo: ORTBBidExtSkadn) -> SKAdImpression? {
         guard let fidelity = getFidelity(from: skadnInfo, fidelityType: 0) else { return nil }
         
         let imp = SKAdImpression()
-        if let itunesitem = skadnInfo.itunesitem, let numberItunesitem = number(from: itunesitem),
+        if let numberItunesitem = skadnInfo.itunesitem?.strictNumberValue,
            let network = skadnInfo.network,
-           let sourceapp = skadnInfo.sourceapp, let numberSourceapp = number(from: sourceapp),
+           let numberSourceapp = skadnInfo.sourceapp?.strictNumberValue,
            let nonce = fidelity.nonce,
-           let timestamp = fidelity.timestamp, let numberTimestamp = number(from: timestamp),
+           let numberTimestamp = fidelity.timestamp?.strictNumberValue,
            let signature = fidelity.signature,
            let version = skadnInfo.version {
             imp.sourceAppStoreItemIdentifier = numberSourceapp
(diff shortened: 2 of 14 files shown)


PrebidMobile/Swift/PrebidMobileRendering/SkadnParametersManager.swift
```diff
@@ -30,28 +30,34 @@ public class SkadnParametersManager: NSObject {
...
+    private static func number(from string: String) -> NSNumber? {
+        Int64(string).map { NSNumber(value: $0) }
+    }
+    
     @available(iOS 14.5, *)
     public static func getSkadnImpression(for skadnInfo: ORTBBidExtSkadn) -> SKAdImpression? {
-        guard let fidelity = SkadnParametersManager.getFidelity(from: skadnInfo, fidelityType: 0) else { return nil }
+        guard let fidelity = getFidelity(from: skadnInfo, fidelityType: 0) else { return nil }
+        
+        let formatter = NumberFormatter()
```
mdanylov-sigma: **Severity: High — correctness**

`getSkadnImpression` parses `itunesitem`/`sourceapp`/`campaign`/`timestamp` with a locale-aware `NumberFormatter()` (default `Locale.current`) instead of the strict `Int64`-based helper defined a few lines above (`number(from:)`, lines 33-35), which is never called anywhere in the file or repo.

**Failure scenario:** On a comma-decimal locale (e.g. `de_DE`, `fr_FR`), a malformed bid value like `"itunesitem": "1,234"` is parsed by `NumberFormatter` as the *Double* `1.234` and forwarded to `SKAdImpression.advertisedAppStoreItemIdentifier`. On an `en_US` device the exact same server response yields `nil` and the impression is safely dropped. So the same malformed input is either silently corrupted or safely rejected purely depending on device locale. Empirically verified: `NumberFormatter().number(from: "1,234")` returns non-nil under `de_DE`/`fr_FR`/`en_UA` but `nil` under `en_US`.

Separately, an oversized identifier like `"99999999999999999999"` is leniently accepted as the inaccurate Double `1e20` instead of being rejected (see the PR's own `testGetSkadnImpression_withOverflowingIdentifier`).

Suggest wiring in the existing `number(from:)` helper (strict `Int64`, locale-independent, rejects overflow) instead of `formatter.number(from:)` for all four fields.

PrebidMobile/Swift/PrebidMobileRendering/AdTypes/AdView/SKOverlayManager.swift
```diff
@@ -69,7 +69,7 @@ public class SKOverlayManager: NSObject {
             return nil
         }
         
-        guard let itunesitem = skadnInfo.itunesitem?.stringValue else {
+        guard let itunesitem = skadnInfo.itunesitem else {
```
mdanylov-sigma, line 72: **Severity: Medium — correctness**

`buildConfig` lost its numeric validation of `itunesitem` when the property was retyped from `NSNumber` to `String`. Before this PR, `itunesitem` was populated via `numberOrString`, so a malformed value became `nil` and this guard cleanly bailed with a warning log. Now `itunesitem` comes from `stringOrNumber`, which returns *any* string verbatim, and this guard only checks for `nil`, not numeric content.

**Failure scenario:** a bid server sends `"itunesitem": ""` (or other non-numeric junk). The guard passes, and `SKOverlay.AppConfiguration(appIdentifier: "", position:)` is built and handed to Apple's `SKOverlay` API with a garbage identifier — silently, with no equivalent warning log to the one `SkadnParametersManager` does for the same field in this same PR. There's no `SKOverlayManagerTest` covering this path.

Suggest validating `itunesitem` is numeric here too before building the config (or centralizing that validation).

mdanylov-sigma (commented):
Automated review pass (Claude Code) — 6 findings, ranked by severity.

PrebidMobile/Swift/PrebidMobileRendering/SkadnParametersManager.swift
```diff
@@ -30,28 +30,34 @@ public class SkadnParametersManager: NSObject {
         return nil
     }
     
+    private static func number(from string: String) -> NSNumber? {
```
mdanylov-sigma: **Severity: Low — simplification (dead code)**

`private static func number(from string: String) -> NSNumber?` is never called anywhere — every real conversion below uses `formatter.number(from:)` (the `NumberFormatter` instance method) instead. Its near-identical name/signature makes it a trap for future maintainers who might assume it's what's actually wired in. Either use it (see the locale-bug comment above) or remove it.

PrebidMobile/Swift/Utils/JSONParsing.swift
```diff
@@ -106,10 +106,32 @@ struct JSONObject<Key: RawRepresentable> where Key.RawValue == String {
...
+    private func number(_ key: Key) -> NSNumber? {
+        guard let value = dict[key.rawValue] as? NSNumber,
+              CFGetTypeID(value) != CFBooleanGetTypeID() else { return nil }
+
+        return value
+    }
+
+    // For fields the IAB spec defines as strings, but which bid servers are known
+    // to send as JSON numbers (e.g. `itunesitem`, `sourceapp`, `campaign`, `timestamp`).
+    func stringOrNumber(_ key: Key) -> String? {
```
mdanylov-sigma, line 109: **Severity: Low — reuse (dead code)**

Separately from this hunk: the generic `subscript(_ key: Key) -> UUID?` at line 52 of this file is now unused across the whole repo. Its only caller was `ORTBSkadnFidelity.nonce`, which this PR retypes from `UUID?` to `String?` (using the plain `String` subscript instead, next to the new `stringOrNumber`/`numberOrString` helpers added here). Worth removing along with the rest of this diff, since it's shared, widely-imported infrastructure carrying dead surface area. (Anchoring this comment here since line 52 itself falls outside any diff hunk.)

PrebidMobile/Swift/PrebidMobileRendering/SkadnParametersManager.swift
```diff
@@ -69,40 +75,42 @@ public class SkadnParametersManager: NSObject {
...
-                if #available(iOS 16.1, *) {
-                    if let sourceIdentifier = skadnInfo.sourceidentifier, let sourceidentifierInteger = Int(sourceIdentifier) {
-                        productParams[SKStoreProductParameterAdNetworkSourceIdentifier] = NSNumber(value: sourceidentifierInteger)
-                    }
+            if let campaign = skadnInfo.campaign {
+                productParams[SKStoreProductParameterAdNetworkCampaignIdentifier] = campaign
+            }
+            
+            if #available(iOS 16.1, *) {
+                if let sourceIdentifier = skadnInfo.sourceidentifier, let sourceidentifierInteger = Int(sourceIdentifier) {
```
mdanylov-sigma: **Severity: Low — altitude**

This PR introduces lenient `NumberFormatter`-based parsing for `itunesitem`/`sourceapp`/`campaign`/`timestamp` (see `getSkadnImpression` above), while the structurally identical `sourceidentifier` field here (and its twin at line 65 in `getSkadnImpression`) is still parsed with strict `Int(sourceIdentifier)!`. Same shape of data (a spec numeric-ID string), two different validation behaviors, purely as an artifact of which fields this PR touched. A single shared "parse spec numeric-ID string" helper (e.g. the `number(from:)` static func already sitting unused at line 33) would remove the inconsistency.

PrebidMobile/Swift/PrebidMobileRendering/SkadnParametersManager.swift
mdanylov-sigma: **Severity: Low — efficiency**

`NumberFormatter()` is constructed fresh on every `getSkadnImpression` call, which fires once per ad impression (called from `PBMAbstractCreative.m`, `AdUnit.swift`, `NativeAd.swift`). `NumberFormatter` init involves locale table setup and is measurably pricier than the `Int64(string)` parse the now-dead helper at line 33 already does. Low severity — the codebase has precedent for this pattern elsewhere (`String+Extensions.swift`) — but it's newly added on this hot path by this diff.

mdanylov-sigma:
This PR contains public API changes, could we postpone merging for the next major release ?

YuriyVelichkoPI:
@OlenaPostindustria, what is the rationale for this fix?

I see that the internal model does not fully match the IAB specification. At the same time, it appears that the SDK simply converts values into the types required to call the relevant OS APIs. For example, `getSkadnImpression` converts `sourceapp` to `NSNumber` and assigns it to `SKAdImpression.sourceAppStoreItemIdentifier`, which is also an `NSNumber`.

My understanding is that the current implementation intentionally relies on OS-specific types rather than RTB-specific ones. Unless there is a strict requirement, I do not think the internal Swift model needs to mirror the JSON protocol model exactly.

@mdanylov-sigma, could you clarify what you mean by “public API changes”? The changed objects appear to be exposed to SDK users through `PrebidMediationDelegate`. Are there other ways these changes could affect SDK integration code?

OlenaPostindustria:
@YuriyVelichkoPI The change is only where the conversion happens, and there are a few reasons it belongs at the point where we build the SKAdN parameters rather than at decode time:

1. Where Apple doesn't force a type, we don't impose one. The `loadProduct` parameters are just `[String: Any]`, so we pass the server's values through as-is and let StoreKit validate them.
2. `campaign` and `sourceidentifier` were read as plain `NSNumber`, so an IAB-compliant response sending it as a string silently decoded to nil - the campaign and source identifiers were dropped from the product parameters, breaking attribution for spec-compliant responses.
3. Decode-time conversion can silently alter signed values. `nonce` was decoded into a UUID and re-emitted via `uuidString`, which potentially can change the value. Keeping the raw string and converting only at the API boundary guarantees we hand Apple exactly what the server sent.
4. There is no single OS type for `itunesitem` - it must be a `NSNumber` for `SKAdImpression`, but a `String` for `SKOverlay`. The previous code converted String => NSNumber at decode and then back for SKOverlay.

mdanylov-sigma:
`getSkadnProductParameters` (`SkadnParametersManager.swift`) builds the dictionary passed to `SKStoreProductViewController.loadProduct(withParameters:)` by forwarding raw `String` for **six** keys — `itunesitem`, `campaign`, `sourceapp`, `timestamp`, `nonce`, and now (since `e62ff78f`) `sourceidentifier` too. I checked this against this machine's actual iOS 26.5 SDK headers rather than going from memory:

```
// SKStoreProductViewController.h
SK_EXTERN NSString * const SKStoreProductParameterITunesItemIdentifier; // iTunes Store item identifier (NSNumber) of the product

// SKAdNetwork.h
SK_EXTERN NSString * const SKStoreProductParameterAdNetworkCampaignIdentifier; // (NSNumber)
SK_EXTERN NSString * const SKStoreProductParameterAdNetworkNonce; // Random entropy value for security (NSUUID)
SK_EXTERN NSString * const SKStoreProductParameterAdNetworkTimestamp; // (NSNumber)
SK_EXTERN NSString * const SKStoreProductParameterAdNetworkSourceAppStoreIdentifier; // (NSNumber)

// SKAdImpression.h
@property (nonatomic, strong) NSNumber *sourceIdentifier; // confirms SourceIdentifier is numeric too
```

Every one of those keys is documented as `NSNumber` or `NSUUID`. `getSkadnProductParameters` now hands StoreKit a plain `String` for all six, with no conversion — unlike the sibling function `getSkadnImpression` a few lines above, which correctly reconstructs `NSNumber` via `.strictNumberValue`. This isn't a malformed-input edge case: it's every normal SKAdNetwork 2.2+/4.0 tap-through that reaches the product-page click-through flow (`PBMAbstractCreative.m` → `SKStoreProductViewControllerPresenter.swift`). `itunesitem`/`campaign`/`sourceapp`/`timestamp`/`nonce` have actually been wrong-typed here since the PR's original commit — `sourceidentifier` was correctly `NSNumber`-converted until `e62ff78f` removed that conversion.

Suggested fix: apply the same `.strictNumberValue` (and `NSUUID(uuidString:)` for `nonce`) conversion in `getSkadnProductParameters` that `getSkadnImpression` already does, for all six fields.

OlenaPostindustria:
@mdanylov-sigma thanks for flagging that, the docs indeed mention the exact expected type, made some adjustments

mdanylov-sigma (approved):
LGTM!

YuriyVelichkoPI (approved):

Issue #1056, opened 2024-10-10
Rewarded Ad Unit Enhancements


teqblaze-yurii:
The current implementation of the `RewardedAdUnit` is limited by the behavior of the general-purpose Interstitial Ad and doesn’t allow configuring either the Reward object or the UX, which is critical for this ad format.

This proposal describes how the Rewarded Ad Unit should work according to the industry and publishers’ expectations and what changes for the `RewardedAdUnit` class will make it more usable and publisher-oriented.

The **Teqblaze** team is already working on implementing this proposal and will introduce the respective PRs soon. 

The proposed changes are based on the market research and Teqblaze’s implementation of the Rewarded Ad Unit. They don’t break the current behavior of `RewardedAdUnit` but extend it.

## The Concept of Rewarded Ads

The mechanic of the rewarded ad has two distinct features:

- Special displaying properties that don't allow users to skip the ad.  
- Providing a **Reward** for viewing the ad once the displaying conditions are met. The **Reward** is a generic object the publisher configures on the server. It is specific to the application or game. 
 
Here is the general workflow of utilizing Rewarded Ad Units in the apps:  

![RewardedAds Consept](https://github.com/user-attachments/assets/13c185cc-f95b-4300-ac28-2e8e6e490fdf)

1. The Publisher creates a Rewarded Ad Unit on the ad platform with a particular size, format, etc. These configurations are similar to interstitial ad unit configuration. *In the case of Prebid, these configurations will be set via stored configs on PBS.*
2. The Publisher configures the Reward Object associated with the ad unit. This object contains metadata like reward *type* and reward *count*. *In the case of Prebid publisher will create a Reward object only in the case of No Ad Server integration. In all other cases, the Reward will be set on the particular ad server. However, the UX configuration should still be set via PBS stored configuration.*
3. The SDK makes an ad request for the Rewarded Ad Unit. SDK also adds an `imp.rwdd:1` flag to indicate that the request is a subject for the rewarded ad place.  
4. The Ad Platform broadcasts the request to the bidders. The `imp.rwdd:1` flag signals about the specific ad unit behavior so Demand partners can bid on it as on the premium inventory.  
5. The buyer responds with a bid. The bid CPM for the rewarded ad unit might be higher. Because the buyer knows that the rewarded ads are usually non-skippable and can be opt-in, this suggests better user engagement and motivates higher CPMs to be paid.  
[...]

PR #1305, opened 2026-08-04
Fix #1296: configurable video interstitial auto-close

### LINKED ISSUE #1296: Video interstitials without an end card auto-close after playback — no API to disable auto-close and require an explicit close
**Describe the bug**

A video interstitial without a companion (end card) automatically dismisses itself as soon as playback completes, with zero user interaction. This is hardcoded in the SDK: `PBMVideoCreative.videoViewCompletedDisplay` calls `creativeDisplayCompleted` and then invokes `dismissInterstitialModalState()` directly (the code comment says *"no companion ads so pass this event to the PBMModalManager and close video automatically"*).

There is no public API to disable this auto-close behavior or to keep the ad on screen until the user taps the close button. Rewarded ads can control the close behavior via the server-side rewarded configuration (`rwdd.close.action`), but interstitials have no equivalent — neither server-side nor client-side.

This differs from the interstitial UX of most ad SDKs (user-controlled dismissal) and creates an inconsistent flow for publishers running multiple networks.

> It's a default behaviour of the interstitials without an end card. There should always be a way to disable this auto-close behaviour and make the ad close only when the close button was tapped. However, it seems to be missing in the API. Please file an issue in the corresponding SDK repo.

**To Reproduce**

1. Load and present a video interstitial without an end card (e.g. stored response `prebid-ita-video-interstitial-320-480`, custom bidding integration / `InterstitialController`).
2. Do not interact with the ad; let the video play to the end.
3. The interstitial dismisses itself automatically at playback completion.

**Expected behavior**

Playback-completion behavior should be controllable by the publisher (e.g. via `VideoControlsConfiguration` / `AdConfiguration`): keep the interstitial on screen with a close button after the video ends, instead of force-dismissing. Keeping the current behavior as the default is fine — an opt-out is what's missing.

**Screenshots**

Screen recording available on request.

 - OS: iOS 26.5

**Additional context**

- Auto-dismiss implementation (no configuration gate): https://github.com/prebid/prebid-mobile-ios/blob/3.3.1/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMVideoCreative.m#L167-L190
- The skip-delay control exists ([combined ad experience controls — skip delay](https://docs.prebid.org/prebid-mobile/modules/rendering/combined-ad-experience-controls.html#skip-delay)) but it only controls when the skip/close button appears — not the auto-close at completion.
- Android counterpart (same API gap, reported separately): prebid/prebid-mobile-android#972

mdanylov-sigma:
## Summary

Adds a publisher-configurable `isAutoCloseOnCompletionEnabled` flag (default `true`, preserving existing behavior) on `VideoControlsConfiguration`, exposed on `InterstitialRenderingAdUnit`, `RewardedAdUnit`, and `MediationBaseInterstitialAdUnit`. When disabled, a full-screen video interstitial without a companion/end card no longer auto-dismisses when playback completes — the publisher (or the rewarded close-action flow) controls when it closes.

Closes #1296

## Behavior

- Default (`true`): unchanged from current behavior — video without a companion still auto-closes on completion.
- `false`: `[modalManager creativeDisplayCompleted:]` is still called unconditionally (so rewarded completion is still reported to `ModalManager` and rewarded close-action handling — `closeButton`/`autoClose`/`unknown` — still runs correctly), but the automatic `dismissInterstitialModalState()` call is skipped.
- Non-rewarded interstitials without a companion still get a close ("X") button via the existing `closeDelay` countdown in `ModalViewController`, independent of this flag, so they can never get stuck.

## Testing

- `PrebidMobileTests` (unit): **25/25 passed**, including new tests:
  - `testInterstitialWithoutCompanionAutoClosesByDefault`
  - `testInterstitialWithoutCompanionCanDisableAutoClose`
  - `testRewardedCompletionIsReportedWhenAutoCloseIsDisabled` — explicitly verifies `ModalManager.creativeDisplayCompleted(_:)` still fires for a rewarded ad when auto-close is disabled.
  - `testAutoCloseOnCompletion` (config default/override)
- `PrebidDemoSwiftUITests` (real UI test, live ad serving): `testInAppVideoInterstitialAd` passed.

## API / compatibility

- Additive `@objcMembers` property, consistent with the existing `isSoundButtonVisible` pattern across all three ad unit surfaces — no Objective-C interop concerns, no breaking changes, default preserves current behavior for existing integrations.

### DIFF
--- a/PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/RewardedAdUnit.swift
+++ b/PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/RewardedAdUnit.swift
@@ -83,7 +83,8 @@ public class RewardedAdUnit: NSObject, BaseInterstitialAdUnitProtocol {
         set { adUnitConfig.adConfiguration.videoControlsConfig.isSoundButtonVisible = newValue }
     }
 
-    /// A Boolean value indicating whether video ads without an end card close when playback completes.
+    /// Controls whether full-screen video ads without an end card close when playback completes.
+    /// Set to `false` to keep the ad open with a **Watch Again** button. The default value is `true`.
     public var isAutoCloseOnCompletionEnabled: Bool {
         get { adUnitConfig.adConfiguration.videoControlsConfig.isAutoCloseOnCompletionEnabled }
         set { adUnitConfig.adConfiguration.videoControlsConfig.isAutoCloseOnCompletionEnabled = newValue }
--- a/PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/RewardedAdUnit.swift
+++ b/PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/RewardedAdUnit.swift
@@ -83,13 +83,6 @@ public class RewardedAdUnit: NSObject, BaseInterstitialAdUnitProtocol {
         set { adUnitConfig.adConfiguration.videoControlsConfig.isSoundButtonVisible = newValue }
     }
 
-    /// Controls whether full-screen video ads without an end card close when playback completes.
-    /// Set to `false` to keep the ad open with a **Watch Again** button. The default value is `true`.
-    public var isAutoCloseOnCompletionEnabled: Bool {
-        get { adUnitConfig.adConfiguration.videoControlsConfig.isAutoCloseOnCompletionEnabled }
-        set { adUnitConfig.adConfiguration.videoControlsConfig.isAutoCloseOnCompletionEnabled = newValue }
-    }
-    
     // MARK: Internal Properties
     
     // Note: exposed for tests
--- a/PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/RewardedAdUnit.swift
+++ b/PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/RewardedAdUnit.swift
@@ -82,6 +82,12 @@ public class RewardedAdUnit: NSObject, BaseInterstitialAdUnitProtocol {
         get { adUnitConfig.adConfiguration.videoControlsConfig.isSoundButtonVisible }
         set { adUnitConfig.adConfiguration.videoControlsConfig.isSoundButtonVisible = newValue }
     }
+
+    /// A Boolean value indicating whether video ads without an end card close when playback completes.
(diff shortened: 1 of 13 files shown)


OlenaPostindustria:
When `isAutoCloseOnCompletionEnabled` is set to false, the video just ends without providing an option to replay. Since video banners have a `Watch Again` button in this scenario, would it make sense to make the behaviour consistent? Alternatively, we could have the video automatically loop until the ad is closed.

OlenaPostindustria (dismissed):
LGTM, left a suggestion

OlenaPostindustria:
It would also be good to document the new property in the public documentation

OlenaPostindustria (dismissed):
LGTM!

PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/RewardedAdUnit.swift
```diff
@@ -82,6 +82,13 @@ public class RewardedAdUnit: NSObject, BaseInterstitialAdUnitProtocol {
         get { adUnitConfig.adConfiguration.videoControlsConfig.isSoundButtonVisible }
         set { adUnitConfig.adConfiguration.videoControlsConfig.isSoundButtonVisible = newValue }
     }
+
+    /// Controls whether full-screen video ads without an end card close when playback completes.
+    /// Set to `false` to keep the ad open with a **Watch Again** button. The default value is `true`.
+    public var isAutoCloseOnCompletionEnabled: Bool {
```
YuriyVelichkoPI: Rewarded ad unit has own configuration approach 
https://github.com/prebid/prebid-mobile-ios/issues/1056 

Does this property align with it? Can publisher disable auto close via `rwdd` properties?  It looks like cusomization of `close.action` object can change the behaviour in required way.

YuriyVelichkoPI (changes requested):
Please support new property in ORTBAdConfiguration and update the spec for Ad Experience Controls respectively

https://docs.prebid.org/prebid-mobile/modules/rendering/combined-ad-experience-controls.html

mdanylov-sigma:
Addressed in 097c52d0: the value now parses through ORTBAdConfiguration, rewarded ads defer to rwdd.close.action, and the public documentation update is in prebid/prebid.github.io#6697.

YuriyVelichkoPI (approved):
LGTM

PR #1306, opened 2026-08-04
Fix #1297: configurable rewarded completion timeout

### LINKED ISSUE #1297: Rewarded end card locks the user for ~120 seconds when the bid omits the rewarded configuration (rwdd) — fallback is hardcoded and not configurable
**Describe the bug**

For rewarded ads, the close control is withheld until the reward completion criteria are met — which is fine when the bid contains a rewarded configuration (`ext.prebid.passthrough[].rwdd`). However, when the bid **omits** `rwdd` (or its completion criteria), the SDK falls back to a hardcoded default completion time of **120 seconds** (`RewardedConfig.defaultCompletionTime`).

For a creative with an end card, this means: the video finishes, the end card appears, and the user is stuck on the end card for roughly **2 minutes with no close button** before the close/auto-close logic finally runs. There is no public API to adjust this fallback, and 120 seconds is excessive as a default from a UX standpoint.

> Is there a public API to configure this default? Unfortunately not.

(they also confirmed that `rwdd` *should* normally be present for rewarded video and can be changed in the ad response — but when it is omitted for any reason, the publisher has no way to mitigate the 120-second lock.)

**To Reproduce**

1. Load a rewarded video whose bid has no `rwdd` in `ext.prebid.passthrough` (e.g. stored response `prebid-demo-video-rewarded-endcard-default`, custom bidding integration / `InterstitialController` with `isRewarded = true` — same approach as the AdMob adapter).
2. Watch the video to completion; the end card appears.
3. No close button is shown for ~120 seconds; the user cannot leave the ad during that time.

**Expected behavior**

Either (or both) of:
- a more reasonable default completion time when the bid omits `rwdd`, and/or
- a public API (e.g. on `RewardedConfig` / `AdConfiguration`) allowing the publisher to configure the fallback completion time and close behavior for bids without a rewarded configuration.

**Screenshots**

Screen recording available on request.

 - OS: iOS 26.5

**Additional context**

- Hardcoded default: https://github.com/prebid/prebid-mobile-ios/blob/3.3.1/PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/RewardedConfig.swift#L71
- Where the default is applied to the end card timer (`endcardTime ?? defaultCompletionTime`): https://github.com/prebid/prebid-mobile-ios/blob/3.3.1/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/PBMHTMLCreative.m#L205-L233
- Rewarded UI/UX policy for reference: https://docs.prebid.org/prebid-mobile/modules/rendering/combined-ui-ux-policy.html#rewarded-ad
- Related iOS/Android close-behavior API gap: prebid/prebid-mobile-android#972

mdanylov-sigma:
## Summary

Makes the previously hard-coded 120s `RewardedConfig.defaultCompletionTime` publisher-configurable via a new `rewardedCompletionTimeout` property on `VideoControlsConfiguration` (default `120`, matching prior behavior), with pass-throughs on `RewardedAdUnit`, `MediationBaseInterstitialAdUnit`, and a new direct property on `InterstitialController`.

Closes #1297

## Behavior

- Rejects negative values safely (logs a warning, keeps the previous value) rather than throwing or silently accepting bad state.
- Propagates through both `DisplayView.loadAd()` and `InterstitialController.loadAd()` into `RewardedConfig.defaultCompletionTime` before the transaction factory is created.
- Used **only** as a fallback for the endcard/banner rewarded-completion timer (`PBMHTMLCreative.setupRewardTimerIfNeeded`) when the bid omits `endcardTime`/`bannerTime` — it does not affect the primary VAST video reward-completion path (playback-event/percentage based), which is unrelated and untouched.

## Testing

- `PrebidMobileTests` (unit): **22/22 passed**, including new tests:
  - `testRewardedCompletionTimeout` (default/override/negative-rejection)
  - `testInterstitialControllerPropagatesRewardedCompletionTimeout` — exercises `InterstitialController.loadAd()` end-to-end and asserts `rewardedConfig.defaultCompletionTime` reflects the configured value.
  - `testDisplayViewPropagatesRewardedCompletionTimeout` — same for `DisplayView.loadAd()`.
  - `testRewardEvent_EndcardUsesConfiguredFallback`

## API / compatibility

- `RewardedConfig.defaultCompletionTime` changes `let` → `var`: additive, non-breaking for Objective-C consumers.
- New `TimeInterval` properties are plain `@objcMembers`-compatible, consistent with existing config surface.

### DIFF
--- a/PrebidMobile/Swift/PrebidMobileRendering/AdTypes/AdView/VideoControlsConfiguration.swift
+++ b/PrebidMobile/Swift/PrebidMobileRendering/AdTypes/AdView/VideoControlsConfiguration.swift
@@ -90,6 +90,18 @@ public class VideoControlsConfiguration: NSObject {
     
     /// This property indicates whether mute controls is visible on the screen.
     public var isSoundButtonVisible = false
+
+    /// The fallback rewarded-ad completion timeout, in seconds, used when the bid does not provide completion criteria.
+    public var rewardedCompletionTimeout: TimeInterval {
+        set {
+            if newValue >= 0 {
+                _rewardedCompletionTimeout = newValue
+            } else {
+                Log.warn("The rewarded completion timeout must not be negative.")
+            }
+        }
+        get { _rewardedCompletionTimeout }
+    }
     
     /// Use to initialize video controls with server values.
     public func initialize(with ortbAdConfiguration: ORTBAdConfiguration?) {
@@ -136,4 +148,6 @@ public class VideoControlsConfiguration: NSObject {
     
     private var _skipButtonArea = PrebidConstants.BUTTON_AREA_DEFAULT.doubleValue
     private var _skipButtonPosition = Position.topLeft
+
+    private var _rewardedCompletionTimeout: TimeInterval = 120
 }
--- a/PrebidMobile/Swift/PrebidMobileRendering/AdTypes/AdView/VideoControlsConfiguration.swift
+++ b/PrebidMobile/Swift/PrebidMobileRendering/AdTypes/AdView/VideoControlsConfiguration.swift
@@ -91,18 +91,6 @@ public class VideoControlsConfiguration: NSObject {
     /// This property indicates whether mute controls is visible on the screen.
     public var isSoundButtonVisible = false
 
-    /// The fallback rewarded-ad completion timeout, in seconds, used when the bid does not provide completion criteria.
-    public var rewardedCompletionTimeout: TimeInterval {
-        set {
-            if newValue >= 0 {
-                _rewardedCompletionTimeout = newValue
-            } else {
-                Log.warn("The rewarded completion timeout must not be negative.")
-            }
-        }
-        get { _rewardedCompletionTimeout }
-    }
-    
     /// Use to initialize video controls with server values.
     public func initialize(with ortbAdConfiguration: ORTBAdConfiguration?) {
         
@@ -149,5 +137,4 @@ public class VideoControlsConfiguration: NSObject {
     private var _skipButtonArea = PrebidConstants.BUTTON_AREA_DEFAULT.doubleValue
     private var _skipButtonPosition = Position.topLeft
 
-    private var _rewardedCompletionTimeout: TimeInterval = 120
 }
--- a/PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/RewardedConfig.swift
+++ b/PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/RewardedConfig.swift
@@ -68,7 +68,7 @@ public class RewardedConfig: NSObject {
     // MARK: - Default Values
     
     /// The timeout duration for rewarded completion, measured in seconds.
-    public let defaultCompletionTime: NSNumber = 120
+    public var defaultCompletionTime: NSNumber = 120
     
     /// The playback event when the SDK should send a signal to the application that the user has earned the reward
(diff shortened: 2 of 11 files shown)


PrebidMobile/Swift/PrebidMobileRendering/AdTypes/AdView/VideoControlsConfiguration.swift
```diff
@@ -136,4 +148,6 @@ public class VideoControlsConfiguration: NSObject {
     
     private var _skipButtonArea = PrebidConstants.BUTTON_AREA_DEFAULT.doubleValue
     private var _skipButtonPosition = Position.topLeft
+
+    private var _rewardedCompletionTimeout: TimeInterval = 120
```
OlenaPostindustria: the constant is duplicated across the SDK

PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/RewardedConfig.swift
```diff
@@ -68,7 +68,7 @@ public class RewardedConfig: NSObject {
     // MARK: - Default Values
     
     /// The timeout duration for rewarded completion, measured in seconds.
-    public let defaultCompletionTime: NSNumber = 120
+    public var defaultCompletionTime: NSNumber = 120
```
OlenaPostindustria: this value was meant to stay constant, can we initialize `RewardedConfig` with the API value instead?

PrebidMobile/Swift/PrebidMobileRendering/AdTypes/AdView/VideoControlsConfiguration.swift
```diff
@@ -90,6 +90,18 @@ public class VideoControlsConfiguration: NSObject {
     
     /// This property indicates whether mute controls is visible on the screen.
     public var isSoundButtonVisible = false
+
+    /// The fallback rewarded-ad completion timeout, in seconds, used when the bid does not provide completion criteria.
+    public var rewardedCompletionTimeout: TimeInterval {
```
OlenaPostindustria: I would rename `rewardedCompletionTimeout` => `rewardedCompletionTime`. This property serves as the fallback for the time fields of the `rwdd` schema introduced in [#1056](https://github.com/prebid/prebid-mobile-ios/issues/1056) - `completion.banner.time`, `completion.video.time`, and `completion.video.endcard.time`, all of which use time rather than timeout.

OlenaPostindustria: I don't think this is the right place for this property. The completion time value could also apply to rewarded banner ads, so placing it in `VideoControlsConfiguration` may be misleading.

OlenaPostindustria (changes requested):

OlenaPostindustria:
Also, worth adding the new property to public docs

OlenaPostindustria (approved):
LGTM!

YuriyVelichkoPI (changes requested):
Blocked since there is a way to implement the requested functionality without changes in the SDK: 
https://github.com/prebid/prebid-mobile-ios/issues/1297#issuecomment-5226775282

mdanylov-sigma:
The supported rwdd passthrough and setImpORTBConfig approach is now documented in prebid/prebid.github.io#6697. I have left this SDK PR open because the latest review rejects the SDK API itself, so closing or withdrawing it requires an explicit decision.

PR #1307, opened 2026-08-04
Fix #1298: interstitial/rewarded video countdown indicator

### LINKED ISSUE #1298: Feature request: remaining-time countdown / playback progress indicator for video interstitials (parity with rewarded)

During video playback, a **rewarded** ad shows a remaining-time countdown indicator, but a **video interstitial does not display any remaining time / progress indication at all** — the user has no way to know how long the ad will run. Our QA flagged this as an inconsistency between the two full-screen formats, and it also differs from the interstitial UX of most other ad SDKs.

In the current implementation the countdown view is explicitly gated to rewarded ads only (`PBMVideoView.updateProgressBar` returns early when the ad is not rewarded), so there is no configuration a publisher can enable.

> Apparently rewarded video ads had this feature in the design, whilst Interstitial did not. In another implementation, we saw a playback progress indicator. If you think this makes sense to be consistent with either timer or playback progress bar please submit the corresponding issue into GitHub so it will be implemented ASAP.

Show a remaining-time countdown (or a playback progress indicator) during interstitial video playback, consistent with the rewarded experience — ideally togglable via `VideoControlsConfiguration` so publishers can opt in/out.

- A playback progress bar instead of a numeric countdown, as suggested by Magnite — this would also avoid the rounding/flooring display questions that the numeric timer raises (e.g. an 8.3-second creative showing "8").
- Rendering an overlay on the publisher side is not feasible: the SDK does not expose the playback position, and the interstitial is presented in the SDK's own view controller.
- Reusing the existing rewarded countdown component (`PBMCircularProgressBarView`) for interstitials looks like the lowest-effort path, since the view and update plumbing already exist.

**Additional context**

- Countdown gated to rewarded only: https://github.com/prebid/prebid-mobile-ios/blob/3.3.1/PrebidMobile/Objc/PrebidMobileRendering/AdTypes/AdView/PBMVideoView.m#L355-L362
- The same gap exists in the Prebid Android SDK (rewarded shows remaining time, interstitial does not), so a cross-platform design would be ideal.
- Environment where this was reviewed: Prebid Mobile iOS SDK 3.3.1, iPhone 16 / iOS 26.5.

mdanylov-sigma:
## Summary

Adds a remaining-time countdown indicator for full-screen video ads. Previously the circular progress bar only appeared for rewarded video; it's now shown for **any** full-screen video (rewarded or plain interstitial), gated by a new `isVideoProgressIndicatorVisible` property on `VideoControlsConfiguration` (default `true`), exposed on `InterstitialRenderingAdUnit`, `RewardedAdUnit`, and `MediationBaseInterstitialAdUnit`. Outstream video is explicitly excluded.

Closes #1298

## Behavior

- A single `shouldShowProgressBar` helper now consistently gates view creation (`updateProgressBar`), duration assignment (`initTimeObserver`), and the per-frame update (`handlePeriodicTimeEvent`) — an earlier iteration of this fix only updated the first of these three, which would have made the countdown appear but never actually count down for plain interstitials. All three are now aligned.
- View cleanup: the progress bar reference is explicitly nilled (not just removed from its superview) whenever it's torn down, avoiding a stale reference when the config is toggled off at runtime.
- Outstream video (no `isRewarded`, no `presentAsInterstitial`) never shows the indicator.

## Testing

- `PrebidMobileTests` (unit): **35/35 passed**, including new tests:
  - `testProgressIndicatorShownForInterstitialVideo`
  - `testProgressIndicatorUpdatesForInterstitialVideo` — drives a real `AVPlayer` through `initTimeObserver()`/`handlePeriodicTimeEvent()` and asserts `duration`/`value` are actually set, not just that the view exists. This is the regression test that would catch the "shown but frozen" failure mode.
  - `testProgressIndicatorCanBeHiddenForInterstitialVideo`
  - `testProgressIndicatorHiddenForOutstreamVideo`
  - `testVideoProgressIndicatorVisibility`
- **Live verification**: built and ran the demo app (`PrebidDemoSwift`, In-App Video Interstitial 320x480) in the iOS Simulator against a real bid. Confirmed the countdown indicator is visible and live (ticking down) for a plain non-rewarded interstitial video — the exact scenario this PR adds support for. (Screenshot captured locally; can attach on request — not embedded here since it isn't hosted anywhere linkable from this PR.)

## API / compatibility

- Default `true` means this turns on a new visible UI element for existing non-rewarded interstitial integrations on upgrade (not opt-in). Flagging for maintainer awareness — happy to flip the default if preferred.

### DIFF
--- a/PrebidMobile/Swift/PrebidMobileRendering/AdTypes/AdView/VideoControlsConfiguration.swift
+++ b/PrebidMobile/Swift/PrebidMobileRendering/AdTypes/AdView/VideoControlsConfiguration.swift
@@ -99,8 +99,15 @@ public class VideoControlsConfiguration: NSObject {
     /// Obtained from `ext.prebid.passthrough[].adConfiguration.isautocloseoncompletionenabled` or set by the user.
     public var isAutoCloseOnCompletionEnabled = true
 
-    /// This property indicates whether the remaining-time indicator is visible for full-screen video ads.
-    public var isVideoProgressIndicatorVisible = true
+    /// Indicates whether the remaining-time countdown indicator is visible during full-screen video playback.
+    ///
+    /// The default preserves the SDK's existing behavior: the indicator is shown for rewarded ads and hidden for
+    /// interstitials. Set this property explicitly to show it on interstitials or hide it on rewarded ads.
+    /// Obtained from `ext.prebid.passthrough[].adConfiguration.isvideoprogressindicatorvisible` or set by the user.
+    public var isVideoProgressIndicatorVisible: Bool {
+        set { _isVideoProgressIndicatorVisible = newValue as NSNumber }
+        get { _isVideoProgressIndicatorVisible?.boolValue ?? true }
+    }
     
     /// Use to initialize video controls with server values.
     public func initialize(with ortbAdConfiguration: ORTBAdConfiguration?) {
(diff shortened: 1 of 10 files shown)


OlenaPostindustria (dismissed):
Overall looks good, but worth a release notes callout, as all apps upgrading the SDK will suddenly show a countdown on interstitial video with no code change. It's worth noting that this flag also allows hiding the countdown for rewarded ads.

YuriyVelichkoPI (changes requested):
Please support new property in ORTBAdConfiguration and update the spec for Ad Experience Controls respectively 
https://docs.prebid.org/prebid-mobile/modules/rendering/combined-ad-experience-controls.html

PrebidMobile/Swift/PrebidMobileRendering/AdTypes/AdView/VideoControlsConfiguration.swift
```diff
@@ -90,6 +90,9 @@ public class VideoControlsConfiguration: NSObject {
     
     /// This property indicates whether mute controls is visible on the screen.
     public var isSoundButtonVisible = false
+
+    /// This property indicates whether the remaining-time indicator is visible for full-screen video ads.
+    public var isVideoProgressIndicatorVisible = true
```
YuriyVelichkoPI, line 95: There is no reason to change the current behaviour. 

The count down indicator should be off for regular interstitials and on for rewareded ads

YuriyVelichkoPI (changes requested):

PR #1308, opened 2026-08-04
Fix #1299: interstitial orientation lock by creative size

### LINKED ISSUE #1299: Interstitial orientation lock never engages — calculateLayoutFromSize: left as FIXME, always returns aspectRatio (portrait creatives clip in landscape)
**Describe the bug**

A portrait video interstitial (e.g. 320x480) does not lock its orientation: rotating the device to landscape is allowed, and the creative/end card is rendered clipped after rotation (no re-layout occurs either).

What makes this a clear defect rather than a missing feature: **the orientation-lock logic already exists but is never engaged, because the size-based layout determination was left unfinished with a `FIXME`.**

- `ModalViewController` locks rotation whenever the interstitial layout is determined as portrait or landscape (`shouldAutorotate` / `supportedInterfaceOrientations`), and rotation stays enabled only for `.aspectRatio`.
- `PBMInterstitialLayoutConfigurator.calculateLayoutFromSize:` is supposed to determine that layout from the creative size — but it unconditionally returns `PBMInterstitialLayoutAspectRatio`:

```objc
//FIXME: - Need to figure out how to determine orientation properly. Autorotate is enabled for now.
+ (PBMInterstitialLayout)calculateLayoutFromSize:(CGSize)size {
    return PBMInterstitialLayoutAspectRatio;
}
```

- Ironically, the `portraitSizes` / `landscapeSizes` lookup tables (which include 320x480 as a portrait size) are implemented right below this method — they are just never used.

So every interstitial runs with rotation enabled, and since rotation is not accompanied by any re-layout, portrait creatives get clipped in landscape.

**To Reproduce**

1. Load and present a 320x480 video interstitial with an end card (e.g. stored response `prebid-demo-video-interstitial-320-480-end-card-skadn-skoverlay`) in an app that allows rotation.
2. Wait for the end card (or during video playback), rotate the device to landscape.
3. The ad is not orientation-locked and the creative/end card renders clipped.

**Expected behavior**

`calculateLayoutFromSize:` should use the existing `portraitSizes` / `landscapeSizes` tables (as the surrounding code clearly intends), so that portrait creatives lock to portrait and landscape creatives to landscape. Alternatively, if free rotation is kept, the modal should re-layout the creative on orientation change so it is not clipped.

**Screenshots**

Screen recording available on request (portrait 320x480 end card clipped after rotating to landscape).

 - OS: iOS 26.5

**Additional context**

- Unfinished size-based layout determination (FIXME + unused tables): https://github.com/prebid/prebid-mobile-ios/blob/3.3.1/PrebidMobile/Objc/PrebidMobileRendering/PBMInterstitialLayoutConfigurator.m#L32-L60
- Orientation lock that would engage if the layout were determined: https://github.com/prebid/prebid-mobile-ios/blob/3.3.1/PrebidMobile/Swift/PrebidMobileRendering/AdTypes/AdView/Modals/ModalViewController.swift#L80-L91
- `isRotationEnabled` derivation (`interstitialLayout == .aspectRatio`): https://github.com/prebid/prebid-mobile-ios/blob/3.3.1/PrebidMobile/Swift/PrebidMobileRendering/AdTypes/AdView/InterstitialDisplayProperties.swift#L32-L36
- Workaround we found: `AdConfiguration.interstitialLayout` is public and honored when set explicitly, so a publisher can force `.portrait` — but the default size-based path should work without this.
- The Android SDK also shows rotation-handling issues on interstitials (controls mis-rendered after rotation), so a cross-platform review of rotation behavior may be worthwhile.

mdanylov-sigma:
## Summary

Wires up the previously dead `PBMInterstitialLayoutConfigurator.calculateLayoutFromSize:` to the existing (already-tested) `isPortrait:`/`isLandscape:` size-lookup helpers instead of unconditionally returning `.aspectRatio`. Interstitials whose creative size matches a known portrait or landscape IAB size now get their orientation locked; unrecognized/zero sizes keep the previous auto-rotate (`.aspectRatio`) fallback unchanged.

Closes #1299

## Behavior

- Portrait/landscape size lookup uses the existing hard-coded size tables (`portraitSizes`/`landscapeSizes`), unchanged by this PR.
- Zero-size and any unrecognized size fall through to `.aspectRatio`, identical to current behavior — no regression for existing integrations that don't hit a known portrait/landscape size.
- Orientation restoration on push/pop and stacked-modal interaction traced end-to-end: `ModalManager.popModal()` → `display(state:)` → `ModalViewController.setupState(_:)` recomputes `isRotationEnabled`/`interstitialLayout` from the new top-of-stack state on every push and pop. A video interstitial and its own companion/end card share the same `InterstitialDisplayProperties` instance, so the lock can't desync between them within a single ad.
- One pre-existing (not introduced by this PR) robustness note for the maintainer: neither `ModalViewController` nor `ModalManager` calls `setNeedsUpdateOfSupportedInterfaceOrientations()` after `isRotationEnabled`/`interstitialLayout` change on an already-presented `ModalViewController` instance, which iOS 16+ recommends for orientation-mask changes to reliably propagate without a fresh present/dismiss. This was rarely exercised before (auto-detected layout was always `.aspectRatio`), and is now exercised much more often since this PR locks common ad sizes by default. Recommend a device QA pass on iOS 16+ for: locked interstitial → stacked/deferred modal (e.g. in-app click-through browser) → dismiss → confirm the lock is re-enforced. Not blocking this PR — flagging for awareness/follow-up.

## Testing

- `PrebidMobileTests` (unit): **4/4 passed** in `PBMInterstitialLayoutConfiguratorTest`, including the re-enabled `testAdConfigurationNoLayoutWithSize` (portrait/landscape/unknown-size assertions) and pre-existing `testAdSizeConstants`/`testDefaultAdConfiguration`.
- **Live verification**: built and ran the demo app (`PrebidDemoSwift`, In-App Video Interstitial Landscape 320x480) in the iOS Simulator against a real bid — the landscape-sized creative renders correctly under the new size-based layout classification. (A live device-rotation test to visually confirm the lock itself wasn't completed in this pass — see robustness note above.)

## API / compatibility

- `PBMInterstitialLayoutConfigurator` is Objective-C-internal only (private header), no public API surface, no interop concerns.

### DIFF
--- a/PrebidMobile/Objc/PrebidMobileRendering/PBMInterstitialLayoutConfigurator.m
+++ b/PrebidMobile/Objc/PrebidMobileRendering/PBMInterstitialLayoutConfigurator.m
@@ -29,8 +29,15 @@
     displayProperties.interstitialLayout = [self calculateLayoutFromSize:adConfiguration.size];
 }
 
-//FIXME: - Need to figure out how to determine orientation properly. Autorotate is enabled for now.
 + (PBMInterstitialLayout)calculateLayoutFromSize:(CGSize)size {
+    if ([self isPortrait:size]) {
+        return PBMInterstitialLayoutPortrait;
+    }
+
+    if ([self isLandscape:size]) {
+        return PBMInterstitialLayoutLandscape;
+    }
+
     return PBMInterstitialLayoutAspectRatio;
 }
 
--- a/PrebidMobileTests/RenderingTests/Tests/PBMInterstitialLayoutConfiguratorTest.swift
+++ b/PrebidMobileTests/RenderingTests/Tests/PBMInterstitialLayoutConfiguratorTest.swift
@@ -85,27 +85,26 @@ class PBMInterstitialLayoutConfiguratorTest: XCTestCase {
         XCTAssertTrue(displayProperties.isRotationEnabled)
     }
     
-    // FIXME: - Auto rotation is enabled by default for now. 
-//    func testAdConfigurationNoLayoutWithSize() {
-//        let displayProperties = InterstitialDisplayProperties()
-//        let adConfig = AdConfiguration()
-//
-//        //test portrait size
-//        adConfig.size = CGSize(width: 360, height: 480)
-//        PBMInterstitialLayoutConfigurator.configureProperties(with: adConfig, displayProperties: displayProperties)
-//        XCTAssertEqual(displayProperties.interstitialLayout, .portrait)
-//        XCTAssertFalse(displayProperties.isRotationEnabled)
-//
-//        //test landscape size
-//        adConfig.size = CGSize(width: 1024, height: 768)
-//        PBMInterstitialLayoutConfigurator.configureProperties(with: adConfig, displayProperties: displayProperties)
-//        XCTAssertEqual(displayProperties.interstitialLayout, .landscape)
-//        XCTAssertFalse(displayProperties.isRotationEnabled)
-//
-//        //test aspectRatio size
-//        adConfig.size = CGSize(width: 400, height: 300)
-//        PBMInterstitialLayoutConfigurator.configureProperties(with: adConfig, displayProperties: displayProperties)
-//        XCTAssertEqual(displayProperties.interstitialLayout, .aspectRatio)
-//        XCTAssertTrue(displayProperties.isRotationEnabled)
-//    }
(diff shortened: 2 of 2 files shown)


OlenaPostindustria (approved):
LGTM

PR #1310, opened 2026-08-07
Expose VideoParameters skippable setter to Objective-C

### LINKED ISSUE #1309: VideoParameters.isSkippable is not accessible from Objective-C
Prebid Mobile iOS version: 3.3.3

The VideoParameters class contains the following Swift property:

public var isSkippable: Bool?

It also exposes:

public var rawSkippable: NSNumber? {
    get {
        guard let isSkippable else { return nil }
        return NSNumber(value: isSkippable ? 1 : 0)
    }
}

However, when using the Prebid Mobile iOS SDK from Objective-C, isSkippable is not exposed as a writable property.

For example:

VideoParameters *parameters =
    [[VideoParameters alloc] initWithMimes:@[@"video/mp4"]];

parameters.isSkippable = YES;

is not available.

The following also fails:

[parameters setSkippable:YES];

with:

No visible @interface for 'VideoParameters'
declares the selector 'setSkippable:'
Expected behavior

Since isSkippable is declared as a public writable Swift property, it should ideally be accessible from Objective-C, for example:

parameters.isSkippable = YES;

or through an Objective-C-compatible setter:

[parameters setSkippable:YES];

This is required for an Objective-C application to set the OpenRTB video skip attribute to 1.

Current workaround

There does not appear to be a public Objective-C API to set isSkippable, although rawSkippable is available as a read-only helper.

Could you please confirm whether this is an Objective-C interoperability issue and whether an Objective-C-compatible setter/property should be exposed for VideoParameters.isSkippable? Please fixed

mdanylov-sigma:
Expose an Objective-C-only `setSkippable:` API because Swift's optional `Bool` property is not writable from Objective-C. Add an Objective-C regression test covering unset, `YES`, and `NO` values.

Test: targeted `AdUnitSuccessorObjCTests` test passes.

Closes #1309

### DIFF
--- a/PrebidMobile/Swift/AdUnits/Parameters/VideoParameters.swift
+++ b/PrebidMobile/Swift/AdUnits/Parameters/VideoParameters.swift
@@ -105,6 +105,13 @@ public class VideoParameters: NSObject {
             return NSNumber(value: isSkippable ? 1 : 0)
         }
     }
+
+    /// Objective-C API for setting the OpenRTB video skip attribute.
+    @available(swift, unavailable, message: "Use isSkippable instead.")
+    @objc(setSkippable:)
+    public func setSkippable(_ isSkippable: Bool) {
+        self.isSkippable = isSkippable
+    }
     
     /// - Parameter mimes: supported MIME types
     public init(mimes: [String]) {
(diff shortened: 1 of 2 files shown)


OlenaPostindustria (approved):
LGTM

PR #1312, opened 2026-08-12
Fix: memory leak in AdUnit.baseFetchDemand timeout DispatchWorkItem

### LINKED ISSUE #1311: Memory leak in AdUnit.baseFetchDemand
**Describe the bug**
AdUnit.baseFetchDemand creates a DispatchWorkItem that strongly captures self, then schedules it with DispatchQueue.main.asyncAfter. When the bid response arrives before the timeout, timeoutHandler.cancel() is called — but cancel() on a DispatchWorkItem does not release its captured references. The work item remains alive in GCD's internal queue storage until the deadline elapses, keeping the AdUnit instance (and its entire owner chain) alive for up to timeoutMillis milliseconds after the response completes.

**To Reproduce**
1. Create an AdUnit (e.g. BannerAdUnit) and call fetchDemand
2. Receive a successful bid response before the timeout fires
3. Release all strong references to the AdUnit immediately after the completion closure fires
4. Observe in Instruments / Memory Graph that the AdUnit is still alive until the timeout deadline elapses

**Expected behavior**
After timeoutHandler.cancel() is called and all external references are released, the AdUnit should be deallocated immediately. The timeout work item should not extend the object's lifetime.

**Screenshots**
Image
Image

 - Device: iPhone/iPad
 - OS: iOS 17+
 - Browser n/a
 - Version PrebidMobile 3.3.3

**Additional context**
In the original API (V1) integration, BannerAdUnit is typically created as a local variable inside the bid-fetching function (e.g. PrebidFetcher.attachHeaderBidding). Once fetchDemand is called and the function returns, the caller holds no reference to the ad unit — the GCD timeout work item becomes the sole retainer of the object. This means the BannerAdUnit and its entire dependency graph (network connection, bid requester, etc.) are kept alive purely as a side effect of the unresolved DispatchWorkItem, for the full timeout duration. In a feed loading multiple ads simultaneously, these pile up noticeably. 
Debug mode for prebid is disabled.

mdanylov-sigma:
Closes #1311

The timeout `DispatchWorkItem` in `AdUnit.baseFetchDemand` captured `self` strongly. Cancelling it on a successful bid response doesn't release its captured references until GCD reaches the scheduled deadline, keeping the `AdUnit` (and its owner chain) alive for up to `timeoutMillis` after the response completes.

Fix: capture `self` weakly in the timeout handler, consistent with the `bidRequester.requestBids` completion closure nearby.

### DIFF
--- a/PrebidMobile/Swift/AdUnits/AdUnit.swift
+++ b/PrebidMobile/Swift/AdUnits/AdUnit.swift
@@ -169,7 +169,8 @@ public class AdUnit: NSObject, DispatcherDelegate {
         lastFetchDemandCompletion = completion
         adServerObject = adObject
         
-        let timeoutHandler = DispatchWorkItem {
+        let timeoutHandler = DispatchWorkItem { [weak self] in
+            guard let self = self else { return }
             if (!self.didReceiveResponse) {
                 self.timeOutSignalSent = true
                 completion(BidInfo(resultCode: .prebidDemandTimedOut))


OlenaPostindustria (approved):
LGTM

PR #1313, opened 2026-08-13
Add guards and review tooling

OlenaPostindustria:
# Description

This PR introduces a machine-enforced quality layer for the repository: fast, deterministic architecture checks ("guards") that run locally and in CI, a SwiftLint pass gating the style layer guards deliberately ignore, plus the review tooling and contributor documentation built around them. No SDK source code changes - the only production-adjacent edits are documentation, CI, and repo configuration.

### Structural guards (`scripts/guards/`)

- **adapter-isolation** — `EventHandlers/` may only use the public PrebidMobile API (no `__PrebidMobileInternal`, `@_spi`, PrivateHeaders, or non-public `PBM*` symbols)
- **public-api-baseline** — the public API surface is a committed, structured JSON lockfile, namespaced by exposure path (`swift` / `swift-spi` / `objc-header` / `objc-internal` / `spm-product`); any surface change requires regenerating the baseline in the same PR
- **swift-migration-direction** — no new ObjC files in the core, with an explicit-grant allowlist for rare exceptions
- **skiplist-ratchet** — the quick-plan skip-list is enumerated by test identifier, so swaps (un-skip one, skip another) fail by name, not just by count
- **ortb-test-presence** — every ORTB model (discovered by file name) must be referenced by a test
- **logging-hygiene**, **fixme-ratchet**, **deprecation-hygiene**, **api-doc-coverage**, **api-naming**, **api-test-presence**, **string-dup-ratchet** — hygiene ratchets derived from recurring upstream review asks
- **ast-rule-ratchet** — ast-grep pattern rules (main-thread delegate callbacks, force-unwraps, non-weak delegates) with per-rule baselined counts; the ast-grep version is pinned identically in CI and the runner

Failure modes are handled conservatively throughout: a failed scan is never treated as zero findings, empty-scope probes fail rather than pass, and every grandfathered allowlist entry carries a required, schema-validated `reason` field rather than a comment no parser can see.

Baselines and allowlists snapshot the tree as of the current master (the branch is rebased onto it), so the recently migrated Swift surface is grandfathered with per-entry reasons and burns down from here.

### SwiftLint (`.swiftlint.yml`, `scripts/lint/`)

Adopted as a **separate step** — not as a replacement for any guard. SwiftLint could express roughly two of the thirteen guards; the rest are cross-file, ObjC, git-diff-based, or not source at all, and SwiftLint has no equivalent of the ratchet (its baselines suppress pre-existing violations but never fail when an entry goes stale).

- `.swiftlint.yml` — replaces the dormant 2019 config, which disabled `force_cast`/`force_try`, the exact rules the force-unwrap guard exists to hold. 44 rules under `only_rules` rather than `disabled_rules`, so a version bump cannot silently introduce a rule
- **One owner per rule** — rules a guard already enforces (`force_cast`/`force_try`/`force_unwrapping`, `weak_delegate`, `todo`, `missing_docs`) are deliberately absent, so two systems can never report disagreeing counts for the same rule
- `scripts/lint/run-swiftlint.sh` — intersects findings with the lines the branch **adds** versus the merge-base with master, and **fails on them**. A whole-tree run reports ~13k pre-existing violations; that backlog is deliberately not being fixed, since reformatting untouched code would bury every real diff in the release it landed in. New code arrives clean; legacy lines stay silent until someone edits them on purpose. Blocking is the default locally too, so a local run gives the same verdict as CI
- `.github/workflows/swiftlint.yml` — ubuntu, pinned binary, blocking, with findings echoed into the job summary. Nothing is skipped silently: a missing swiftlint, an unusable diff, or a failed install exits non-zero rather than passing green
- Removes `scripts/swiftLint.sh` (installed SwiftLint 0.31.0 via `wget` + `sudo installer`; both call sites had been commented out for years)

### CI and hooks

- `.github/workflows/guards.yml` — runs guard self-tests (103 unit tests) and all guards on every PR and master push, on ubuntu in ~1 minute, separate from the macOS build pipeline so results land early and cheap
- `.github/workflows/swiftlint.yml` — the style pass, deliberately its own workflow rather than folded into guards: architecture checks and style checks fail for different reasons and should read as separate signals
- `.githooks/pre-push` — opt-in (`git config core.hooksPath .githooks`) local run of the guards before every push (`git push --no-verify` bypasses; CI remains authoritative)

### Agent instructions and review tooling (`AGENTS.md`, `.claude/`)

- `AGENTS.md` is the canonical, agent-agnostic home for the repo's rules — repo map, hard rules with their guards, test-integrity policy, the OpenRTB spec-grounding gate, environment matrix (what runs on Linux vs. needs macOS), and the guard-failure workflow (baseline `--update-*` flags, shrink-only allowlists). `CLAUDE.md` is a thin `@AGENTS.md` import, so Codex, Cursor, Copilot, and Claude Code all read the same instructions with nothing to drift
- `/sdk-review` - verdict-first PR review / full-tree audit checklist: guards first, then human-judgment items no guard can see (privacy logging, API stability, rendering/accessibility, test integrity), plus an ObjC→Swift migration checklist
- `/guard` - playbook for turning a prose architecture rule into a new guard (mechanism choice, grandfathering, true-positive/true-negative proof, registration)
- `/verify-spec` - strictly read-only audit of ORTB models/tests against the pinned IAB OpenRTB 2.x spec
- `/benchmark` - measures the download size the SDK adds to a publisher's app, and one-off wall-clock timing of hot paths (bid-request building, response parsing, creative scanning)
- `qc-validator` agent and rules docs (code patterns, testing patterns, quality gates, session completion) — plain markdown any agent can read

### Contributor documentation

- `CONTRIBUTING.md` - repo layout, the three-build-systems rule, test commands, PR conventions; AI assistance is explicitly optional, guards apply equally to all contributions
- `.github/pull_request_template.md` - checklist mirroring the guard contract
- `docs/guards/README.md` - every guard, its rationale, its fix, and how to add one
- `docs/lint/README.md` - SwiftLint: why it is diff-scoped, which rules it leaves to guards, and why `--fix` is not a shortcut
- `.gitignore` - tracks the shared `.claude/` config, keeps per-user local state and Python bytecode ignored


CONTRIBUTING.md
```diff
@@ -0,0 +1,71 @@
...
+- Unit tests (full): `./scripts/testPrebidMobile.sh --latest`
+- Adapter tests: `./scripts/testPrebidMobileAdapters.sh`
+- Demo integration tests: `./scripts/testPrebidDemo.sh -l` (add `-ui` for UI tests)
+
+CI runs the quick unit-test plan on every PR. The full suite and demo tests run when the
+PR carries the `run-full-tests` label or the branch name contains `bump-to`.
+
+## Quality guards
+
+Optionally, run the guards automatically before every commit (one-time, per clone):
```
YuriyVelichkoPI: Can it be pre-push hook, instead of pre-commit?

YuriyVelichkoPI:
@OlenaPostindustria is this PR still draft? Please take a look on merge conflicts. Initial introduction of CLAUDE md was made as "reference" to AGENTS.md  with keeping the agent's skills in more generic place accessible for other codding agents. 

How do you think will this approach work ok, or it's better to follow Caldue requrements and think about general solution later if needed. 

@mdanylov-sigma @ChrisHuie, thoughts?

PR #1319, opened 2026-08-19
Remove dead Travis setup

OlenaPostindustria:
Removes the abandoned Travis CI setup and the code-signing assets that existed only to serve it. 

The signing material here was committed in 2017 and encrypted with a Travis secret that is long gone, so the certs are expired. Deleting the files removes them from the working tree but not from git history.

If Travis is ever revived, the config and signing flow would need to be recreated from scratch anyway, since the Xcode 12 image and the secret it depended on are both retired.


mdanylov-sigma (approved):
LGTM! Thanks

PR #1321, opened 2026-08-20
Surface winning-bid economics on the original (GAM) API

yevhenii-audienzz:
## Surface winning-bid economics on the original (GAM) API

On the original (GAM) API, Prebid today exposes only the bucketed `hb_pb` keyword. This adds a way to read the winning bid's **exact** economics, without switching to the Rendering API or issuing a second auction.

### What's added (purely additive — no existing public API changed)

**`BidInfo`** — new public read-only properties, populated in `create()` from the winning bid's ORTB values:

- `cpm: NSNumber?` — exact net price. `NSNumber?` (not `Double?`) so it's visible from Objective-C and carries the price exactly as Prebid Server sent it (reading it as a `Float` first would round most decimals, e.g. `3.14` → `3.1400001…`).
- `currency: String?` — ISO-4217 for `cpm`; defaults to the ORTB default `"USD"` when the response omits `cur`.
- `creativeId: String?` — the creative id (`crid`).
- `adId: String?` — the ad id (`adid`).
- `requestId: String?` — the ORTB `BidResponse.id` (a per-request UUID the SDK generates); identifies the request, response-level.

`cpm`/`currency`/`creativeId`/`adId` describe the bid whose targeting was attached to the ad object, and are `nil` unless a bid was designated — a nil `cpm` is not a zero-price win. When `topBidFiltered` is `true` they describe the promoted cached runner-up.

**`AdUnit`** — new `fetchDemandBidInfo(adObject:completion:)`: attaches the Prebid targeting keywords into `adObject` for ad serving (like `fetchDemand(adObject:completion:)`) but returns the full `BidInfo`. `baseFetchDemand` already produced the `BidInfo`; the result-code variant now forwards to this method. A distinct method name (not a `fetchDemand` overload) avoids the trailing-closure ambiguity that would break existing call sites.

### Tests

`ResponseParsingTests` asserts, from a JSON-text fixture: an exact `3.14` `cpm` (no `Float` rounding), the `"USD"` default and an explicit `"EUR"`, distinct `crid`/`adid`/`id`, and that a parsed response with `cur` and bids but no designated winner leaves the winning-bid fields `nil` while `requestId` still reports the response id.

### DIFF
--- a/PrebidMobile/Swift/AdUnits/AdUnit.swift
+++ b/PrebidMobile/Swift/AdUnits/AdUnit.swift
@@ -132,7 +132,26 @@ public class AdUnit: NSObject, DispatcherDelegate {
             }
         }
     }
-    
+
+    /// Makes a bid request for a specific ad object and provides the full `BidInfo`. Like
+    /// `fetchDemand(adObject:completion:)` it sets up the Prebid targeting keywords into `adObject`
+    /// for ad serving, but returns the full `BidInfo` (including the winning-bid economics) instead
+    /// of only the result code.
+    ///
+    /// - Parameters:
+    ///   - adObject: The ad object for which demand is being fetched.
+    ///   - completionBidInfo: A closure called with a `BidInfo` object representing the fetched demand.
+    dynamic public func fetchDemand(
+        adObject: AnyObject,
+        completionBidInfo: @escaping (_ bidInfo: BidInfo) -> Void
+    ) {
+        baseFetchDemand(adObject: adObject) { bidInfo in
+            DispatchQueue.main.async {
+                completionBidInfo(bidInfo)
+            }
+        }
+    }
+
     // SDK internal
     func baseFetchDemand(
         adObject: AnyObject? = nil,
--- a/PrebidMobile/Swift/AdUnits/MultiformatAdUnit/BidInfo.swift
+++ b/PrebidMobile/Swift/AdUnits/MultiformatAdUnit/BidInfo.swift
@@ -40,7 +40,27 @@ public class BidInfo: NSObject {
     
     /// Events related to the bid
     public private(set) var events: [String: String]
-    
+
+    // MARK: - Winning-bid economics
+    // Exact economics of the winning bid, surfaced on the original (GAM) API. Prebid normally
+    // exposes only the bucketed `hb_pb` keyword; these read the winning bid's real values off the
+    // (module-internal) ORTB response inside `create(...)`.
+
+    /// Winning bid net price (exact CPM), if any.
+    public private(set) var cpm: Double?
+
+    /// Bid currency (ISO-4217), from the ORTB bid response.
+    public private(set) var currency: String?
+
+    /// Winning creative id (`crid`).
+    public private(set) var creativeId: String?
+
+    /// Auction id — the ORTB response id (falls back to the bidder response id `bidid`).
+    public private(set) var auctionId: String?
+
+    /// Winning ad id (`adid`).
+    public private(set) var adId: String?
+
     /// Initializes a new `BidInfo` instance with the specified parameters.
     /// - Parameters:
     ///   - resultCode: The result code of the bid request.
@@ -86,7 +106,17 @@ public class BidInfo: NSObject {
         if let impURL = bidResponse.winningBid?.events?.win {
             bidInfo.addEvent(key: BidInfo.EVENT_IMP, value: impURL)
         }
-        
+
+        // Surface exact economics of the winning bid (original-API analytics). These read the
+        // module-internal ORTB objects, which are only reachable here inside PrebidMobile.
+        if let winningBid = bidResponse.winningBid {
+            bidInfo.cpm = Double(winningBid.price)
+            bidInfo.creativeId = winningBid.bid.crid
+            bidInfo.adId = winningBid.bid.adid
+        }
+        bidInfo.currency = bidResponse.rawResponse?.cur
+        bidInfo.auctionId = bidResponse.rawResponse?.requestID ?? bidResponse.rawResponse?.bidid
+
         return bidInfo
     }
     
--- a/PrebidMobileTests/ResponseParsingTests.swift
+++ b/PrebidMobileTests/ResponseParsingTests.swift
@@ -513,4 +513,38 @@ class ResponseParsingTests: XCTestCase {
         XCTAssertTrue(passthrough.sdkConfiguration is CustomSDKConfiguration)
         XCTAssertTrue(bidExt.skadn!.fidelities![0] is CustomSkadnFidelity)
     }
+
+    // MARK: - BidInfo winning-bid economics (original/GAM API)
+
+    // `BidInfo.create` should surface the winning bid's exact economics (cpm/currency/creativeId/
+    // adId/auctionId) alongside the targeting keywords, so integrators on the original API can read
+    // them without switching to the Rendering API or issuing a second auction.
+    func testBidInfoSurfacesWinningBidEconomics() {
+        let bidResponse = WinningBidResponseFabricator.makeWinningBidResponse(bidPrice: 3.14)
+        guard let winningBid = bidResponse.winningBid else {
+            return XCTFail("Fabricated response should have a winning bid")
+        }
+
+        let bidInfo = BidInfo.create(resultCode: .prebidDemandFetchSuccess, bidResponse: bidResponse)
+
+        XCTAssertEqual(bidInfo.resultCode, .prebidDemandFetchSuccess)
+        XCTAssertEqual(bidInfo.cpm, Double(winningBid.price))
+        XCTAssertEqual(bidInfo.currency, bidResponse.rawResponse?.cur)
+        XCTAssertEqual(bidInfo.creativeId, winningBid.bid.crid)
+        XCTAssertEqual(bidInfo.adId, winningBid.bid.adid)
+        XCTAssertEqual(bidInfo.auctionId,
+                       bidResponse.rawResponse?.requestID ?? bidResponse.rawResponse?.bidid)
+    }
+
+    // With no winning bid the economics must stay nil (e.g. a no-bid response).
+    func testBidInfoEconomicsAreNilWithoutWinningBid() {
+        let bidResponse = BidResponse(jsonDictionary: [:])
+        XCTAssertNil(bidResponse.winningBid)
+
+        let bidInfo = BidInfo.create(resultCode: .prebidDemandNoBids, bidResponse: bidResponse)
+
+        XCTAssertNil(bidInfo.cpm)
+        XCTAssertNil(bidInfo.creativeId)
+        XCTAssertNil(bidInfo.adId)
+    }
 }


PrebidMobile/Swift/AdUnits/MultiformatAdUnit/BidInfo.swift
```diff
@@ -86,7 +106,17 @@ public class BidInfo: NSObject {
         if let impURL = bidResponse.winningBid?.events?.win {
             bidInfo.addEvent(key: BidInfo.EVENT_IMP, value: impURL)
         }
-        
+
+        // Surface exact economics of the winning bid (original-API analytics). These read the
+        // module-internal ORTB objects, which are only reachable here inside PrebidMobile.
+        if let winningBid = bidResponse.winningBid {
+            bidInfo.cpm = Double(winningBid.price)
```
OlenaPostindustria, line 113: Nit: `winningBid.bid.price?.doubleValue` would avoid widening an already-rounded Float here

PrebidMobile/Swift/AdUnits/MultiformatAdUnit/BidInfo.swift
```diff
@@ -40,7 +40,27 @@ public class BidInfo: NSObject {
...
     public private(set) var events: [String: String]
-    
+
+    // MARK: - Winning-bid economics
+    // Exact economics of the winning bid, surfaced on the original (GAM) API. Prebid normally
+    // exposes only the bucketed `hb_pb` keyword; these read the winning bid's real values off the
+    // (module-internal) ORTB response inside `create(...)`.
+
+    /// Winning bid net price (exact CPM), if any.
+    public private(set) var cpm: Double?
```
OlenaPostindustria, line 50: `Double?` is not visible from Obj-C code

PrebidMobile/Swift/AdUnits/AdUnit.swift
```diff
@@ -132,7 +132,26 @@ public class AdUnit: NSObject, DispatcherDelegate {
...
+
+    /// Makes a bid request for a specific ad object and provides the full `BidInfo`. Like
+    /// `fetchDemand(adObject:completion:)` it sets up the Prebid targeting keywords into `adObject`
+    /// for ad serving, but returns the full `BidInfo` (including the winning-bid economics) instead
+    /// of only the result code.
+    ///
+    /// - Parameters:
+    ///   - adObject: The ad object for which demand is being fetched.
+    ///   - completionBidInfo: A closure called with a `BidInfo` object representing the fetched demand.
+    dynamic public func fetchDemand(
```
OlenaPostindustria, line 144: This overload is source-breaking. With a trailing closure Swift ignores the label, so `fetchDemand(adObject:completion:)` and `fetchDemand(adObject:completionBidInfo:)` are only told apart by the closure's parameter type. Any call whose body doesn't pin that type now fails with ambiguous use of `fetchDemand`. You can observe this in the `PrebidDemoSwift` - `GAMOriginalAPINativeViewController.swift:73` no longer compiles, and `PrebidOriginalAPINativeController.swift:123` in `InternalTestApp` has the same shape.

OlenaPostindustria (changes requested):

PrebidMobile/Swift/AdUnits/MultiformatAdUnit/BidInfo.swift
```diff
@@ -86,7 +106,17 @@ public class BidInfo: NSObject {
...
+
+        // Surface exact economics of the winning bid (original-API analytics). These read the
+        // module-internal ORTB objects, which are only reachable here inside PrebidMobile.
+        if let winningBid = bidResponse.winningBid {
+            bidInfo.cpm = Double(winningBid.price)
+            bidInfo.creativeId = winningBid.bid.crid
+            bidInfo.adId = winningBid.bid.adid
+        }
+        bidInfo.currency = bidResponse.rawResponse?.cur
+        bidInfo.auctionId = bidResponse.rawResponse?.requestID ?? bidResponse.rawResponse?.bidid
```
mdanylov-sigma, line 118: `auctionId` ends up holding the request id, and the `bidid` fallback can't run.

**What the value is.** `rawResponse.requestID` is the ORTB `BidResponse.id`, which echoes the id of the bid request. The SDK generates that id itself, as a new UUID for every request:

https://github.com/prebid/prebid-mobile-ios/blob/d7d7b4eca96c0a470233de39987d2a1d28a2e9ac/PrebidMobile/Objc/PrebidMobileRendering/Prebid/PBMCore/PBMPrebidParameterBuilder.m#L72

So it identifies a single SDK request rather than an auction, and `requestId` would describe it more accurately.

**The fallback is unreachable.** `ORTBBidResponse` fails to initialize when `id` is missing, so whenever `rawResponse` is non-nil, `requestID` is non-nil too:

https://github.com/prebid/prebid-mobile-ios/blob/d7d7b4eca96c0a470233de39987d2a1d28a2e9ac/PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCore/ORTB/ORTBBidResponse.swift#L75-L81

**It would change the meaning if it could run.** The field would silently switch to `bidid`, which ORTB defines as a bidder-generated response tracking id, not a request or auction id.

Suggest renaming the property to `requestId` (or documenting exactly what it holds), dropping the fallback, and updating the doc comment on line 58, which currently describes the fallback.

PrebidMobile/Swift/AdUnits/MultiformatAdUnit/BidInfo.swift
```diff
@@ -86,7 +106,17 @@ public class BidInfo: NSObject {
...
-        
+
+        // Surface exact economics of the winning bid (original-API analytics). These read the
+        // module-internal ORTB objects, which are only reachable here inside PrebidMobile.
+        if let winningBid = bidResponse.winningBid {
+            bidInfo.cpm = Double(winningBid.price)
+            bidInfo.creativeId = winningBid.bid.crid
+            bidInfo.adId = winningBid.bid.adid
+        }
+        bidInfo.currency = bidResponse.rawResponse?.cur
```
mdanylov-sigma, line 117: `currency` and `auctionId` are assigned outside the `if let winningBid` block, while the section comment describes all five properties as winning-bid economics. A response that parses but has no winning bid still gets an `auctionId` (and a `currency` when the response includes `cur`), while `cpm`, `creativeId` and `adId` stay nil.

That case is reachable with the default configuration:

- `Targeting.shared.forceSdkToChooseWinner` defaults to `false` ([Targeting.swift#L119](https://github.com/prebid/prebid-mobile-ios/blob/d7d7b4eca96c0a470233de39987d2a1d28a2e9ac/PrebidMobile/Swift/ConfigurationAndTargeting/Targeting.swift#L118-L119)).
- With that setting, `setUp` attaches the keywords to the ad object and returns `.prebidDemandFetchSuccess` whenever the response has bids but no winning bid ([AdUnit.swift#L254-L267](https://github.com/prebid/prebid-mobile-ios/blob/8439f39f83541fc825fc4e01e25b6d187d4dc25b/PrebidMobile/Swift/AdUnits/AdUnit.swift#L254-L267)).
- A bid only counts as winning when its targeting has `hb_pb`, `hb_bidder` and `hb_cache_id` ([Bid.swift#L110-L127](https://github.com/prebid/prebid-mobile-ios/blob/d7d7b4eca96c0a470233de39987d2a1d28a2e9ac/PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCore/Bid.swift#L110-L127)). `hb_cache_id` is required because `AdUnit.init` turns on `useCacheForReportingWithRenderingAPI` ([AdUnit.swift#L88](https://github.com/prebid/prebid-mobile-ios/blob/8439f39f83541fc825fc4e01e25b6d187d4dc25b/PrebidMobile/Swift/AdUnits/AdUnit.swift#L87-L88)), so a top bid without a cache id is enough to hit this path.

The integrator then gets `.prebidDemandFetchSuccess`, keywords attached to the ad object, and `cpm == nil`.

Suggest either moving both assignments into the `if let winningBid` block, or documenting them as response-level values. Either way, the property docs should say the winning-bid fields are nil unless Prebid Server designated a winning bid, so integrators don't record a nil `cpm` as zero revenue.

PrebidMobile/Swift/AdUnits/AdUnit.swift
```diff
@@ -132,7 +132,26 @@ public class AdUnit: NSObject, DispatcherDelegate {
             }
         }
     }
-    
+
+    /// Makes a bid request for a specific ad object and provides the full `BidInfo`. Like
```
mdanylov-sigma, line 136: `PrebidAdUnit` already offers this call on the original API:

https://github.com/prebid/prebid-mobile-ios/blob/d7d7b4eca96c0a470233de39987d2a1d28a2e9ac/PrebidMobile/Swift/AdUnits/MultiformatAdUnit/PrebidAdUnit.swift#L39-L54

It attaches targeting to the ad object and returns a `BidInfo`, going through the same `AdUnit.baseFetchDemand` and `BidInfo.create` path as this method. So the new `BidInfo` properties already reach `PrebidAdUnit` users without any new method.

This overload only adds value for the `AdUnit` subclasses (`BannerAdUnit`, `InterstitialAdUnit`, `RewardedVideoAdUnit`, `InstreamVideoAdUnit`, `NativeRequest`). Worth deciding whether that justifies another public `fetchDemand` variant. If it does, it needs a distinct method name rather than a distinct parameter label, because of the trailing-closure ambiguity already raised on line 144.

PrebidMobile/Swift/AdUnits/MultiformatAdUnit/BidInfo.swift
```diff
@@ -40,7 +40,27 @@ public class BidInfo: NSObject {
...
+    // MARK: - Winning-bid economics
+    // Exact economics of the winning bid, surfaced on the original (GAM) API. Prebid normally
+    // exposes only the bucketed `hb_pb` keyword; these read the winning bid's real values off the
+    // (module-internal) ORTB response inside `create(...)`.
+
+    /// Winning bid net price (exact CPM), if any.
+    public private(set) var cpm: Double?
+
+    /// Bid currency (ISO-4217), from the ORTB bid response.
+    public private(set) var currency: String?
```
mdanylov-sigma, line 53: Minor: ORTB makes `cur` optional with a default of `"USD"`:

https://github.com/prebid/prebid-mobile-ios/blob/d7d7b4eca96c0a470233de39987d2a1d28a2e9ac/PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCore/ORTB/ORTBBidResponse.swift#L45-L47

As written, a response without `cur` produces a non-nil `cpm` with `currency == nil`. Consider falling back to `"USD"` when there is a winning bid, or documenting that nil means the ORTB default, so consumers don't need to know the spec to interpret the value.

PrebidMobile/Swift/AdUnits/AdUnit.swift
mdanylov-sigma, line 144: Minor: `dynamic` isn't needed on this method. On the existing `fetchDemand` methods it's there so tests can exchange implementations (the `// TODO: dynamic is used by tests` above):

https://github.com/prebid/prebid-mobile-ios/blob/d7d7b4eca96c0a470233de39987d2a1d28a2e9ac/PrebidMobileTests/AdUnitSwizzleHelper.swift#L29-L33

Nothing exchanges this selector, so `public func` is enough.

PrebidMobile/Swift/AdUnits/AdUnit.swift
```diff
@@ -132,7 +132,26 @@ public class AdUnit: NSObject, DispatcherDelegate {
...
+    ///   - completionBidInfo: A closure called with a `BidInfo` object representing the fetched demand.
+    dynamic public func fetchDemand(
+        adObject: AnyObject,
+        completionBidInfo: @escaping (_ bidInfo: BidInfo) -> Void
+    ) {
+        baseFetchDemand(adObject: adObject) { bidInfo in
+            DispatchQueue.main.async {
+                completionBidInfo(bidInfo)
+            }
+        }
```
mdanylov-sigma, line 152: Minor: this body is the same as `fetchDemand(adObject:completion:)` apart from the callback it invokes. Once the naming question is settled, the result-code variant can forward to this one and pass `bidInfo.resultCode` to its completion, so the main-queue dispatch lives in one place.

PrebidMobile/Swift/AdUnits/MultiformatAdUnit/BidInfo.swift
```diff
@@ -40,7 +40,27 @@ public class BidInfo: NSObject {
...
+    /// Winning creative id (`crid`).
+    public private(set) var creativeId: String?
+
+    /// Auction id — the ORTB response id (falls back to the bidder response id `bidid`).
+    public private(set) var auctionId: String?
+
+    /// Winning ad id (`adid`).
+    public private(set) var adId: String?
+
     /// Initializes a new `BidInfo` instance with the specified parameters.
```
mdanylov-sigma, line 64: Minor: the new properties are `private(set)` and this initializer doesn't accept them, so only `BidInfo.create` can populate them. Integrators who unit-test their analytics or revenue code against `BidInfo` can't build one with a `cpm`. Consider adding them here as defaulted parameters, like the existing optional ones.

mdanylov-sigma (commented):
A few implementation notes inline, on top of the existing review comments.

PrebidMobile/Swift/AdUnits/MultiformatAdUnit/BidInfo.swift
mdanylov-sigma: Agree, and I'd treat this as a correctness issue rather than a nit, since an exact CPM is the point of this PR. There are two separate problems on this line.

**1. Precision is lost before the conversion.** `Bid.price` isn't the raw value, it's a computed `Float`:

https://github.com/prebid/prebid-mobile-ios/blob/d7d7b4eca96c0a470233de39987d2a1d28a2e9ac/PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCore/Bid.swift#L31-L33

The raw ORTB `price` (an `NSNumber?` parsed from the response JSON) is narrowed to a 32-bit `Float` first. `Double(...)` then widens the already rounded value, which can't bring the lost digits back. Most decimal prices have no exact `Float` representation, so the error lands directly in the value integrators will log. Parsing prices from JSON into an `NSNumber` and reading them both ways:

| `price` in the response | `Double(winningBid.price)` (this PR) | `winningBid.bid.price?.doubleValue` |
|---|---|---|
| `3.14` | `3.140000104904175` | `3.14` |
| `0.1` | `0.10000000149011612` | `0.1` |
| `12.345` | `12.345000267028809` | `12.345` |

**2. A missing price becomes `0.0` instead of nil.** The getter falls back with `?? 0.0`, so a winning bid without `price` reports `cpm == 0.0`. The property is documented as "if any", and in analytics a `0.0` reads as a real zero-price win rather than an unknown value.

Reading the raw value fixes both:

```swift
bidInfo.cpm = winningBid.bid.price?.doubleValue
```

`Double` is still binary floating point, but it round-trips the value Prebid Server sent, which is what integrators will reconcile against. If `cpm` becomes `NSNumber?` for Objective-C visibility (see the comment on line 50), `winningBid.bid.price` can be passed through as is.

mdanylov-sigma (commented):
Thanks for the update. I checked all ten threads against c7948ad6.

**Resolved**

- Float precision and missing price: `cpm` now passes the raw `bid.price` through. A price of `3.14` parsed from JSON text comes back as exactly `3.14`.
- Obj-C visibility: the generated header now has `NSNumber * _Nullable cpm`.
- Source-breaking overload: with the distinct name, an untyped trailing closure such as `nativeUnit.fetchDemand(adObject: gamRequest) { resultCode in ... }` type-checks again.
- `auctionId` is renamed to `requestId`, the `bidid` fallback is gone, and the docs are updated.
- `currency` moved inside the winning-bid block, `requestId` is documented as response-level, and the docs warn that a nil `cpm` is not a zero-price win. A parsed response with `cur` and bids but no designated winner now gives nil `cpm`/`currency`/`creativeId`/`adId` and a non-nil `requestId`.
- `"USD"` default, `dynamic` removed, the result-code variant forwards to the new method, and the init takes the new fields as defaulted parameters.

**Still open**

1. **Conflict with master.** #1327 added `topBidFiltered` to the same `BidInfo` init. The merge itself is mechanical: I kept both and appended the new parameters after `topBidFiltered`, and `ResponseParsingTests`, `AdUnitTests` and `PrebidAdUnitTests` pass on the result.

   The docs need one change after the rebase, though. With `Prebid.shared.filterOutUncachedBids`, `removeBidsWithoutSuccessfulCache()` promotes a cached runner-up into `winningBid` when the bid Prebid Server picked has no cache entry. `cpm`, `creativeId` and `adId` then describe the promoted bid. I checked this: `topBidFiltered == true`, `cpm == 1.25`, and `creativeId` is the runner-up's `crid`. Those are the right values, because the promoted bid's targeting is what gets attached to the ad object. But "`nil` unless Prebid Server designated a winning bid" is no longer accurate. Suggest describing them as the bid whose targeting was attached, and pointing to `topBidFiltered`.

2. **The new tests don't lock the fixes.**
   - `testBidInfoEconomicsAreNilWithoutWinningBid` builds `BidResponse(jsonDictionary: [:])`. `rawResponse` is nil there, so the previous version of `create()` passes it too. The case from the thread is a parsed response with `cur` and bids but no winner, for example a bid whose targeting has `hb_pb` and `hb_cache_id` but no `hb_bidder`. It should assert `winningBid == nil`, `targetingInfo != nil`, nil `cpm` and `currency`, and `requestId` equal to the response `id`.
   - In `testBidInfoSurfacesWinningBidEconomics`, the fabricator sets no `crid` or `adid` and uses an empty `id`. The `creativeId`/`adId` assertions compare nil with nil, and the `currency`/`requestId` assertions repeat the implementation's own expression. Swapping `crid` and `adid` in `create()` would still pass. Suggest a fixture with distinct `crid`, `adid` and `id` values, a literal `"USD"` expectation, and a second case with `"cur": "EUR"`.
   - Parsing the fixture from JSON text rather than an in-memory dictionary would cover the exact-price path end to end.

3. **The PR description is out of date.** It still describes a `fetchDemand(adObject:completionBidInfo:)` overload and `auctionId`.

4. **The `PrebidAdUnit` overlap thread is only half answered.** It also asked whether a new public method on `AdUnit` is worth adding, since `PrebidAdUnit.fetchDemand(adObject:request:completion:)` already attaches targeting and returns a `BidInfo`. The rename fixes the technical problem. A reply there with the use case (for example, existing `BannerAdUnit`/`NativeRequest` integrations that can't move to `PrebidAdUnit`) would help the maintainers decide.

Nit: the paragraph starting "Named `fetchDemandBidInfo` rather than adding a `fetchDemand` overload on purpose" ships in the public API docs and the generated Obj-C header. That rationale fits better in the commit message or a `//` comment.

PrebidMobile/Swift/AdUnits/AdUnit.swift
yevhenii-audienzz: Kept the new method for the classic original-API units (`BannerAdUnit`, `InterstitialAdUnit`, `RewardedVideoAdUnit`, `InstreamVideoAdUnit`, `NativeRequest`) — those integrations use `AdUnit` directly and would need a real migration to move to `PrebidAdUnit`. This gives them the economics without a second auction.

PrebidMobile/Swift/AdUnits/MultiformatAdUnit/BidInfo.swift
yevhenii-audienzz: Done — falls back to "USD" when the response omits cur.

PrebidMobile/Swift/AdUnits/AdUnit.swift
yevhenii-audienzz: Dropped, it's public func now.

PrebidMobile/Swift/AdUnits/MultiformatAdUnit/BidInfo.swift
yevhenii-audienzz: Done

PR #1325, opened 2026-08-26
fix: timeoutMillisDynamic treated as seconds instead of milliseconds (#1323)

### LINKED ISSUE #1323: timeoutMillis doesn't work
**Describe the bug**  
`Prebid.shared.timeoutMillis` is configured in **milliseconds**, but `PBMBidRequester` appears to handle `timeoutMillisDynamic` as **seconds** when passing the value as an `NSTimeInterval`.

In `PBMBidRequester.m`:

```objc
const NSInteger rawTimeoutMS_onRead = self.sdkConfiguration.timeoutMillis;
NSNumber * const dynamicTimeout_onRead = self.sdkConfiguration.timeoutMillisDynamic;

const NSTimeInterval postTimeout =
    (dynamicTimeout_onRead
        ? dynamicTimeout_onRead.doubleValue
        : (rawTimeoutMS_onRead / 1000.0));
```

The fallback path correctly converts milliseconds to seconds:

```objc
rawTimeoutMS_onRead / 1000.0
```

However, the dynamic timeout is passed directly:

```objc
dynamicTimeout_onRead.doubleValue
```

Since `NSTimeInterval` represents seconds, a configured timeout of `2000` milliseconds can become `2000` seconds instead of `2` seconds.

**To Reproduce**  

1. Configure the Prebid timeout:

```swift
Prebid.shared.timeoutMillis = 2000
```

2. Make a Prebid request using `fetchDemand`.
3. Use a Prebid Server endpoint that takes longer than 2 seconds to respond, or simulate a slow network response.
4. Observe that the network request does not time out after approximately 2 seconds.
5. Inspect `PBMBidRequester` and observe that `timeoutMillisDynamic.doubleValue` is passed to the network layer without converting milliseconds to seconds.

For example:

```text
Configured:
timeoutMillis = 2000 ms

Expected NSTimeInterval:
2000 / 1000 = 2 seconds

Actual dynamic value:
timeoutMillisDynamic.doubleValue = 2000

Effective NSTimeInterval:
2000 seconds
```

**Expected behavior**  
`timeoutMillisDynamic` should be converted from milliseconds to seconds before being used as an `NSTimeInterval`.

For example:

```objc
const NSTimeInterval postTimeout =
    (dynamicTimeout_onRead
        ? dynamicTimeout_onRead.doubleValue / 1000.0
        : rawTimeoutMS_onRead / 1000.0);
```

With:

```swift
Prebid.shared.timeoutMillis = 2000
```

the network timeout should be approximately **2 seconds**, not `2000` seconds.

**Screenshots**  
N/A

- OS: macOS
- Browser: N/A
- Version: N/A

- Device: iPhone
- OS: iOS
- Browser: N/A
- Version: Prebid Mobile iOS version: 3.3.4

**Additional context**  
`timeoutMillis` is explicitly defined as a millisecond value. Updating it also updates `timeoutMillisDynamic`:

```swift
public var timeoutMillis: Int {
    didSet {
        timeoutMillisDynamic = NSNumber(value: timeoutMillis)
    }
}
```

Therefore, `timeoutMillisDynamic` contains the same millisecond value as `timeoutMillis`.

The two branches in `PBMBidRequester` currently treat the same timeout value using different units:

```text
timeoutMillis = 2000

rawTimeoutMS_onRead:
2000 / 1000.0 = 2 seconds       ✅

timeoutMillisDynamic:
2000.doubleValue = 2000 seconds ❌
```

The dynamic path should also divide the value by `1000.0` before passing it as an `NSTimeInterval`.

This can cause Prebid requests to remain active significantly longer than the configured timeout.

mdanylov-sigma:
## Summary

Fixes #1323.

`Prebid.shared.timeoutMillis` is documented and used as a **milliseconds** value everywhere — including `timeoutMillisDynamic`, which mirrors it (`Prebid.swift`'s `didSet`) and is read as milliseconds in `AdUnit.swift`'s timeout scheduling. `PBMBidRequester.m`, however, had two places where that unit contract was broken:

1. **Read path (the reported bug):** `dynamicTimeout_onRead.doubleValue` was passed directly into an `NSTimeInterval` (seconds) without dividing by 1000. A configured `timeoutMillis = 2000` became a 2000 second network timeout instead of 2 seconds.
2. **Write path (a related bug in the same unit contract):** when the server's `tmaxrequest` adaptively updates `timeoutMillisDynamic`, the already-in-seconds computed value was stored back *without* converting to milliseconds. So once an adaptive timeout kicked in, the next request's read-path fix would divide it by 1000 again, producing a timeout **1000x too small**.

## Fix

Both sites in `PBMBidRequester.m` now consistently treat `timeoutMillisDynamic` as milliseconds:
- Divide by 1000 when reading it into the `NSTimeInterval` passed to the network layer.
- Multiply by 1000 when writing the adaptively-computed (seconds-based) timeout back into it.

## Tests

Added two regression tests in `PBMBidRequesterTest.swift`:
- `testTimeoutMillisDynamic_ConvertedFromMillisecondsToSecondsForRequest` — a 500ms dynamic timeout now produces a 0.5s `NSTimeInterval` on the network call.
- `testAdaptiveTimeout_StoredInMillisecondsAfterBidResponse` — the adaptive timeout computed from `tmaxrequest` is stored back in milliseconds, not seconds.

Ran the full `PBMBidRequesterTest` suite (iPhone 17, iOS 26.5 simulator): 16/16 passed, including both new tests.


YuriyVelichkoPI (approved):
LGTM

PR #1265, opened 2026-05-15
PS bid caching failures

mdanylov-sigma:
#1258

PR #1327, opened 2026-08-27
PS bid caching failures v2: align with reviewed design doc

### LINKED ISSUE #1258: Handle Prebid Server bid caching failures gracefully
This issue is related to a potential blank ads situation. One of the most common reasons for blank ads is due to Prebid Server not having enough time to cache the bid or due to hard cache failures. SDK doesn't handle this in any way, and from Prebid Server perspective, a failure to cache the bid is not a basis to fail the auction itself. Because the bid itself is still present in the response.

Bid caching failure may happen due to multiple reasons. One of which is pretty organic: not enough time left after the actual auction. This feature https://github.com/prebid/prebid-server/issues/2620 should help mitigate this issue to some degree, but not entirely negate it, as the issue itself is rooted in how much time is given to run the auction (tmax) versus how quickly the bidders will actually respond. If bidders will be taking the entire time budget, then there might not be enough time left after waiting for all the bidder responses to cache the bids.

This is a proposal to add a toggle into an SDK that will allow the SDK to ignore the bids that weren't cached. Potentially ignoring the entire action if all bids in the response weren't cached successfully. This should allow publishers to add logic around it and potentially either retry the auction or skip the ad slot refresh.

When the bid caching is successful, Prebid Server populates `response.seatbid[].bid[].ext.prebid.cache.bids.bids/vastXml` objects. This can be used as a signal in cases when server-side bid caching is required.

One alternative might be caching bids asynchronously, but that creates a risk of losing feedback in case of hard cache failures.

Another alternative might be implementing a local bid caching mechanism similar to Prebid.js.

PS. This applies to both iOS and Android SDK.

mdanylov-sigma:
Follow-up to #1265, addressing the shared design doc review for #1258 (Yuriy Velichko). Android parity: prebid/prebid-mobile-android#943.

## Changes on top of #1265

- Renamed `Prebid.shared.requireServerSideBidCache` to `filterOutUncachedBids` (default `false`) and gave it a single responsibility: it only filters which bids reach the ad server, it never requests caching on its own.
- Scoped the response filtering and its result codes to Original API only. Rendering API renders creatives directly from the bid's own markup and never depends on Prebid Cache, so a cache failure there is not a demand failure.
- Dropped the flag from the request builder's cache-request trigger. Original API already requests caching via `useCacheForReportingWithRenderingAPI`, so the cache is always requested wherever the filter can apply.
- Fixed a bug: when the PBS-designated winning bid was filtered out for failing the cache check, no other bid was promoted to winner, so the SDK discarded all remaining cached demand and returned `noWinningBid`. The highest-priced cached bid is now promoted instead.
- A promoted winner is still a success: the result code is `.prebidDemandFetchSuccess` and the new `BidInfo.topBidFiltered` flag is `true`, so publishers can track the yield impact. It is deliberately not a separate result code, because integrations that gate on `resultCode == .prebidDemandFetchSuccess` would silently drop valid demand.
- When every returned bid is filtered out, fetch demand returns `.prebidDemandNoCachedBids`. A response without a winner for any other reason still returns the existing no winning bid error.
- Tightened the cache-success predicate to require `cacheId` instead of accepting `url` or `cacheId`. Prebid Cache is read with `GET /cache?uuid=<cacheId>`, and Prebid Universal Creative builds that request from the `hb_cache_id` targeting key. `url` is just `cacheId` joined with the cache host and path, so it never counts on its own.
  - https://docs.prebid.org/prebid-server/endpoints/pbs-endpoints-pbc.html
  - https://docs.prebid.org/prebid-server/use-cases/pbs-sdk.html
- Added the "GAM (Original API) Display Banner 320x50 [Filter Uncached Bids]" example (`GAMOriginalAPIFilterUncachedBidsBannerViewController`) showing how to handle both outcomes.

## Usage

```swift
Prebid.shared.filterOutUncachedBids = true

adUnit.fetchDemand(adObject: gamRequest, request: prebidRequest) { bidInfo in
    switch bidInfo.resultCode {
    case .prebidDemandFetchSuccess:
        if bidInfo.topBidFiltered {
            // The top bid failed to cache, a lower-priced cached bid was promoted
        }
    case .prebidDemandNoCachedBids:
        // No returned bid was cached, no Prebid targeting is attached
        break
    default:
        break
    }
    gamBanner.load(gamRequest)
}
```

## Notes

- A promoted bid carries only bidder-suffixed keys (for example `hb_pb_appnexus`), since the unsuffixed keys belonged to the filtered winner. Line items that target only the unsuffixed `hb_pb` will not match it. Android behaves the same way.
- The removed-bid count is logged only. Automatic retry is deferred, and enabling the flag by default in 4.0 is still an open question.

## Testing

- `PBMBidResponseTransformerTest`: winner promotion, `url` without `cacheId`, `vastXml` cache entries.
- `PBMBidRequesterTest`: Original API only scoping (Rendering API bids are not filtered), `noCachedBids` vs `noWinningBid`.
- `AdUnitTests`: a promoted winner reports fetch success, and `BidInfo` carries `topBidFiltered`.
- `PrebidParameterBuilderTest`: updated for the flag no longer triggering the cache request.

Rebased onto current master (Swift migration phases 3 and 4). PR test plan green locally after the rebase: 863/863.

Closes #1258.


YuriyVelichkoPI (changes requested):
It's hard to verify from the diff alone: how is a publisher actually supposed to consume the two new result codes without it silently breaking their existing `fetchDemand` handling? 

Right now the PR adds `prebidDemandNoCachedBids` and `prebidDemandTopBidFiltered` but doesn't show a single call site — in the SDK itself or in the demo app — that treats them correctly. That makes this feel unfinished from an integration standpoint, even though the core logic is right.

 The concrete problem I'd consider a blocker: `AdUnit.setUp` returns `.prebidDemandTopBidFiltered` instead of `.prebidDemandFetchSuccess` when the top bid gets filtered and a runner-up is promoted. But the demand was successfully returned — targeting is attached to the ad object either way. Our own `GAMOriginalAPIVideoInstreamViewController` example gates the actual ad request on `resultCode == .prebidDemandFetchSuccess`, which is the pattern most publishers copy. With this change, that exact pattern will drop perfectly good, cached demand on the floor instead of just tracking a yield signal, which is the opposite of what the design doc intends. I think this needs to be fixed before merge — either by having `prebidDemandTopBidFiltered` also satisfy a "did I get an ad" check, or by clearly documenting that publishers must treat it as a success variant, with the SDK's own examples updated to prove it.

 On top of that fix, could we add (or point to, if I'm missing something) integration examples in the demo app that actually exercise both new codes end to end? Right now there's nothing in PrebidDemo that enables `filterOutUncachedBids` or handles either
  `prebidDemandNoCachedBids` or `prebidDemandTopBidFiltered` in a completion handler. Since this is genuinely new publisher-facing behavior, I'd like to see at least one Original API example (banner or interstitial is fine) that turns the setting on and shows the recommended handling for both outcomes — that's the best way to make sure the "correct" integration  pattern actually exists somewhere before we ask publishers to adopt it. And we'll use these snippets in docs

mdanylov-sigma:
Thanks, the blocker was real and I agree with the reasoning. Fixed in 30fa23db.

**The result code.** `AdUnit.setUp` now returns `.prebidDemandFetchSuccess` whenever a winning bid exists, promoted or not. You're right that the demand is successfully returned in that case: the bid is cached and its targeting is attached to the ad object, so gating it behind a different code meant the common `resultCode == .prebidDemandFetchSuccess` pattern would drop exactly the demand the promotion logic exists to save. This also lines up with the convention already used a few lines above, where the `forceSdkToChooseWinner` path returns `prebidDemandFetchSuccess` once targeting is attached.

The yield signal moved to **`BidInfo.topBidFiltered`**. I went with this over documenting the code as a "success variant" because any success variant on `ResultCode` stays a silent-drop hazard for every integration that doesn't get updated, and docs don't reach code that's already written. Keeping "did I get an ad" and "was there yield impact" on separate channels removes the footgun by construction instead of asking publishers to work around it. Consequently I removed `prebidDemandTopBidFiltered` rather than leave an enum case nothing returns. `prebidDemandNoCachedBids` stays as-is, since that one is a genuine no-demand failure and gating on it is correct.

Worth noting for the doc: this contradicts the Outcome Matrix row that specifies a `topBidFiltered` result code, so that row needs updating to describe a success plus a `BidInfo` flag. Happy to make that edit if you agree with the shape.

**Demo example.** Added `GAMOriginalAPIFilterUncachedBidsBannerViewController` (registered in the integration case list as "GAM (Original API) Display Banner 320x50 [Filter Uncached Bids]"). It turns `filterOutUncachedBids` on and handles both outcomes explicitly: the success path, including logging when `topBidFiltered` is set, and `prebidDemandNoCachedBids`. It loads the GAM request in both cases so the no-cached-demand path falls through to the ad server rather than dropping the request.

One thing that surfaced while writing it: `BannerAdUnit.fetchDemand(adObject:completion:)` only reports a `ResultCode`, so it structurally cannot expose the new flag. The example therefore uses `PrebidAdUnit.fetchDemand(adObject:request:completion:)`, which already returns a full `BidInfo` while still attaching targeting to the ad object. I did try adding a `fetchDemand(adObject:completionBidInfo:)` overload to `AdUnit`, but it makes existing trailing-closure call sites ambiguous (it broke `GAMOriginalAPINativeViewController` locally, and would break publisher code the same way), so I dropped it. If we want the legacy `AdUnit` API to expose this too, it needs a distinct method name rather than an overload; let me know if you'd like that in this PR or a follow-up.

Verified: full unit suite 1244/1244, and `PrebidDemoSwift` builds.

YuriyVelichkoPI (dismissed):
LGTM

Note (feedback to the last comment): yes, the current variant looks more robust and reflects the response state in a more accurate way. Publisher won't get implicit API change, and will be able to process the new response state respectively

Feel free to update the doc. Thanks!

YuriyVelichkoPI (dismissed):
It's already approved but requires rebasing

YuriyVelichkoPI (approved):
LGTM

PR #1328, opened 2026-08-27
Swift Migration phase 3: networking parameter builders

YuriyVelichkoPI:
Ports `PrebidMobile/Objc/PrebidMobileRendering/Networking/Parameters/` to Swift — migration steps
**S3.1** and **S3.2**. Continues #1280 (phases 0–2).

Full write-up, including every deliberate divergence, is in
[`docs/migration/pr-phase-3.md`](https://github.com/prebid/prebid-mobile-ios/blob/swift-migration-phase-3/docs/migration/pr-phase-3.md).

## Summary

- **S3.1** — 8 parameter builders + `ORTBParameterBuilder`, the `ParameterBuilder` protocol, and
  the two support types the builders were the only consumers of (`BundleProtocol`,
  `InternalUserConsentDataManager`).
- **S3.2** — `ParameterBuilderService`. Landed with S3.1 because `PBMParameterBuilderService.m` is
  the only non-test consumer of seven of the eight builders; a commit porting only the builders
  would not compile.
- 12 new Swift files, 23 ObjC files deleted, 3 consumers re-pointed.

### Scope — what this PR does *not* cover

`Networking/Parameters/` is now empty of `.m` files, but two ObjC files remain elsewhere under
`Networking/`: `Tracking/PBMTrackingRecord` and `URLBuilding/PBMURLComponents`. Whether they are an
S3.3 or belong to a later phase is not decided here — **please confirm the phase boundary.**

### Deliberate divergences from the ObjC originals

1. `InternalUserConsentDataManager.gppSID` is a non-optional `[NSNumber]` built with `compactMap`.
   The ObjC version passed `numberFromString:`'s result to `addObject:`, which throws
   `NSInvalidArgumentException` on a malformed `IABGPP_GppSID`. Same treatment for
   `SKAdNetworksParameterBuilder.skAdNetworkIds()`.
2. `ORTBParameterBuilder.buildOpenRTB(for:)` returns a non-optional dictionary — the ObjC signature
   was nullable but no path returned `nil`.
3. `PBMAssert` nil-guards dropped in the six builders whose init parameters are non-optional in
   Swift. `BasicParameterBuilder` keeps its optional properties and its "Invalid properties" log,
   because three tests assert on it.
4. Optional `NSMutableDictionary` writes go through a new `pbmSetValue(_:forKey:)`: ObjC
   `dict[key] = nil` removes the key, but the Swift subscript stores a boxed `Optional.none`.
5. `PBMPrebidParameterBuilder.h` gains an explicit `NS_SWIFT_NAME(build(_:))` — an ObjC class
   conforming to a Swift `@objc` protocol does not inherit the protocol's Swift method name.

### Playbook

- 8 new gaps (`S3.1-A` … `S3.1-G`, `S3.2-A`).
- **Gap 4 / Gap 6 corrected.** Both claimed all ObjC consumers of the ORTB Swift twins are gone
  after Phase 3, so they could be demoted from `@objc public` to `internal`. False —
  `PBMPrebidParameterBuilder.m`, `PBMBidRequester.m`, `PBMBidResponseTransformer.m` and
  `PBMWebView.m` still consume them. Demotion stays deferred to S9.2.
- **Orphan-header inventory** added: the 39 headers under `PrebidMobile/Objc/` with no matching
  `.m`. The per-class porting recipe never applies to them; each is retired when its last importer
  is ported. Two are already dead with zero importers anywhere — `PBMAdLoadFlowController.h` (its
  Swift twin already ships) and `PBMORTB_NotImplemented.h` — left in place to keep this diff scoped.

## Test plan

- [x] `./scripts/buildPrebidMobile.sh` — 4 XCFrameworks clean
- [x] `./scripts/buildPrebidMobilePackage.sh` — SwiftPM clean
- [x] `./scripts/testPrebidMobile.sh --latest --quick` — **811 tests, 0 failures, no retries**.
      This PR adds and removes no test methods; the count differs from the 765 in `pr-phase-012.md`
      because of the three upstream PRs merged since b21cbc2b (#1301, #1305, #1325).
- [x] `./scripts/testPrebidMobile.sh --latest` — full suite, before merge
- [ ] Reviewer: confirm the phase boundary (see Scope)
- [ ] Reviewer: confirm the five deliberate divergences
- [ ] Reviewer: confirm the orphan-header inventory against the phase plan


docs/migration/playbook.md
```diff
@@ -638,6 +646,230 @@ The counts above are a snapshot, not an enforced budget: nothing in CI gates the
...
+### Gap S3.1-A — ObjC `dict[key] = nil` removes the key; the Swift subscript boxes `Optional.none`
+
+`PBMBasicParameterBuilder.m` and `PBMUserConsentParameterBuilder.m` write nullable values straight
+into an `NSMutableDictionary`:
+
+```objc
+bidRequest.regs.ext[@"gdpr"] = self.targeting.getSubjectToGDPR;   // nil ⇒ key removed
+```
+
+The literal Swift translation `dict["gdpr"] = value` where `value` is `T?` **stores a boxed
```
mdanylov-sigma: Swift 5 does not behave as described here. Assigning a simple `String?` or `NSNumber?` whose value is `nil` to `NSMutableDictionary` (and `[String: Any]`) removes the key; optional chaining also leaves a single-level optional. I verified this with `swiftc -swift-version 5`. The helper is harmless, but please correct this playbook rule and the matching comment in `NSMutableDictionary+PBMExtensions.swift` / PR doc so future migrations are not based on a nonexistent wire-format bug.

docs/migration/pr-phase-3.md
```diff
@@ -0,0 +1,160 @@
...
+1. **`InternalUserConsentDataManager.gppSID` is non-optional `[NSNumber]` and uses `compactMap`.**
+   The ObjC version built an `NSMutableArray` and called `addObject:` with the result of
+   `numberFromString:`, which returns `nil` for an unparseable component — an
+   `NSInvalidArgumentException` crash on malformed `IABGPP_GppSID`. The Swift version skips the
+   bad component. `InternalUserConsentDataManagerTests.assertIABGPPSID` was retyped accordingly.
+2. **`SKAdNetworksParameterBuilder.skAdNetworkIds()` uses `compactMap`** over the
+   `SKAdNetworkItems` plist array, likewise skipping malformed entries instead of inserting `nil`.
+3. **`ORTBParameterBuilder.buildOpenRTB(for:)` returns a non-optional `[String: String]`.**
+   The ObjC signature was nullable but every path returned a dictionary; the error path returns
+   `["openrtb": ""]` as before, after logging "Not valid JSON object".
```
mdanylov-sigma: The implementation initializes `result` as `[:]` and inserts `openrtb` only when serialization succeeds; the Objective-C original did the same. Please change this to say the error path returns an empty dictionary. It would also be useful for `testAppendBuilderParametersWitError` to capture the result and assert `isEmpty`.

PrebidMobile/Swift/ConfigurationAndTargeting/InternalUserConsentDataManager.swift
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    static var gppSID: [NSNumber] {
+        guard let gppSID = UserDefaults.standard.string(forKey: InternalUserConsentDataManager.IABGPP_GppSID),
+              !gppSID.isEmpty else {
+            return []
+        }
+
+        let numberFormatter = NumberFormatter()
+        return gppSID.components(separatedBy: "_").compactMap { numberFormatter.number(from: $0) }
```
mdanylov-sigma: This intentional crash-to-skip behavior needs a regression test. Existing coverage only exercises `2_3_4_5`, and `testIABGPPSID_Unset` currently calls `assertIABGPPString(nil)`, so it does not test `gppSID` at all. Please add a malformed input case such as `2_bad_5 -> [2, 5]` and fix the unset test.

PrebidMobile/Swift/PrebidMobileRendering/Networking/Parameters/SKAdNetworksParameterBuilder.swift
```diff
@@ -0,0 +1,68 @@
...
+        guard #available(iOS 14.0, *) else {
+            return nil
+        }
+
+        let itemsKey = SKAdNetworksParameterBuilder.SKAdNetworkItemsKey
+        guard let skadNetworks = bundle.infoDictionary?[itemsKey] as? [[String: Any]] else {
+            return nil
+        }
+
+        return skadNetworks.compactMap { $0[SKAdNetworksParameterBuilder.SKAdNetworkIdentifierKey] as? String }
```
mdanylov-sigma, line 66: Please add a test that exercises the real `skAdNetworkIds()` parsing with a bundle containing a missing or wrong-type identifier and verifies that only valid IDs are retained. The current test subclasses the builder and overrides this method, so this deliberate behavioral divergence is not covered.

YuriyVelichkoPI:
All four comments addressed in a5c970fa — thanks, you were right about the Swift 5 subscript behavior, so Gap S3.1-A is withdrawn and the helper is gone. Also fixed the error-path wording and added the missing tests for the gppSID and skAdNetworkIds skip behavior.

YuriyVelichkoPI:
Local test run: 
image

YuriyVelichkoPI:
fixing the tests on CI

PrebidMobile/Swift/ConfigurationAndTargeting/InternalUserConsentDataManager.swift
```diff
@@ -0,0 +1,50 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+import Foundation
+
+@objcMembers
```
mdanylov-sigma: `@objcMembers` looks unnecessary here — a repo-wide grep for `InternalUserConsentDataManager` across `*.h`/`*.m`/`*.mm` turns up zero remaining Objective-C importers now that this PR deletes the old `.h`/`.m` and the bridging-header reference. The only consumers are `BasicParameterBuilder.swift`, `UserConsentParameterBuilder.swift`, and a `@testable` test. Marking it `@objcMembers` unnecessarily exposes every member through the generated `-Swift.h` header for a type nothing in ObjC calls. `ParameterBuilderService` already gets this right (scoped to its one real ObjC consumer) — might be worth the same per-type check here.

PrebidMobile/Swift/PrebidMobileRendering/Networking/Parameters/DeviceInfoParameterBuilder.swift
```diff
@@ -0,0 +1,77 @@
...
+import Foundation
+
+class DeviceInfoParameterBuilder: NSObject, ParameterBuilder {
+
+    static var ifaKey: String { "ifa" }
+    static var lmtKey: String { "lmt" }
+    static var ifvKey: String { "ifv" }
+    static var attsKey: String { "atts" }
+
+    private static let zeroedIFA = "00000000-0000-0000-0000-000000000000"
```
mdanylov-sigma: `zeroedIFA` re-literals the all-zero IDFA sentinel string. `Constants.swift:24` already defines `String.kIFASentinelValue` with the identical value — worth reusing that instead of carrying two sources of truth for the same magic string.

PrebidMobile/Swift/ConfigurationAndTargeting/InternalUserConsentDataManager.swift
```diff
@@ -0,0 +1,50 @@
...
+        UserDefaults.standard.string(forKey: InternalUserConsentDataManager.IABGPP_HDR_GppString)
+    }
+
+    static var gppSID: [NSNumber] {
+        guard let gppSID = UserDefaults.standard.string(forKey: InternalUserConsentDataManager.IABGPP_GppSID),
+              !gppSID.isEmpty else {
+            return []
+        }
+
+        let numberFormatter = NumberFormatter()
```
mdanylov-sigma: This parses GPP SID digit components with a fresh `NumberFormatter()`, which is locale-sensitive (reads grouping/decimal separators from `Locale.current`). `String+Extensions.swift:57` already has `strictNumberValue` (`Int64(self).map { NSNumber(value: $0) }`) purpose-built for parsing spec-defined digit strings without locale sensitivity, and it's already reused for this exact class of ORTB field in `SkadnParametersManager.swift`. Might be worth switching to that here for consistency.

PrebidMobile/Swift/PrebidMobileRendering/Networking/Parameters/BasicParameterBuilder.swift
```diff
@@ -0,0 +1,117 @@
...
+            rtbImp.displaymanagerver = adConfiguration.isOriginalAPI ? nil : sdkVersion
+
+            rtbImp.instl = adConfiguration.presentAsInterstitial ? 1 : 0
+
+            // set secure=1 for https or secure=0 for http
+            rtbImp.secure = 1
+        }
+
+        bidRequest.regs.coppa = targeting?.coppa
+        bidRequest.regs.ext?["gdpr"] = targeting?.getSubjectToGDPR()
```
mdanylov-sigma: `regs.ext["gdpr"]` gets set here from `targeting?.getSubjectToGDPR()`, and then set again in `UserConsentParameterBuilder.swift:24` from `Targeting.shared.getSubjectToGDPR()`. Since `UserConsentParameterBuilder` runs after `BasicParameterBuilder` in `ParameterBuilderService`'s builder array, this write (and its `getSubjectToGDPR()` call) is always discarded — looks like carried-over dead work from the ObjC originals that this rewrite was a natural point to dedupe. `UserConsentParameterBuilder`'s own doc comment already names it the intended owner of GDPR/CCPA signals.

mdanylov-sigma (commented):
Automated review pass over the parameter-builder Swift port. 9 comments below; I also checked 4 other candidates against your latest "Address review of #1328" commit and dropped them since it already addresses them (pbmSetValue removal, added compactMap test coverage, doc correction on `buildOpenRTB`'s error path).

PrebidMobile/Swift/PrebidMobileRendering/Networking/Parameters/BasicParameterBuilder.swift
```diff
@@ -0,0 +1,117 @@
...
+    static var allowRedirectsVal: String { "true" }
+    static var sdkVersionKey: String { "sv" }
+    static var urlKey: String { PrebidConstants.APP_STORE_URL_SCHEME }
+    static var rewardedVideoKey: String { "vrw" }
+    static var rewardedVideoValue: String { "1" }
+
+    // Note: properties below are optional for UnitTests to be able to write 'nil' into them.
+    // TODO: Prove that 'init' arguments are never nil; convert to 'let'; remove redundant checks and tests.
+
+    var adConfiguration: AdConfiguration?
```
mdanylov-sigma: These 4 properties stay mutable optionals (with the guard/log below) only so `testInvalidProperties` can inject `nil` post-construction — the only production call site (`ParameterBuilderService.swift`) always passes non-optional values, same as all 7 sibling builders which already use non-optional `let`s. Since this file is being rewritten from scratch here, might be a good point to update the test instead (construct with the ObjC-nil-equivalent rather than mutating a live var) and resolve the TODO on line 30, keeping all 8 builders symmetric.

PrebidMobile/Swift/PrebidMobileRendering/Networking/Parameters/DeviceInfoParameterBuilder.swift
```diff
@@ -0,0 +1,77 @@
...
+
+    private let deviceAccessManager: DeviceAccessManager
+
```
mdanylov-sigma, line 22: This and 4 sibling builders (`GeoLocationParameterBuilder`, `AppInfoParameterBuilder`, `NetworkParameterBuilder`, `SKAdNetworksParameterBuilder`) dropped the ObjC `PBMAssert` nil-guard now that init params are non-optional `let`s. That's safe for every current call site (all Swift, type-checked at compile time) — flagging only because the old ObjC guard logged-and-continued in Release rather than crashing, so if any future ObjC entry point (an EventHandler/adapter, or an `@objc`-annotated construction path) ever crosses a `nil` into one of these initializers, it now hard-traps in every build configuration instead of degrading gracefully. Probably fine as-is given today's call sites, just worth a note in the playbook as a latent behavior-class change.

PrebidMobile/Swift/PrebidMobileRendering/Networking/Parameters/AppInfoParameterBuilder.swift
```diff
@@ -0,0 +1,53 @@
...
+        super.init()
+    }
+
+    func build(_ bidRequest: ORTBBidRequest) {
+        if bidRequest.app.bundle == nil, let bundleIdentifier = bundle.bundleIdentifier {
+            bidRequest.app.bundle = bundleIdentifier
+        }
+
+        if let bundleDict = bundle.infoDictionary {
+            let bundleDisplayName = bundleDict[AppInfoParameterBuilder.bundleDisplayNameKey] as? String
```
mdanylov-sigma, line 39: The bundle display-name/name lookup now goes through `as? String`, which isn't in the PR's list of 5 disclosed deliberate divergences. The ObjC original did an untyped `bundleDict[key]` assignment; if `CFBundleDisplayName` ever resolved to a non-string value (e.g. a broken `$(VAR)` build-setting substitution), ObjC would propagate the wrong-typed value while Swift now silently discards it and falls through to `bundleName`. Probably a safe/desirable behavior change, but might be worth adding to the divergence list in `pr-phase-3.md` since it wasn't reviewed as one.

PrebidMobile/Swift/PrebidMobileRendering/Networking/Parameters/BasicParameterBuilder.swift
```diff
@@ -0,0 +1,117 @@
...
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+import Foundation
+
+class BasicParameterBuilder: NSObject, ParameterBuilder {
+
+    static var platformKey: String { "sp" }
```
mdanylov-sigma: These 8 properties (plus `ifaKey`/`lmtKey`/`ifvKey`/`attsKey` in `DeviceInfoParameterBuilder.swift:21-24`) are mechanically ported from the ObjC headers but aren't referenced anywhere else in the codebase (repo-wide grep comes back empty for all 12). The ObjC header for `PBMDeviceInfoParameterBuilder` even carried a `//TODO: move this properties to extensions for tests` comment flagging them as vestigial — the TODO didn't make it into the port, and neither did a cleanup. Might be worth deleting them if nothing external actually needs the string constants.

PrebidMobile/Swift/PrebidMobileRendering/Networking/Parameters/GeoLocationParameterBuilder.swift
```diff
@@ -0,0 +1,47 @@
...
+        guard Prebid.shared.shareGeoLocation else {
+            return
+        }
+
+        guard locationManager.coordinatesAreValid else {
+            return
+        }
+
+        // Rounds with the precision defined in Targeting, or returns the original coordinates if precision is nil.
+        let coordinates = Utils.shared.round(
```
mdanylov-sigma: This coordinate-rounding-then-`NSNumber`-wrapping block (`Utils.shared.round(coordinates:precision:)` + `Targeting.shared.locationPrecision`) is duplicated in `ParameterBuilderService.swift:146` for `user.geo` — same call shape, different target (`device.geo` here). A future rounding/precision fix would need to land in both places by hand. Might be worth a small shared helper, e.g. `applyRounded(_:precision:to:)`, called from both sites.

PrebidMobile/Swift/ConfigurationAndTargeting/InternalUserConsentDataManager.swift
YuriyVelichkoPI: Done in f9f8d574 — and I went one step further than the comment.

Confirmed first: `rg -l "InternalUserConsentDataManager" -g "*.h" -g "*.m" -g "*.mm"` is empty. The only consumers are `BasicParameterBuilder`, `UserConsentParameterBuilder` and a `@testable` test.

Dropping just `@objcMembers` would not have achieved the stated goal, though: an *internal* `NSObject` subclass is still emitted as an `@interface` into the module's generated header, so the type would have stayed in `-Swift.h` regardless. So the `NSObject` base is gone too and it is now a plain `final class`.

Generalised into playbook Gap S3.2-A as a grep to run before reaching for `@objc` — empty output ⇒ no `@objc`, no `@objcMembers`, and no `NSObject` base unless the type needs one for another reason.

PrebidMobile/Swift/PrebidMobileRendering/Networking/Parameters/DeviceInfoParameterBuilder.swift
YuriyVelichkoPI: Correct — `String.kIFASentinelValue` at `PrebidMobile/Swift/Constants.swift:24` already holds the identical literal. Local copy deleted in f9f8d574; the guard now reads:

```swift
if ifa == nil || ifa == String.kIFASentinelValue {
```

PrebidMobile/Swift/ConfigurationAndTargeting/InternalUserConsentDataManager.swift
YuriyVelichkoPI: Agreed, and `strictNumberValue` exists for exactly this (`String+Extensions.swift:57` — `Int64(self).map { NSNumber(value: $0) }`, with a doc comment about locale independence). Switched in f9f8d574:

```swift
// `strictNumberValue` rather than `NumberFormatter`: the section IDs are a spec-defined run
// of digits and must resolve identically regardless of `Locale.current`'s separators.
return gppSID.components(separatedBy: "_").compactMap { $0.strictNumberValue }
```

Worth being explicit that this is a *narrowing* relative to both the ObjC original and the first Swift draft, not just a like-for-like swap: `NumberFormatter` would also have accepted decimals and signs the GPP spec does not define. No test changed — every existing input is a plain integer string — and `testIABGPPSID_Malformed` still covers the skip path. Recorded on the divergence list in `pr-phase-3.md` as an amendment to entry 1.

PrebidMobile/Swift/PrebidMobileRendering/Networking/Parameters/BasicParameterBuilder.swift
YuriyVelichkoPI: Right for the production pipeline, and removed in f9f8d574 — but the removal turned up one case where it was *not* dead, which is worth recording.

`ParameterBuilderService` runs Basic → GeoLocation → AppInfo → DeviceInfo → Network → **UserConsent** → SKAdNetworks, and `UserConsentParameterBuilder` writes `regs.ext["gdpr"]` unconditionally, so through that path the Basic write is always overwritten. Confirmed.

`PrebidParameterBuilderTest.buildBidRequest(with:)`, however, assembles its own partial pipeline — Basic + DeviceInfo + `PBMPrebidParameterBuilder`, no `UserConsentParameterBuilder` — and was silently relying on the duplicate write. `testSubjectToGDPR` failed the moment I deleted it:

```
XCTAssertEqual failed: ("nil") is not equal to ("Optional(1)")
```

The test was right about the expected value and wrong about how it got there. Fixed by adding `UserConsentParameterBuilder()` to that helper in its production-relative position, which is what it should have been doing anyway — the helper now models the real ordering rather than depending on a duplicate.

Production line replaced with a comment naming the owner so the next reader does not re-add it:

```swift
bidRequest.regs.coppa = targeting.coppa
// `regs.ext["gdpr"]` is owned by `UserConsentParameterBuilder`, which runs after this one.
```

133 tests across the nine parameter-builder / consent / `Utils` classes, 0 failures.

PrebidMobile/Swift/PrebidMobileRendering/Networking/Parameters/BasicParameterBuilder.swift
YuriyVelichkoPI: Agreed, and done in f9f8d574. The four `var`s are now `private let`, the initializer parameters are non-optional, the `Log.error("Invalid properties")` guard is gone, and `testInvalidProperties` went with them — it could only reach that guard by nilling the properties through the ObjC test-extension header this PR deletes, so it was already testing an unreachable path. The test class's now-unused `logToFile` lock is gone too.

Playbook Gap S3.1-C rewritten to prescribe this rather than the `TODO`: deleting the ObjC test-extension header means deleting the looseness it existed to serve, not preserving it in Swift.

**One thing this surfaced, which I have flagged rather than taken.** `sdkConfiguration` is now an unused stored property. It was already unused in substance in ObjC — `git show master:…/PBMBasicParameterBuilder.m` shows `_sdkConfiguration` read only by the `PBMAssert` and the `Invalid properties` guard, both of which this round removes. I stopped short of deleting it because the initializer has 13 call sites across four test files, and dropping it would also strand the `sdkConfiguration:` parameter of `ParameterBuilderService.buildParamsDict(with:bundle:…)` — the seam `ParameterBuilderServiceTest` injects through. Happy to do it as a follow-up commit on this branch if you want it; just say.

PrebidMobile/Swift/PrebidMobileRendering/Networking/Parameters/DeviceInfoParameterBuilder.swift
YuriyVelichkoPI: Agreed — this is the sharpest of the nine, because it is a behaviour *class* change and not a behaviour change: `PBMAssert` is compiled out in Release, so an ObjC caller passing `nil` logged and continued, whereas a non-optional Swift parameter traps unconditionally in every configuration.

Recorded as **playbook Gap S3.1-H** in f9f8d574, with:

- the mechanism spelled out (Release-mode log-and-continue → Release-mode trap);
- why it is safe *today*: all eight builders are constructed only by `ParameterBuilderService`, in Swift, with non-optional arguments — so no `nil` can reach them and no publisher app can trigger the trap;
- the grep to re-run before assuming that still holds;
- and the rule if it stops holding: re-introduce an explicit `Log.error` + early return rather than accept the trap, because an SDK must not crash a publisher app over a bad argument.

Divergence 4 in `pr-phase-3.md` rewritten to point at it, and it now covers all the builders rather than carving `BasicParameterBuilder` out as the exception.

PrebidMobile/Swift/PrebidMobileRendering/Networking/Parameters/AppInfoParameterBuilder.swift
YuriyVelichkoPI: Fair — it was in the diff but not on the list, which is the thing the list exists to prevent. Added as **divergence 6** in `pr-phase-3.md` (f9f8d574), with the ObjC behaviour it replaces spelled out.

For the record, the original was an unchecked static cast:

```objc
NSString *bundleDisplayName = bundleDict[PBMAppInfoParameterBuilder.bundleDisplayNameKey];
```

so a plist whose `CFBundleDisplayName` is not a string handed that object straight through to `bidRequest.app.name` (declared `NSString *`), surfacing later as an unrecognised selector or a mistyped JSON value. `as? String` yields `nil` for the same input, so a mistyped `CFBundleDisplayName` now **falls back to `CFBundleName`**, and a mistyped pair leaves `app.name` unset.

Not optional in Swift — the only way to reproduce the ObjC behaviour is an unsafe cast that would trap at first use — but a real change, and it belongs on the list.

PrebidMobile/Swift/PrebidMobileRendering/Networking/Parameters/BasicParameterBuilder.swift
YuriyVelichkoPI: Verified unused repo-wide and deleted in f9f8d574 — your count of 12 was right:

- `BasicParameterBuilder` (8): `platformKey`, `platformValue`, `allowRedirectsKey`, `allowRedirectsVal`, `sdkVersionKey`, `urlKey`, `rewardedVideoKey`, `rewardedVideoValue`
- `DeviceInfoParameterBuilder` (4): `ifaKey`, `lmtKey`, `ifvKey`, `attsKey`

One correction to the scope: `AppInfoParameterBuilder.bundleNameKey` / `bundleDisplayNameKey` are **kept** — `MockBundle.swift:42,47` reads both, so they are load-bearing for the tests rather than dead. Same reason `SKAdNetworkItemsKey` / `SKAdNetworkIdentifierKey` lost their `private` in review round 1.

PrebidMobile/Swift/PrebidMobileRendering/Networking/Parameters/GeoLocationParameterBuilder.swift
YuriyVelichkoPI: Done in f9f8d574 — extracted as an `ORTBGeo` extension, so both call sites are now one line and the rounding rule lives in one place:

```swift
extension ORTBGeo {
    func setRoundedCoordinates(_ coordinates: CLLocationCoordinate2D, precision: NSNumber?) {
        let rounded = Utils.shared.round(coordinates: coordinates, precision: precision)
        lat = NSNumber(value: rounded.latitude)
        lon = NSNumber(value: rounded.longitude)
    }
}
```

Used by `GeoLocationParameterBuilder` for `device.geo` and by `ParameterBuilderService.createORTBBidRequest(with:)` for `user.geo`.

On placement: I put it in `GeoLocationParameterBuilder.swift` rather than `ORTBGeo.swift`, to keep the ORTB twins pure data now that they are being ported wholesale, and rather than a new file, to avoid a `project.pbxproj` edit for six lines. Happy to move it if you would rather it sat with the model.

YuriyVelichkoPI:
## Review round 2 — all nine comments addressed

Pushed as f9f8d574 (code + docs) and 670019bd (verification results). All nine accepted; seven are code changes, two are documentation. Individual replies are on each thread — the short version:

| # | Comment | Change |
|---|---------|--------|
| 1 | `@objcMembers` unnecessary | Dropped, **and** the `NSObject` base — an internal `NSObject` subclass is still emitted into `-Swift.h`, so removing only the attribute would not have achieved the goal. Now `final class`. |
| 2 | `zeroedIFA` duplicates `String.kIFASentinelValue` | Local copy deleted. |
| 3 | `NumberFormatter` is locale-sensitive | → `String.strictNumberValue`. |
| 4 | `regs.ext["gdpr"]` write is dead | Removed. **See below** — it was dead in production but not in one test pipeline. |
| 5 | Mutable optionals exist only for `testInvalidProperties` | Four `var`s → `private let`, guard gone, `testInvalidProperties` and its `logToFile` lock deleted. |
| 6 | Dropping `PBMAssert` is a behaviour-class change | Recorded as playbook **Gap S3.1-H**. |
| 7 | `as? String` is an undisclosed divergence | Added as **divergence 6**. |
| 8 | 12 unused static constants | Deleted. `AppInfoParameterBuilder`'s two are kept — `MockBundle` reads them. |
| 9 | Duplicated coordinate rounding | Extracted `ORTBGeo.setRoundedCoordinates(_:precision:)`. |

### Two things worth pulling out of the threads

**Comment 4 was right about production and wrong about the tests, which is the more interesting half.** `UserConsentParameterBuilder` does overwrite `regs.ext["gdpr"]` unconditionally, so the `BasicParameterBuilder` write is dead through `ParameterBuilderService`. But `PrebidParameterBuilderTest.buildBidRequest(with:)` assembles a *partial* pipeline that omits `UserConsentParameterBuilder`, and was silently depending on the duplicate — `testSubjectToGDPR` failed the moment the write went away. The helper now runs `UserConsentParameterBuilder` in its production-relative position, which is what it should have been doing regardless. Worth knowing that this test file's pipeline is hand-rolled and can drift from the real one.

**One finding I have flagged rather than taken.** `BasicParameterBuilder.sdkConfiguration` is now an unused stored property — it was already unused in substance in ObjC, read only by the `PBMAssert` and the `Invalid properties` guard, both of which this round deletes. I stopped short of removing it because the initializer has 13 call sites across four test files and dropping it would also strand the `sdkConfiguration:` parameter of `ParameterBuilderService.buildParamsDict(with:bundle:…)`, the seam `ParameterBuilderServiceTest` injects through. That is wider than the comment asked for. **Your call** — happy to do it as a follow-up commit on this branch.

### Verification

- `./scripts/testPrebidMobile.sh --latest --quick` — **812 tests, 0 failures, no retries** (813 − 1: `testInvalidProperties` deleted, nothing else moved)
- `./scripts/buildPrebidMobile.sh` — all 4 XCFrameworks clean
- `swiftlint --config .swiftlint.yml` — no new violations from this round
- CI on f9f8d574: **green**. Also confirms the `SetupTests` user-agent seeding from 8ef9d554 held — the WebKit cold-start flake that failed [run 33369608203](https://github.com/prebid/prebid-mobile-ios/actions/runs/33369608203) has not recurred across two runs since.

`docs/migration/pr-phase-3.md` has the full round-2 table, the amended divergence list (1, 4 and the new 6), and the updated test plan. `docs/migration/playbook.md` has the rewritten Gap S3.1-C, the new Gap S3.1-H, and the extended Gap S3.2-A.

PrebidMobile/Swift/PrebidMobileRendering/Networking/Parameters/UserConsentParameterBuilder.swift
```diff
@@ -0,0 +1,30 @@
...
+import Foundation
+
+/// `UserConsentParameterBuilder` is responsible for enriching its provided bid request
+/// with the GDPR and CCPA consent signals collected from `NSUserDefaults`.
+class UserConsentParameterBuilder: NSObject, ParameterBuilder {
+
+    func build(_ bidRequest: ORTBBidRequest) {
+        // GDPR
+        bidRequest.regs.ext?["gdpr"] = Targeting.shared.getSubjectToGDPR()
+        bidRequest.user.ext?["consent"] = Targeting.shared.gdprConsentString
```
mdanylov-sigma, line 25: `PBMPrebidParameterBuilder.m:115-116` (Objective-C, not touched by this diff so I can't anchor a comment there directly) independently writes `user.ext["consent"]` from `self.targeting.gdprConsentString`, duplicating this line's write of the same key from `Targeting.shared.gdprConsentString` — the same class of ordering-dependent duplicate-write bug just fixed for `regs.ext["gdpr"]` in `BasicParameterBuilder`. Currently benign since `PBMPrebidParameterBuilder` always runs last (appended via `extraParameterBuilders`) and every production call site injects `Targeting.shared` into both, so the two reads agree. But `PBMBidRequester` constructor-injects its `targeting` (`PBMBidRequester.m:40`) rather than hardcoding `Targeting.shared` — nothing stops a future caller from passing a different instance, at which point the two would silently disagree and `PBMPrebidParameterBuilder`'s write would win unconditionally, with no comment on either side naming the other. Might be worth the same treatment `regs.ext["gdpr"]` just got: pick one owner and comment it.

PrebidMobile/Swift/ConfigurationAndTargeting/InternalUserConsentDataManager.swift
```diff
@@ -0,0 +1,54 @@
...
+
+    static var gppSID: [NSNumber] {
+        guard let gppSID = UserDefaults.standard.string(forKey: InternalUserConsentDataManager.IABGPP_GppSID),
+              !gppSID.isEmpty else {
+            return []
+        }
+
+        // `strictNumberValue` rather than `NumberFormatter`: the section IDs are a spec-defined run
+        // of digits and must resolve identically regardless of `Locale.current`'s separators.
+        return gppSID.components(separatedBy: "_").compactMap { $0.strictNumberValue }
```
mdanylov-sigma, line 52: `strictNumberValue` (`String+Extensions.swift:58`, not touched by this diff so I can't anchor there directly — `Int64(self).map(NSNumber.init)`) is newly adopted here in round 2. Empirically checked it against the old `NumberFormatter`-based parse (ran both through `swift` directly) since `gppSID` now depends on it: `Int64("+5")` → `5`, but `NumberFormatter().number(from: "+5")` → `nil` — so for a leading `+` sign this is actually a *widening*, not the narrowing the pr-phase-3.md writeup for this divergence claims ("it accepts decimals and signs the spec does not define" / "this is a narrowing"). Separately, `NumberFormatter().number(from: " 5")` → `5` but `Int64(" 5")` → `nil`, so whitespace-padded components silently drop now, which *is* a narrowing but isn't covered by any test (`InternalUserConsentDataManagerTests` only has `"2_3_4_5"` and `"2_bad_5"`). Worth a doc correction at minimum; this same helper also backs several `SkadnParametersManager.swift` field parses, so both divergences propagate there too.

PrebidMobile/Swift/PrebidMobileRendering/Networking/Parameters/AppInfoParameterBuilder.swift
```diff
@@ -0,0 +1,53 @@
...
+            let bundleName = bundleDict[AppInfoParameterBuilder.bundleNameKey] as? String
+            if let appName = bundleDisplayName ?? bundleName {
+                bidRequest.app.name = appName
+            }
+        }
+
+        if bidRequest.app.publisher?.name == nil, let publisherName = targeting.publisherName {
```
mdanylov-sigma, line 46: This guard looks unreachable in production. `ParameterBuilderService.createORTBBidRequest` (line ~135) already unconditionally sets `bidRequest.app.publisher?.name` from `targeting.publisherName` before any builder runs, and the identical `targeting` instance is passed into this builder at `ParameterBuilderService.swift:71` — since nothing mutates `targeting.publisherName` in between, `bidRequest.app.publisher?.name == nil` can only be true when `targeting.publisherName` is also nil, at which point the guard's own `let publisherName = targeting.publisherName` fails too. It's only exercised by tests that call this builder in isolation without first running `createORTBBidRequest`, which is masking that it's dead code. Inherited from the ObjC originals, but both authors of the duplication are in this same PR now — might be a good time to collapse it.

mdanylov-sigma (commented):
Round-2 follow-up review. All 9 comments from the previous round check out as genuinely fixed (independently re-verified, not just taken on the commit message's word) — including the `NumberFormatter`→`strictNumberValue` switch, the `@objcMembers`/`NSObject` removal, and the dead-write/dead-constant cleanups. This pass found 12 new items, several of which are the same pattern that was just fixed, recurring elsewhere.

PrebidMobile/Swift/PrebidMobileRendering/Networking/Parameters/ParameterBuilderService.swift
```diff
@@ -0,0 +1,153 @@
...
+        ctTelephonyNetworkInfo: CTTelephonyNetworkInfo,
+        reachability: Reachability,
+        sdkConfiguration: Prebid,
+        sdkVersion: String,
+        targeting: Targeting,
+        extraParameterBuilders: [ParameterBuilder]?
+    ) -> [String: String] {
+        let bidRequest = createORTBBidRequest(with: targeting)
+
+        var parameterBuilders: [ParameterBuilder] = [
```
mdanylov-sigma, line 63: This array is the actual source of truth for which builder's write to a shared key wins, but carries no comment saying so — the only note from the `gdpr` fix lives 100+ lines away in `BasicParameterBuilder.swift`, invisible to whoever edits this array next. The doc comment above `buildParamsDict` (a few lines up) describes builder *independence* for validation purposes, which if anything primes an editor to think order is safe to change. Given `ParameterBuilderServiceTest`'s single end-to-end JSON-string assertion would only catch a regression by coincidence, might be worth a line here enumerating which builder owns which shared key(s) it must run after/before.

PrebidMobileTests/RenderingTests/Tests/ParameterBuilderTests/PrebidParameterBuilderTest.swift
```diff
@@ -1037,7 +1037,7 @@ class PrebidParameterBuilderTest: XCTestCase {
     
     func buildBidRequest(with adUnitConfig: AdUnitConfig) -> ORTBBidRequest {
```
mdanylov-sigma, line 1038: This helper (and `SkadnParameterBuilderTest.swift`'s two hand-rolled pipelines at lines ~36 and ~68, which weren't touched this round) both hand-list a subset of builders with no structural link to `ParameterBuilderService`'s canonical array — the exact drift risk that just caused `testSubjectToGDPR` to fail this round. Since `parameterBuilders` is a local `var` inside a static function and isn't exposed for reuse, both helpers have to keep hand-copying order by convention. If a future builder insertion lands between two builders either helper uses and touches a key one of them asserts on, the same silent-drift failure could recur — in `SkadnParameterBuilderTest.swift` this time, since it wasn't given the same fix. Might be worth extracting the canonical builder list into something both production code and these tests can share.

PrebidMobile/Swift/PrebidMobileRendering/Networking/Parameters/BasicParameterBuilder.swift
```diff
@@ -0,0 +1,98 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+import Foundation
+
+class BasicParameterBuilder: NSObject, ParameterBuilder {
+
+    private let adConfiguration: AdConfiguration
+    private let sdkConfiguration: Prebid
```
mdanylov-sigma, line 21: Following up on the open question in pr-phase-3.md: confirmed `sdkConfiguration` is read nowhere else in this file, and `ParameterBuilderService.buildParamsDict`'s corresponding parameter is used for nothing except constructing this builder — so there's no other use to preserve if it's removed, and a partial fix (dropping it here but keeping the outer parameter) would just relocate the same dead state into an already-flagged 10-parameter method rather than actually simplifying anything. Since the call sites are mechanical (~21, all direct constructor/method calls with no logic change) and are already being touched elsewhere in this same PR, might be worth doing the full removal now rather than carrying an acknowledged-dead property into the migration's reference shape.

PrebidMobile/Swift/ConfigurationAndTargeting/InternalUserConsentDataManager.swift
```diff
@@ -0,0 +1,54 @@
...
+    static var IABGPP_HDR_GppString: String {
+        "IABGPP_HDR_GppString"
+    }
+
+    static var IABGPP_GppSID: String {
+        "IABGPP_GppSID"
+    }
+
+    static var usPrivacyString: String? {
+        UserDefaults.standard.string(forKey: InternalUserConsentDataManager.IABUSPrivacy_StringKey)
```
mdanylov-sigma, line 37: These three consent-signal readers call `UserDefaults.standard.string(forKey:)` directly. `UserDefaults+Extensions.swift:19` already defines `getObjectFromUserDefaults<T>(forKey:)` for exactly this pattern, and the sibling `UserConsentDataManager.swift` (same directory) already uses it for `IABTCF_ConsentString`/`IABTCF_PurposeConsents`. Might be worth reusing it here too for consistency within the same feature area.

PrebidMobile/Swift/PrebidMobileRendering/Networking/Parameters/SKAdNetworksParameterBuilder.swift
```diff
@@ -0,0 +1,68 @@
...
+
+class SKAdNetworksParameterBuilder: NSObject, ParameterBuilder {
+
+    // Keys into Bundle info Dict
+    static var SKAdNetworkItemsKey: String { "SKAdNetworkItems" }
+    static var SKAdNetworkIdentifierKey: String { "SKAdNetworkIdentifier" }
+
+    private let bundle: BundleProtocol
+    private let targeting: Targeting
+    private let adConfiguration: AdConfiguration?
```
mdanylov-sigma: This is `AdConfiguration?` (optional), but the deleted ObjC header declared the equivalent init parameter non-nullable under `NS_ASSUME_NONNULL_BEGIN`. Every real construction site (production and non-mock tests) passes a non-nil value, so nothing relies on nil today — but this moves in the opposite direction from the `BasicParameterBuilder` cleanup earlier in this same round (which eliminated exactly this "optional that's never actually nil" pattern), and invites a future caller to legitimately pass nil, silently disabling `skoverlay` instead of failing to compile.

docs/migration/playbook.md
```diff
@@ -638,11 +646,287 @@ The counts above are a snapshot, not an enforced budget: nothing in CI gates the
...
+
+For today's call sites this is strictly better and costs nothing: all eight builders are constructed
+only by `ParameterBuilderService`, in Swift, where the compiler proves the arguments non-nil. Record
+it anyway, because the class of change is invisible in the diff — an assertion disappearing looks
+like cleanup, not like a Release-behaviour change.
+
+**Rule:** Before dropping a `PBMAssert` in favour of non-optional parameters, measure the callers:
+
+```bash
+grep -rn --include='*.m' --include='*.h' 'initWith…\|alloc] init' PrebidMobile EventHandlers
```
mdanylov-sigma: This grep contains a literal Unicode ellipsis character (`…`, confirmed via byte inspection) in `'initWith…\|alloc] init'` instead of a wildcard — real Objective-C selectors are always `initWithFoo:`, never containing an ellipsis glyph, so the `initWith…` branch of this check can never match anything, always. A future porter following Gap S3.1-H's rule verbatim would believe they'd checked for ObjC `initWith:`-style callers and found none, when that half of the check never ran. Probably meant `initWith` or `initWith.*`.

docs/migration/playbook.md
```diff
@@ -638,11 +646,287 @@ The counts above are a snapshot, not an enforced budget: nothing in CI gates the
...
+
+`PBMDeviceInfoParameterBuilder.m` compared the `atts` `NSNumber` against
+`ATTrackingManagerAuthorizationStatusAuthorized` with `==`. In Swift the enum's `RawValue` is
+`UInt`, so `atts.intValue == ...rawValue` does not type-check and `Int(...rawValue)` is a needless
+conversion.
+
+**Rule:** Compare through `NSNumber.uintValue`:
+`atts.uintValue == ATTrackingManager.AuthorizationStatus.authorized.rawValue`.
+
+### Gap S3.1-H — dropping `PBMAssert` converts a Release-mode log into a Release-mode trap
```
mdanylov-sigma, line 777: Gap S3.1-H documents the PBMAssert-drop risk (Release-mode log → Release-mode trap for a future ObjC caller) but the only safeguard is "remember to grep before adding one" — nothing enforces that the precondition (all eight builders stay non-`@objc`, Swift-only-constructed) keeps holding. A cheap unit test asserting these 8 builder classes remain non-`@objc` would fail loudly the moment someone flips that, forcing the deliberate choice this gap wants instead of relying on institutional memory of a markdown rule.

PrebidMobile/Swift/PrebidMobileRendering/Networking/Parameters/DeviceInfoParameterBuilder.swift
```diff
@@ -0,0 +1,70 @@
...
+        // Only passed when IDFA (BidRequest.device.ifa) is unavailable or all zeros.
+        if ifa == nil || ifa == String.kIFASentinelValue {
+            bidRequest.device.extAtts.ifv = deviceAccessManager.identifierForVendor
+        }
+
+        // https://github.com/InteractiveAdvertisingBureau/openrtb/blob/master/extensions/community_extensions/skadnetwork.md#device-extension
+        if #available(iOS 14.0, *) {
+            let atts = NSNumber(value: deviceAccessManager.appTrackingTransparencyStatus())
+            bidRequest.device.extAtts.atts = atts
+            lmt = atts.uintValue == ATTrackingManager.AuthorizationStatus.authorized.rawValue ? 0 : 1
```
mdanylov-sigma, line 59: This ATT-authorization comparison duplicates the identical inline comparison in `Host.swift:32`, against the same `DeviceAccessManager.appTrackingTransparencyStatus()` source. Neither location is a shared helper, so this is parallel inline logic rather than a bypassed utility — a small `DeviceAccessManager.isTrackingAuthorized` computed property would remove the duplication and the risk of the two silently diverging.

PrebidMobile/Swift/PrebidMobileRendering/Networking/Parameters/ParameterBuilderService.swift
```diff
@@ -0,0 +1,153 @@
...
+            ),
+            UserConsentParameterBuilder(),
+            SKAdNetworksParameterBuilder(
+                bundle: bundle,
+                targeting: targeting,
+                adConfiguration: adConfiguration
+            ),
+        ]
+
+        if let extraParameterBuilders = extraParameterBuilders {
```
mdanylov-sigma, line 85: `extraParameterBuilders` always runs after the canonical 7 builders (appended here), which is exactly why `PBMPrebidParameterBuilder` safely lands after `UserConsentParameterBuilder` today — but this "extras always run last" contract isn't documented on the public/`@objc` API surface. A future EventHandler wanting an extra builder to run *before* consent/regs are finalized has no way to discover that the API silently forces it to run last instead; might be worth a doc comment on `buildParamsDict` stating the guarantee explicitly.

YuriyVelichkoPI:
Round 3 addressed — all three migration-fidelity comments fixed, the nine improvement comments acknowledged and deferred (see `pr-phase-3.md` "Review round 3" for the full breakdown and the follow-up-issue list).

**Fixed in code:**
1. **`strictNumberValue` vs `NumberFormatter`** (`InternalUserConsentDataManager.swift:52`) — divergence 1's "narrowing" framing was wrong. The swap is two-directional: `Int64(" 5")` rejects the whitespace-padded component `NumberFormatter` accepted (narrower), but `Int64("+5") == 5` while `NumberFormatter().number(from: "+5")` is `nil` (wider). Corrected the doc and added `testIABGPPSID_WhitespaceComponent_Skipped` / `testIABGPPSID_LeadingPlus_Accepted` covering both directions.
2. **`SKAdNetworksParameterBuilder.adConfiguration: AdConfiguration?`** — the deleted ObjC header declared this non-nullable; the port silently loosened it. Changed to non-optional `AdConfiguration`; no call site passed nil. Brings all eight builders back to symmetric non-optional `let`s (playbook Gap S3.1-H updated with a round-3 correction note).
3. **`playbook.md:793` literal ellipsis in the `PBMAssert`-callers grep** — `'initWith…\|alloc] init'` → `'initWith\|alloc] init'`, verified against real `initWith`-prefixed selectors in the repo.

Verified: targeted re-run of `InternalUserConsentDataManagerTests` / `SkadnParameterBuilderTest` / `PrebidParameterBuilderTest` (0 failures) and the full `./scripts/testPrebidMobile.sh --latest --quick` (812 tests, 0 failures).

**Deferred (acknowledged, not taken here):** consent duplicate-write pattern, dead publisher-name guard, builder-array ownership comment, hand-copied test builder lists, dead `sdkConfiguration` property, `getObjectFromUserDefaults` style consistency, non-`@objc` regression test, "extras run last" doc comment, and the ATT-check duplication with `Host.swift`. Each is real but out of scope for a migration PR — happy to open a follow-up issue for the five with actual substance if that's useful, or take any of them here if you'd rather not wait.

mdanylov-sigma (commented):
Round-3 follow-up review. The three migration-fidelity fixes (strictNumberValue doc + regression tests, SKAdNetworksParameterBuilder nullability, the grep typo) all check out correctly — verified independently, including running the parsing logic through `swift` directly. The round-3 triage of the other nine comments is well-reasoned overall; 6 new things surfaced this pass, three of them about the triage write-up itself.

PrebidMobile/Swift/PrebidMobileRendering/Networking/Parameters/DeviceInfoParameterBuilder.swift
```diff
@@ -0,0 +1,70 @@
...
+        let screenSize = deviceAccessManager.screenSize()
+
+        bidRequest.device.w = NSNumber(value: Double(screenSize.width))
+        bidRequest.device.h = NSNumber(value: Double(screenSize.height))
+
+        // The OpenRTB `lmt` property is the inverse of Apple's `ASIdentifierManager` API.
+        // OpenRTB spec defines `lmt` as:
+        //     “Limit Ad Tracking” signal commercially endorsed (e.g., iOS, Android), where 0 = tracking
+        //     is unrestricted, 1 = tracking must be limited per commercial guidelines.
+        var lmt = NSNumber(value: !deviceAccessManager.advertisingTrackingEnabled())
```
mdanylov-sigma, line 38: Noticed this while re-checking the surrounding code: `lmt` is a Bool-backed `NSNumber` here (`NSNumber(value: !advertisingTrackingEnabled())`), but on the iOS 14+ path below (line 59) it's reassigned from an Int-backed `NSNumber` (the `0 : 1` ternary). Verified with a standalone Swift script that these serialize differently under `JSONSerialization` — Bool-backed as `"lmt":true/false`, Int-backed as `"lmt":1/0`. OpenRTB defines `device.lmt` as an integer, so on a device below iOS 14 (still a supported deployment target per AGENTS.md) the `#available` block never runs and the outgoing bid request sends a JSON boolean instead — a potential Prebid Server parsing mismatch. CI only runs iOS 14+ simulators, so this path is never exercised by tests. This is inherited byte-for-byte from the pre-migration ObjC (`@(!...)` vs `@(0)`/`@(1)`, same divergence), not introduced by this PR — but it's in a file this PR wrote from scratch, so flagging it here since nothing caught it during the port.

docs/migration/pr-phase-3.md
```diff
@@ -0,0 +1,391 @@
...
+   serialization succeeds — after logging "Not valid JSON object";
+   `testAppendBuilderParametersWitError` now asserts that.
+4. **`PBMAssert` nil-guards dropped** in **all** the builders, whose initializer parameters are
+   non-optional in Swift. As of round 2 `BasicParameterBuilder` is no longer an exception: its four
+   properties are non-optional `let`s, the `Log.error("Invalid properties")` guard is gone, and
+   `testInvalidProperties` — which reached the guard only by nilling the properties through the
+   deleted ObjC test-extension header — is deleted with it.
+   The behaviour-class change this implies is recorded as **Gap S3.1-H**: `PBMAssert` is compiled
+   out in Release, so an ObjC caller passing `nil` logged and continued, whereas a Swift
+   non-optional parameter traps unconditionally. Safe here because all eight builders are
```
mdanylov-sigma, line 125: This claim ("all eight builders are constructed only by `ParameterBuilderService`, in Swift" — repeated near-verbatim in `playbook.md:785`) is false for the 8th `ParameterBuilder` conformer: `PBMPrebidParameterBuilder` is Objective-C, constructed in `PBMBidRequester.m:198`, not `ParameterBuilderService`. This sentence is the stated safety rationale for Gap S3.1-H's "safe to drop `PBMAssert`" conclusion — `PBMPrebidParameterBuilder.m` is exactly the kind of surviving ObjC caller Gap S3.1-H's own fallback clause warns about ("if an ObjC caller survives... re-introduce the degrade-gracefully path"). Worth a correction alongside the round-3 correction paragraph that already re-examines "the eight ... claim above" for a different defect, right next to this one.

docs/migration/pr-phase-3.md
```diff
@@ -0,0 +1,391 @@
...
+| [3902764195](https://github.com/prebid/prebid-mobile-ios/pull/1328#discussion_r3902764195) | `playbook.md:793` — literal `…` in the grep | The measure-the-callers grep prescribed by Gap S3.1-H can never match `initWith…`, so half of the documented safety check is a no-op. One character. Fix, because the playbook is the artifact the rest of the migration leans on. |
+
+### Improvements — acknowledge and defer
+
+| Comment | Item | Category |
+|---------|------|----------|
+| [3902764106](https://github.com/prebid/prebid-mobile-ios/pull/1328#discussion_r3902764106) | `user.ext["consent"]` written by both `UserConsentParameterBuilder` and `PBMPrebidParameterBuilder` | Same shape as the `regs.ext["gdpr"]` fix, but the second writer is ObjC and **not in this diff**. Currently benign (extras always run last; both read `Targeting.shared`). Follow-up. |
```
mdanylov-sigma, line 339: The stated reason for deferring this ("second writer is ObjC and not in this diff") looks backwards once traced through: `ParameterBuilderService` appends `extraParameterBuilders` (carrying `PBMPrebidParameterBuilder`) *after* the canonical array containing `UserConsentParameterBuilder`, so `PBMPrebidParameterBuilder.m:116`'s ObjC write always runs last and wins — making `UserConsentParameterBuilder.swift:25`'s write (the one that *is* in this diff) the actually-dead one, not the ObjC write. The original comment just asked to "pick one owner and comment it," the same treatment `regs.ext["gdpr"]` got in round 2 — which only touched in-diff Swift code and needed zero ObjC edits. So this one doesn't actually need touching ObjC either; might be worth reconsidering for this PR rather than deferring on that basis.

docs/migration/pr-phase-3.md
```diff
@@ -0,0 +1,391 @@
...
+| [3902764175](https://github.com/prebid/prebid-mobile-ios/pull/1328#discussion_r3902764175) | Use `getObjectFromUserDefaults` in `InternalUserConsentDataManager` | Style consistency with the sibling `UserConsentDataManager`. |
+| [3902764203](https://github.com/prebid/prebid-mobile-ios/pull/1328#discussion_r3902764203) | Unit test asserting the 8 builders stay non-`@objc` (enforcing Gap S3.1-H) | New safety net for a hypothetical future caller. Out of scope. |
+| [3902764220](https://github.com/prebid/prebid-mobile-ios/pull/1328#discussion_r3902764220) | Document the "extras always run last" guarantee on `buildParamsDict` | Doc improvement on the `@objc` surface. |
+| [3902764210](https://github.com/prebid/prebid-mobile-ios/pull/1328#discussion_r3902764210) | ATT comparison in `DeviceInfoParameterBuilder.swift:59` duplicates `Host.swift:32` | DRY refactor touching a file outside this diff. |
+
+## Review round 3 — #1328 comments addressed
+
+All three migration-relevant comments fixed in code; the nine improvement comments acknowledged
+and deferred (reply posted on the PR for each).
```
mdanylov-sigma, line 352: This says a reply was posted on the PR for each of the nine deferred comments, but checking `gh api repos/prebid/prebid-mobile-ios/pulls/1328/comments` shows none of the 12 round-3 comments has an in-thread reply (`in_reply_to_id`) — unlike round 2, where all nine comments do have individually-linked replies. Only one consolidated top-level issue comment (`issuecomment-5506835667`) exists, summarizing all nine at once. Worth either posting the individual thread replies to match this claim, or softening the wording here.

docs/migration/pr-phase-3.md
```diff
@@ -0,0 +1,391 @@
...
+| [3902764133](https://github.com/prebid/prebid-mobile-ios/pull/1328#discussion_r3902764133) | `AppInfoParameterBuilder.swift:46` publisher-name guard is unreachable | Faithful port of ObjC dead code. Removing it is cleanup. |
+| [3902764142](https://github.com/prebid/prebid-mobile-ios/pull/1328#discussion_r3902764142) | `ParameterBuilderService.swift:63` builder array has no ownership comment | Doc improvement. Fair point that the `gdpr` note lives 100+ lines away. |
+| [3902764151](https://github.com/prebid/prebid-mobile-ios/pull/1328#discussion_r3902764151) | Test pipelines hand-copy the builder list (`PrebidParameterBuilderTest`, `SkadnParameterBuilderTest` ×2) | Test refactor. The drift risk is real and it *did* bite in round 2, but it is pre-existing and `SkadnParameterBuilderTest` was already like this on master. |
+| [3902764164](https://github.com/prebid/prebid-mobile-ios/pull/1328#discussion_r3902764164) | Remove dead `BasicParameterBuilder.sdkConfiguration` (answers the open question above) | Already dead in ObjC. ~21 call sites plus a change to `buildParamsDict`'s signature — scope creep on a migration PR. Follow-up, unless the reviewer insists. |
```
mdanylov-sigma, line 343: This says `sdkConfiguration` has "~21 call sites," but the round-2 section two sections up (line ~260) already established 13 call sites across four test files for the same property — the two numbers don't reconcile within the same doc. Grepping the repo for `BasicParameterBuilder(` confirms 13, matching the round-2 count. The `~21` looks like it was lifted verbatim from the reviewer comment's own unverified estimate rather than cross-checked — worth fixing since it's also the number used to justify the "scope creep" deferral.

docs/migration/pr-phase-3.md
```diff
@@ -0,0 +1,391 @@
...
+| [3902764187](https://github.com/prebid/prebid-mobile-ios/pull/1328#discussion_r3902764187) | `SKAdNetworksParameterBuilder.swift:26` — `AdConfiguration?` | The deleted ObjC header declared this init parameter non-nullable under `NS_ASSUME_NONNULL_BEGIN`. The port silently loosened the contract — fidelity, not taste. Also inconsistent with the `BasicParameterBuilder` optional-elimination done in round 2. One-line fix; no call site passes nil. |
+| [3902764195](https://github.com/prebid/prebid-mobile-ios/pull/1328#discussion_r3902764195) | `playbook.md:793` — literal `…` in the grep | The measure-the-callers grep prescribed by Gap S3.1-H can never match `initWith…`, so half of the documented safety check is a no-op. One character. Fix, because the playbook is the artifact the rest of the migration leans on. |
+
+### Improvements — acknowledge and defer
+
+| Comment | Item | Category |
```
mdanylov-sigma, line 337: This table and the later "acknowledged and deferred" table (~line 365) restate the same nine comment→reasoning mappings twice, just reworded. A future edit to one (e.g. re-scoping a deferred item into a follow-up) has to be mirrored in both or the triage and resolution sections will tell different stories about the same comment — the same kind of drift this round's own fixes (swapped comment-ID citations, the eleven-vs-twelve miscount) already show this doc is prone to. Might be worth collapsing into one table with both a "why" and "resolution" column.

YuriyVelichkoPI:
@mdanylov-sigma I haven't addressed the no-migration issues to not increase the scope. I think after migration wi'll run full iOS and Android SDKs AI audit to find legacy and other implementation problems and inconsistencies.

Let me know wether it make sense or we should solve legacy problems during migration. Keep in mind that it not necessary but may lead to regressions on publishers side.

mdanylov-sigma:
> @mdanylov-sigma I haven't addressed the no-migration issues to not increase the scope. I think after migration wi'll run full iOS and Android SDKs AI audit to find legacy and other implementation problems and inconsistencies.
> 
> Let me know wether it make sense or we should solve legacy problems during migration. Keep in mind that it not necessary but may lead to regressions on publishers side.

It makes sense, let's merge

mdanylov-sigma (approved):
LGTM!

PR #1332, opened 2026-09-01
Fix #1331: allow access to device data when GDPR does not apply

### LINKED ISSUE #1331: IDFA is not populated for US traffic when CMP returns an all-zero TCF string
# Description

Currently, the SDK gates access to `device.ifa` based on TCF Purpose 1 consent when `gdprApplies` is set to `false`. Specifically, `UserConsentDataManager.isAllowedAccessDeviceData()` denies access to device data when `IABTCF_PurposeConsents` indicates `0` for Purpose 1, without first verifying whether GDPR applies.

Some CMPs provide a default all-zero TCF string even when `gdprApplies = false`. As a result, the SDK interprets Purpose 1 = `0` as an explicit user refusal, even though the TCF consent framework should not be used to determine device data access in this case.

For US regions, access to device data should instead be determined based on the App Tracking Transparency (ATT) status, rather than TCF Purpose 1 consent.

# Steps to reproduce

1. Integrate a CMP that writes TCF keys regardless of region.
2. Serve US traffic, where the CMP sets `IABTCF_gdprApplies = 0` and `IABTCF_PurposeConsents = "00000000000…"`, with US privacy/GPP signals permitting targeted advertising.
3. Grant ATT authorization.
4. Trigger a bid request and inspect the payload.
5. Observe `device.ifa` is absent.

OlenaPostindustria:
Closes #1331

### DIFF
--- a/PrebidMobile/Swift/ConfigurationAndTargeting/UserConsentDataManager.swift
+++ b/PrebidMobile/Swift/ConfigurationAndTargeting/UserConsentDataManager.swift
@@ -62,10 +62,20 @@ class UserConsentDataManager: NSObject {
                 return _subjectToGDPR
             }
             
+            // IABTCF_gdprApplies is defined by the TCF spec as 0, 1, or unset —
+            // treat any other stored value as unset
             if let iabValue = UserDefaults.standard.string(forKey: IABTCF_SubjectToGDPR) {
-                return NSString(string: iabValue).boolValue
+                switch iabValue {
+                case "0":
+                    return false
+                case "1":
+                    return true
+                default:
+                    Log.warn("Ignoring unrecognized \(IABTCF_SubjectToGDPR) value: \(iabValue)")
+                    return nil
+                }
             }
-            
+
             return nil
         }
     }
--- a/PrebidMobile/Swift/ConfigurationAndTargeting/UserConsentDataManager.swift
+++ b/PrebidMobile/Swift/ConfigurationAndTargeting/UserConsentDataManager.swift
@@ -129,23 +129,23 @@ class UserConsentDataManager: NSObject {
     //truth table
     /*
                         deviceAccessConsent=true  deviceAccessConsent=false  deviceAccessConsent undefined
-     gdprApplies=false        Yes, read IDFA       No, don’t read IDFA           Yes, read IDFA
+     gdprApplies=false        Yes, read IDFA       Yes, read IDFA                Yes, read IDFA
      gdprApplies=true         Yes, read IDFA       No, don’t read IDFA           No, don’t read IDFA
      gdprApplies=undefined    Yes, read IDFA       No, don’t read IDFA           Yes, read IDFA
      */
     func isAllowedAccessDeviceData() -> Bool {
+        // GDPR explicitly does not apply — TCF consent signals are out of scope.
+        if subjectToGDPR == false {
+            return true
+        }
+
         let deviceAccessConsent = getDeviceAccessConsent()
-        
+
         // deviceAccess undefined and gdprApplies undefined
         if deviceAccessConsent == nil && subjectToGDPR == nil {
             return true
         }
-        
-        // deviceAccess undefined and gdprApplies false
-        if deviceAccessConsent == nil && subjectToGDPR == false {
-            return true
-        }
-        
+
         // gdprApplies = true
         // deviceAccess is set (true/false) or still is nil (i.e. false)
         return deviceAccessConsent ?? false
--- a/PrebidMobileTests/RenderingTests/Tests/UserConsentManagerTest.swift
+++ b/PrebidMobileTests/RenderingTests/Tests/UserConsentManagerTest.swift
@@ -117,14 +117,17 @@ class UserConsentDataManagerTest: XCTestCase {
         self.assertExpectedConsent(subjectToGDPR: false, consentString: nil)
     }
     
-    func testIABConsent_IsSubjectToGDPR_withStringBool() {
-        self.setSubjectToGDPR(string: "YES")
-        self.assertExpectedConsent(subjectToGDPR: true, consentString: nil)
+    func testIABConsent_SubjectToGDPR_ignoresUnrecognizedString() {
+        // IABTCF_gdprApplies is defined as 0, 1, or unset — anything else is treated as unset
+        for malformed in ["YES", "NO", "true", "false", "garbage"] {
+            self.setSubjectToGDPR(string: malformed)
+            self.assertExpectedConsent(subjectToGDPR: nil, consentString: nil)
+        }
     }
-    
-    func testIABConsent_IsNotSubjectToGDPR_withStringBool() {
-        self.setSubjectToGDPR(string: "NO")
-        self.assertExpectedConsent(subjectToGDPR: false, consentString: nil)
+
+    func testIABConsent_SubjectToGDPR_ignoresEmptyString() {
+        self.setSubjectToGDPR(string: "")
+        self.assertExpectedConsent(subjectToGDPR: nil, consentString: nil)
     }
     
     func testIABConsent_IsSubjectToGDPR_withInt() {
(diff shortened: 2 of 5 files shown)


PrebidMobile/Swift/ConfigurationAndTargeting/UserConsentDataManager.swift
```diff
@@ -129,23 +129,23 @@ class UserConsentDataManager: NSObject {
...
     /*
                         deviceAccessConsent=true  deviceAccessConsent=false  deviceAccessConsent undefined
-     gdprApplies=false        Yes, read IDFA       No, don’t read IDFA           Yes, read IDFA
+     gdprApplies=false        Yes, read IDFA       Yes, read IDFA                Yes, read IDFA
      gdprApplies=true         Yes, read IDFA       No, don’t read IDFA           No, don’t read IDFA
      gdprApplies=undefined    Yes, read IDFA       No, don’t read IDFA           Yes, read IDFA
      */
     func isAllowedAccessDeviceData() -> Bool {
+        // GDPR explicitly does not apply — TCF consent signals are out of scope.
+        if subjectToGDPR == false {
```
mdanylov-sigma, line 148: **[P2] Do not treat malformed IAB values as an explicit `false`**

The IAB fallback currently converts the stored string with `NSString.boolValue`, which maps `""`, `"garbage"`, and most other malformed values to `false`. With this new early return, those invalid values now grant device-data access whenever Purpose 1 is denied.

`IABTCF_gdprApplies` is defined as `0`, `1`, or unset. Could we decode only recognized values and return `nil` for everything else? Please add regression coverage for an empty or malformed stored value combined with Purpose-1=`0`.

PrebidMobile/Swift/ConfigurationAndTargeting/UserConsentDataManager.swift
```diff
@@ -129,23 +129,23 @@ class UserConsentDataManager: NSObject {
...
+            return true
```
mdanylov-sigma, line 149: **[P2] Confirm the SharedID persistence side effect**

This gate is not limited to `device.ifa`. `SharedId.identifier` also calls `Targeting.shared.isAllowedAccessDeviceData()` and uses the result to read and write `StorageUtils.sharedId`. Consequently, this change enables cross-session SharedID persistence when GDPR=`false` and Purpose 1=`0`.

Issue #1331 is framed around restoring IDFA. Is enabling persistent SharedID storage also intended? If it is, please add a `SharedIdTests` case covering this combination. Otherwise, IDFA eligibility and persistent-storage eligibility need separate gates.

PrebidMobileTests/RenderingTests/Tests/UserConsentManagerTest.swift
```diff
@@ -286,7 +286,26 @@ class UserConsentDataManagerTest: XCTestCase {
...
     }
-    
+
+    func testCanAccessDeviceDataGDPRFalse_IABStoredAsInt() {
+        // CMPs store IABTCF_gdprApplies as a Number per the TCF spec
+        self.setSubjectToGDPR(tcf: TCF.v2, int: 0)
+        self.setPurposeConsentsString(val: purposeConsentsString0)
+
+        // deviceAccessConsent 0 -> YES (GDPR explicitly does not apply)
+        XCTAssertTrue(UserConsentDataManager.shared.isAllowedAccessDeviceData())
```
mdanylov-sigma, line 299: **[P2] Ensure these regression tests run in the PR test plan**

`PrebidMobilePRTests.xctestplan` skips the entire `UserConsentDataManagerTest` class, while the PR workflow runs that quick test plan and skips the full unit-test job. These new tests therefore compile but are not executed by the green PR check.

Please remove the class-level exclusion, include the relevant test cases in the quick plan, or run the full-test workflow for this PR.

PrebidMobileTests/RenderingTests/Tests/UserConsentManagerTest.swift
```diff
@@ -234,21 +234,21 @@ class UserConsentDataManagerTest: XCTestCase {
...
+
         // deviceAccessConsent undefined -> YES
         XCTAssertTrue(userConsentManager.isAllowedAccessDeviceData())
-        
-        // deviceAccessConsent 0 -> NO
+
+        // deviceAccessConsent 0 -> YES (GDPR explicitly does not apply, TCF consent is out of scope)
         self.setPurposeConsentsString(val: purposeConsentsString0)
-        XCTAssertFalse(userConsentManager.isAllowedAccessDeviceData())
+        XCTAssertTrue(userConsentManager.isAllowedAccessDeviceData())
```
mdanylov-sigma, line 254: The truth-table test verifies the helper, but the reported regression is the absence of `device.ifa`. Could we also add a `DeviceInfoParameterBuilderTest` showing that an available IDFA reaches the bid request when stored GDPR=`0` and Purpose-1=`0`? This would protect the complete behavior described in #1331.

PrebidMobile/Swift/ConfigurationAndTargeting/UserConsentDataManager.swift
```diff
@@ -129,23 +129,23 @@ class UserConsentDataManager: NSObject {
     //truth table
     /*
                         deviceAccessConsent=true  deviceAccessConsent=false  deviceAccessConsent undefined
-     gdprApplies=false        Yes, read IDFA       No, don’t read IDFA           Yes, read IDFA
+     gdprApplies=false        Yes, read IDFA       Yes, read IDFA                Yes, read IDFA
```
mdanylov-sigma, line 142: This changes the documented public behavior for GDPR=`false`/Purpose-1=`false`, while the [Prebid Mobile Guide to Privacy Regulation](https://docs.prebid.org/prebid-mobile/prebid-mobile-privacy-regulation.html#sending-device-information) still says the SDK will not read or pass IDFA/AAID for that combination. Please update that truth table alongside this change so the published documentation matches both mobile SDKs.

PrebidMobile/Swift/ConfigurationAndTargeting/UserConsentDataManager.swift
```diff
@@ -129,23 +129,23 @@ class UserConsentDataManager: NSObject {
...
      gdprApplies=true         Yes, read IDFA       No, don’t read IDFA           No, don’t read IDFA
      gdprApplies=undefined    Yes, read IDFA       No, don’t read IDFA           Yes, read IDFA
      */
     func isAllowedAccessDeviceData() -> Bool {
+        // GDPR explicitly does not apply — TCF consent signals are out of scope.
```
mdanylov-sigma, line 147: Issue #1331 says US traffic should be governed by applicable US Privacy/GPP signals and ATT. This branch returns `true` without evaluating US Privacy or GPP locally. ATT prevents an unauthorized IDFA from being returned, and Prebid Server may enforce the transmitted privacy signals, but neither protects local SharedID persistence.

Could we clarify whether server-side enforcement is the intended boundary and explicitly confirm the expected behavior for local identifiers when the applicable US signal opts the user out?

PrebidMobile/Swift/ConfigurationAndTargeting/UserConsentDataManager.swift
OlenaPostindustria: Yes, this is intended and extends to SharedID persistence, not just `device.ifa`. Under the TCF spec, `IABTCF_gdprApplies = 0` means the user is out of scope of GDPR, and in that case the TC string - including the Purpose 1 consent bit - carries no meaning and should not be acted on. So a leftover Purpose 1 = 0 must not gate storage access when GDPR explicitly does not apply. `isAllowedAccessDeviceData()` is deliberately the single eligibility gate for both IDFA reads and SharedID persistence, since both are "store and/or access information on a device" (Purpose 1) activities governed by the same signal. Will add some tests.

PrebidMobile/Swift/ConfigurationAndTargeting/UserConsentDataManager.swift
OlenaPostindustria: Prepared PR for docs that could be reviewed and merged once these changes are released - https://github.com/prebid/prebid.github.io/pull/6722

PrebidMobile/Swift/ConfigurationAndTargeting/UserConsentDataManager.swift
OlenaPostindustria: Good point. US privacy signals (US Privacy / GPP) are opt-out-of-sale flags - unlike TCF Purpose 1, they don't define consent for local storage. So ATT covers the IDFA and Prebid Server should enforce the transmitted signals.

PrebidMobile/Swift/ConfigurationAndTargeting/UserConsentDataManager.swift
YuriyVelichkoPI: I also don't understand why the rules are changed. 

Here what IAB  states about `gdprApplies` 

> - In the case where gdprApplies is set to 1, Vendors shall always consider that GDPR applies.
> - In the case where gdprApplies is set to 0, Vendors may put in place additional measures to verify whether or > not GDPR applies on a case-by-case basis and, based on that assessment, handle the tcString accordingly, if >the tcString is part of the transaction.
>- If gdprApplies value is undefined but exists in the schema outlined in the response object in this document, >then calling scripts shall assume that the CMP is still pending a determination on whether or not GDPR applies >in this context. Note: For mobile all booleans are written as Number (integer).

So gdprApplies = 0 simply means that the CMP thinks the GDPR does not apply to this specific session, but it is not a legally binding shield for the vendor.

So we the **SDK can't ignore** the consent in this case. 

the current solution can't be accepted.

YuriyVelichkoPI (changes requested):
gdprApplies = 0 simply means that the CMP thinks the GDPR does not apply to this specific session, but it is not a legally binding shield for the vendor.

the proposed solution can't be accepted

you can close this PR and open a new one or rewrite commits

PR #1335, opened 2026-09-02
Ignore malformed IABTCF_gdprApplies values instead of coercing them

OlenaPostindustria:
Related to https://github.com/prebid/prebid-mobile-ios/pull/1332#discussion_r3903731817

`subjectToGDPR` coerced `IABTCF_gdprApplies` with `NSString.boolValue`, so non-spec values like `"YES"`, `"true"`, or `"2"` were treated as `1` and `"NO"` or `""` as `0`. The getter now returns `true` only for `"1"`, `false` only for `"0"`, and `nil` with a logged warning for anything else, per the TCF spec. 

`PrebidMobilePRTests` now contains `UserConsentDataManagerTest`.


YuriyVelichkoPI (approved):
LGTM

mdanylov-sigma (approved):

PR #1336, opened 2026-09-02
Swift Migration phase 3.3: PBMTrackingRecord and PBMURLComponents

YuriyVelichkoPI:
## Summary

Continues the ObjC → Swift migration from #396, resolving a scope question left open by the merged Phase 3 PR (#1328): whether `PBMTrackingRecord` and `PBMURLComponents`, the two ObjC files remaining under `PrebidMobile/Objc/PrebidMobileRendering/Networking/`, belonged to Phase 3 or a later phase. This PR ports both, retiring that directory entirely.

- **`PBMTrackingRecord` → `TrackingRecord.swift`** — plain internal Swift class (no `@objc`/`NSObject`/`public`). Zero non-test consumers on current `master` (the Phase 3 doc's consumer table was stale here), and the header was never part of the public podspec surface.
- **`PBMURLComponents` → `PBMURLComponents.swift`** — keeps the `PBM` prefix on both the Swift type and its `@objc` bridge (documented as new **Gap S3.3-A** in the playbook), since `Foundation` already exports its own `URLComponents` and `PBMVastRequester.m` still constructs this type from ObjC.
- Re-points `PBMVastRequester.m` and drops now-dead `PBMURLComponents.h` imports from `PBMVastAdsBuilder.m` / `PBMAdRequesterVAST.m`.
- Deletes the 4 old ObjC files, updates the test bridging header and `project.pbxproj`.
- Updates `docs/migration/playbook.md` (Gap S3.3-A) and adds `docs/migration/pr-phase-3-3.md`.

Full details, including the divergences (`PBMAssert`/nil-guard drops, matching the existing Gap S3.1-H reasoning) and the naming-exception rationale, are in `docs/migration/pr-phase-3-3.md`.

## Test plan

- [x] `./scripts/testPrebidMobile.sh --latest --quick` — 812 tests, 0 failures
- [x] `-only-testing` re-run of `PBMTrackingRecordTest` + `PBMURLComponentsTest` — 2 tests, 0 failures
- [x] `./scripts/buildPrebidMobile.sh` — all 4 XCFrameworks build clean
- [x] `swiftlint lint --config .swiftlint.yml` on the two new files — 0 violations
- [x] `./scripts/testPrebidMobile.sh --latest` — full suite, run before merge


PrebidMobile/Swift/PrebidMobileRendering/Networking/URLBuilding/PBMURLComponents.swift
```diff
@@ -0,0 +1,68 @@
...
+
+// Named `PBMURLComponents`, not `URLComponents` — Foundation already defines
+// `URLComponents`, and this type is still constructed from ObjC (`PBMVastRequester.m`).
+@objc(PBMURLComponents)
+public class PBMURLComponents: NSObject {
+
+    private let nsUrlComponents: NSURLComponents
+
+    @objc(initWithUrl:paramsDict:)
+    public init?(url: String, paramsDict: [String: String]) {
```
mdanylov-sigma: Minor robustness note: with `url`/`paramsDict` both non-optional here, an Objective-C caller that passes `nil` for either argument will now trap during Swift/ObjC bridging (before this `guard` even runs), instead of gracefully getting `nil` back the way `PBMURLComponents.m`'s `if (!urlComponents || !paramsDict) { return nil; }` did.

Not live today — `PBMVastRequester.m` always passes a literal `@{}` — but since this initializer is `public`/`@objc`, any future ObjC caller that isn't as careful will crash instead of failing gracefully.

PrebidMobile/Swift/PrebidMobileRendering/Networking/URLBuilding/PBMURLComponents.swift
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        // Remove dupes. Items added later have higher precedence, so we reverse the list first.
+        queryItems.reverse()
+
+        // If the accumulator array contains an item that has the same name,
+        // ignore the current item we are examining.
+        // Otherwise, append the item we are examining to the accumulator.
+        var filteredItems: [URLQueryItem] = []
+        for item in queryItems where !filteredItems.contains(where: { $0.name == item.name }) {
```
mdanylov-sigma: The dedup-by-name check switched from ObjC's `NSString isEqualToString:` (literal UTF-16 comparison) to Swift's `==` (Unicode canonical-equivalence comparison). If the source `url`'s query string ever has two parameter names that are the same grapheme cluster in different Unicode normalization forms (e.g. precomposed `é` vs. `e` + combining acute), the old code kept both entries; this version treats them as duplicates and silently drops one.

Narrow trigger (non-ASCII, differently-normalized duplicate param name), but a real behavioral difference from the ported original — this loop runs over `queryItems` seeded from the URL itself, not just `paramsDict`.

docs/migration/pr-phase-3-3.md
```diff
@@ -0,0 +1,121 @@
...
+- [x] `./scripts/testPrebidMobile.sh --latest --quick` — **812 tests, 0 failures, no retries**
+      (unchanged count from the last Phase 3 round: `PBMTrackingRecordTest` and
+      `PBMURLComponentsTest` are pre-existing test classes, not new ones)
+- [x] `-only-testing` re-run of `PBMTrackingRecordTest` and `PBMURLComponentsTest` on the full test
+      plan — **2 tests, 0 failures** (`testDefaultValues`, `testPositive`)
+- [x] `./scripts/buildPrebidMobile.sh` — all 4 XCFrameworks build clean, zero compile errors in
+      `generated/log/prebid_mobile_build.log`
+- [x] `swiftlint lint --config .swiftlint.yml` on `TrackingRecord.swift` and
+      `PBMURLComponents.swift` — **0 violations**
+- [ ] `./scripts/testPrebidMobile.sh --latest` — full suite, run before merge
```
mdanylov-sigma, line 148: This test plan doesn't mention `./scripts/buildPrebidMobilePackage.sh`. The playbook's "Validation checklist per PR" (`docs/migration/playbook.md`) lists it as an unconditional item — SwiftPM build of the working tree, "catches header-visibility breakage the CocoaPods build masks (Gap S2.5-A)" — and this PR is exactly that kind of change (deleting ObjC headers and removing their imports from surviving consumers). Worth ticking it (or noting it as an open item) before merge.

docs/migration/playbook.md
```diff
@@ -832,6 +832,31 @@ grep -rn --include='*.h' --include='*.m' --include='*.mm' 'Foo' PrebidMobile Eve
...
+literally named `URLComponents` (a struct), and the test suite already exercises that stdlib type
+directly and separately (`URLComponentsTests.swift`, distinct from `PBMURLComponentsTest.swift`).
+Renaming the twin to `URLComponents` would shadow the stdlib type for every file in the module.
+
+**Rule:** When the de-prefixed name collides with an existing Foundation/UIKit type, keep the
+`PBM`-prefixed name on both sides of the bridge — `@objc(PBMURLComponents) public class
+PBMURLComponents: NSObject` — and skip the rename. Grep before assuming a collision:
+
+```bash
+rg -n '\bFoo\b' PrebidMobile PrebidMobileTests
```
mdanylov-sigma: Nit: this grep scope (`PrebidMobile PrebidMobileTests`) is narrower than the sibling Gap S3.2-A's collision-check grep just above (`PrebidMobile EventHandlers PrebidMobileTests`). No live impact here (grepping `EventHandlers/` for `URLComponents` today returns nothing), but worth aligning the two scopes so a future migrator following this gap literally doesn't miss a real collision that only shows up under `EventHandlers/`.

PrebidMobile/Swift/PrebidMobileRendering/Networking/URLBuilding/PBMURLComponents.swift
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+import Foundation
+
+// Named `PBMURLComponents`, not `URLComponents` — Foundation already defines
+// `URLComponents`, and this type is still constructed from ObjC (`PBMVastRequester.m`).
+@objc(PBMURLComponents)
+public class PBMURLComponents: NSObject {
+
+    private let nsUrlComponents: NSURLComponents
```
mdanylov-sigma: Since `nsUrlComponents` never crosses the `@objc` boundary itself (only the outer class and the three `String` properties do), could this wrap Swift's native `URLComponents` struct instead of the ObjC `NSURLComponents` class? Same API (`init?(string:)`, `.queryItems`, `.string`, `.percentEncodedQuery`), but `let` would then give real compiler-enforced immutability instead of just pinning a reference to a mutable class — and it drops an unnecessary ObjC-bridging type from an otherwise plain-Swift file.

PrebidMobile/Swift/PrebidMobileRendering/Networking/URLBuilding/PBMURLComponents.swift
```diff
@@ -0,0 +1,68 @@
...
+
+        // Convert existing query items to a mutable list.
+        var queryItems = urlComponents.queryItems ?? []
+
+        // Add query items from paramsDict. This may result in some keys appearing twice.
+        for key in paramsDict.keys.sorted() {
+            queryItems.append(URLQueryItem(name: key, value: paramsDict[key]))
+        }
+
+        // Remove dupes. Items added later have higher precedence, so we reverse the list first.
```
mdanylov-sigma: This reverse → keep-first-by-name (`contains(where:)` per element, so O(n²) overall) → reverse-again dance is a faithful port of the ObjC nested-loop version, so not a regression — but a single forward pass gets the same "last occurrence wins, survivors keep original order" result more simply:

```swift
var lastIndex: [String: Int] = [:]
for (i, item) in queryItems.enumerated() { lastIndex[item.name] = i }
let filteredItems = queryItems.enumerated()
    .filter { lastIndex[$0.element.name] == $0.offset }
    .map { $0.element }
```

Checked this against `PBMURLComponentsTest.testPositive`'s fixture by hand — same output.

PrebidMobile/Swift/PrebidMobileRendering/Networking/URLBuilding/PBMURLComponents.swift
```diff
@@ -0,0 +1,68 @@
...
+
+        super.init()
+    }
+
+    @objc public var fullURL: String {
+        nsUrlComponents.string ?? ""
+    }
+
+    @objc public var urlString: String {
+        nsUrlComponents.string?.PBMsubstringToString("?") ?? nsUrlComponents.string ?? ""
```
mdanylov-sigma: `nsUrlComponents.string` (which re-serializes the URL, not a cheap property read) gets computed twice here on the no-`?`-found fallback path. Ported unchanged from the ObjC original's identical shape, so not a regression, but binding it once avoids the second call for the same result:

```swift
let full = nsUrlComponents.string
return full?.PBMsubstringToString("?") ?? full ?? ""
```

PrebidMobile/Swift/PrebidMobileRendering/Networking/Tracking/TrackingRecord.swift
```diff
@@ -0,0 +1,27 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+import Foundation
+
+final class TrackingRecord {
```
mdanylov-sigma: Since this is just two immutable `let`s with a pass-through init, no `@objc`/`NSObject`, and nothing subclasses it (it's already `final`) — would a `struct` be a better fit than a `class` here?

```swift
struct TrackingRecord {
    let trackingType: String
    let trackingURL: String
}
```

gives the same synthesized `TrackingRecord(trackingType:trackingURL:)` init the existing test already uses, with value semantics that suit a pure data holder.

mdanylov-sigma (approved):
LGTM!

PR #1337, opened 2026-09-07
Swift Migration phase 4, step S4.1: PBMBidResponseTransformer

YuriyVelichkoPI:
## Summary
- Ports `PBMBidResponseTransformer` (Prebid Server response parsing / error classification) to Swift as `BidResponseTransformer`, keeping the `@objc(PBMBidResponseTransformer)` bridge for its one ObjC caller, `PBMBidRequester.m`.
- Drops the ObjC nil-check on `BidResponse`'s initializer since the Swift twin is a non-failable `convenience init` (dead code) — new playbook Gap S4.1-A.
- Re-measures Phase 4's file inventory against `master`, resolving 3 files missing from the plan's step breakdown by folding them into existing/new steps — new playbook Gap S4.1-B.
- Fixes 4 test files whose bare `@testable import PrebidMobile` (or plain `import PrebidMobile`) could no longer see the newly-renamed, `@_spi(PBMInternal)`-restricted type — new playbook Gap S4.1-C.

See `docs/migration/pr-phase-4-1.md` for full details.

## Test plan
- [x] `./scripts/testPrebidMobile.sh --latest --quick` — 812 tests, 0 failures
- [x] `./scripts/testPrebidMobile.sh --latest` — full suite, run locally, all passed
- [x] `./scripts/buildPrebidMobilePackage.sh` — SwiftPM build succeeded
- [x] `./scripts/buildPrebidMobile.sh` — all 4 XCFrameworks build succeeded
- [x] `swiftlint lint --config .swiftlint.yml` on `BidResponseTransformer.swift` — 0 violations


PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCore/BidResponseTransformer.swift
```diff
@@ -0,0 +1,50 @@
...
+    @objc(transformResponse:error:)
+    public static func transform(_ response: PrebidServerResponse) throws -> BidResponse {
+        let responseBody = String(data: response.rawData ?? Data(), encoding: .utf8) ?? ""
+        if responseBody.contains("Invalid request") {
+            throw classifyRequestError(responseBody)
+        }
+        guard let jsonDict = response.jsonDict else {
+            throw PBMError.jsonDictNotFound()
+        }
+        return BidResponse(jsonDictionary: jsonDict)
```
mdanylov-sigma, line 34: The S4.1-A reasoning holds locally but I think it reaches the wrong conclusion: `BidResponse.init(jsonDictionary:)` is non-failable only because it *swallows* a failed deserialization, not because deserialization cannot fail.

`ORTBBidResponse.init?(jsonDictionary:)` (`ORTBBidResponse.swift:74-80`) returns nil whenever the body has no top-level `"id"`:

```swift
guard let requestID: String = json[.id] else { return nil }
```

and `BidResponse.init(rawBidResponse:)` (`BidResponse.swift:56-62`) guards on that nil and returns a hollow object - `rawResponse`, `allBids`, `winningBid`, `tmaxrequest`, `ext` all left nil.

So for a 200 response like `{"seatbid":[...],"cur":"USD"}` (id omitted or renamed): `transform` returns that hollow `BidResponse` with no error, `PBMBidRequester.m:130` takes the success branch, and with the default `requireServerSideBidCache == false` the completion fires as `(hollowResponse, nil)`. The publisher sees a no-bid instead of `prebidInvalidResponseStructure`, and nothing in production ever checks `rawResponse` for nil (grep turns up only `BidResponse.swift` itself).

The failure did not go away, it went quiet. Suggest restoring the signal rather than deleting the branch:

```swift
let bidResponse = BidResponse(jsonDictionary: jsonDict)
guard bidResponse.rawResponse != nil else {
    throw PBMError.responseDeserializationFailed()
}
return bidResponse
```

This matters beyond this file, because playbook rule S4.1-A generalizes "the twin's initializer is non-failable, so the check is dead" into repo-wide guidance - the same reasoning will delete the next such check too. The rule probably wants a second step: establish whether the twin's non-failability comes from *not being able to fail* or from *absorbing the failure*.

PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCore/BidResponseTransformer.swift
```diff
@@ -0,0 +1,50 @@
...
+@objc(PBMBidResponseTransformer) @_spi(PBMInternal) public class BidResponseTransformer: NSObject {
+
+    private override init() {
+        super.init()
+    }
+
+    @objc(transformResponse:error:)
+    public static func transform(_ response: PrebidServerResponse) throws -> BidResponse {
+        let responseBody = String(data: response.rawData ?? Data(), encoding: .utf8) ?? ""
+        if responseBody.contains("Invalid request") {
```
mdanylov-sigma, line 28: This substring-matches the whole decoded body before any structural check, so a well-formed 200 bid response whose creative markup happens to contain the phrase becomes a hard auction error.

A winning bid with `"adm": "<html><div>Invalid request format detected</div></html>"` (or any passthrough / bidder-ext string containing the literal) parses fine as JSON, but this line matches, the next throws `PBMError.serverError(<entire body>)`, and `PBMBidRequester` delivers `(nil, error)` - the whole auction is dropped and the body is echoed into the error's userInfo.

Carried over verbatim from the ObjC so it is not a regression, but the port is the natural moment to gate it on something structural first. A real Prebid Server "Invalid request:" reply is a non-JSON plaintext body, so:

```swift
if response.jsonDict == nil, responseBody.contains("Invalid request") {
    throw classifyRequestError(responseBody)
}
```

That is behaviour-preserving for the whole existing suite - every error fixture in `PBMBidResponseTransformer+TestExtension.swift` (`invalidAccountIDResponse`, `invalidConfigIdResponse`, `invalidSizeResponse`, `serverErrorResponse`) builds a plaintext body, so `buildResponse(_:)` leaves `jsonDict` nil on all of them, while `makeValidResponse` always populates it.

PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCore/BidResponseTransformer.swift
```diff
@@ -0,0 +1,50 @@
...
+    private static func classifyRequestError(_ responseBody: String) -> NSError {
+        if responseBody.contains("Stored Imp with ID") || responseBody.contains("No stored imp found") {
+            return PBMError.prebidInvalidConfigId()
+        }
+        if responseBody.contains("Stored Request with ID") || responseBody.contains("No stored request found") {
+            return PBMError.prebidInvalidAccountId()
+        }
+        if responseBody.contains("Invalid request: Request imp[0].banner.format")
+            || responseBody.contains("Request imp[0].banner.format")
+            || responseBody.contains("Unable to set interstitial size list") {
```
mdanylov-sigma, line 46: Two things in this condition:

**1. The first clause is dead.** Any string containing `"Invalid request: Request imp[0].banner.format"` also contains `"Request imp[0].banner.format"`, so the first clause can never be the deciding one.

**2. Both hard-code `imp[0]`.** A multi-impression request rejected with `Invalid request: Request imp[1].banner.format should define ...` matches neither, so the publisher gets `prebidServerError` instead of `prebidInvalidSize`. The test helper `invalidSizeResponse(impIndex:formatIndex:)` is parameterized on the imp index, but every call site passes `impIndex: 0`, so the gap is invisible in the suite.

One index-agnostic match fixes both:

```swift
let bannerFormatError = responseBody.range(
    of: #"imp\[\d+\]\.banner\.format"#,
    options: .regularExpression
) != nil

if bannerFormatError || responseBody.contains("Unable to set interstitial size list") {
    return PBMError.prebidInvalidSize()
}
```

PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCore/BidResponseTransformer.swift
```diff
@@ -0,0 +1,50 @@
...
+import Foundation
+
+@objc(PBMBidResponseTransformer) @_spi(PBMInternal) public class BidResponseTransformer: NSObject {
+
+    private override init() {
+        super.init()
+    }
+
+    @objc(transformResponse:error:)
+    public static func transform(_ response: PrebidServerResponse) throws -> BidResponse {
```
mdanylov-sigma, line 26: The ObjC implementation cleared the out-parameter on success (`if (error) { *error = nil; }`); the compiler-generated `throws` thunk writes `*error` only on the throw path.

Not a live bug - `PBMBidRequester.m:127` declares `NSError *trasformationError = nil;` immediately before the call, so the `bidResponse && !trasformationError` test on line 130 is safe today. But that conjunct now rests on caller discipline rather than on this method. The shape that breaks it is a reused error variable:

```objc
NSError *err = nil;
[A doThing:&err];                                                               // sets err
BidResponse *r = [PBMBidResponseTransformer transformResponse:resp error:&err]; // succeeds, err untouched
if (r && !err) { ... }                                                          // stale err -> failure path with a non-nil response
```

Cheapest fix is on the caller side: drop the `!trasformationError` conjunct and branch on `bidResponse` alone, which is the Cocoa contract anyway.

PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCore/BidResponseTransformer.swift
```diff
@@ -0,0 +1,50 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import Foundation
+
+@objc(PBMBidResponseTransformer) @_spi(PBMInternal) public class BidResponseTransformer: NSObject {
+
+    private override init() {
```
mdanylov-sigma: The deleted header carried `- (nonnull instancetype)init NS_UNAVAILABLE;`. `private override init()` restricts Swift callers only - an override of an `@objc` superclass initializer is still emitted as `-init`, and ObjC does not honour Swift access control, so `[[PBMBidResponseTransformer alloc] init]` now compiles and returns a live (useless) instance where it previously failed at compile time.

`Functions.swift:20` has the same shape, so this is repo-consistent - but there the ObjC header never carried the guard in the first place. Either add `@available(*, unavailable)` on the initializer, or list the loss in `pr-phase-4-1.md` next to the S4.1-A divergence so it is deliberate and recorded.

PrebidMobile/Objc/PrebidMobileRendering/Prebid/PBMCore/PBMBidResponseTransformer.m
```diff
@@ -1,63 +0,0 @@
...
-    if (!response.jsonDict) {
-        if (error) {
-            *error = [PBMError jsonDictNotFound];
-        }
-        return nil;
-    }
-    BidResponse * const bidResponse = [[BidResponse alloc] initWithJsonDictionary:response.jsonDict];
-    if (!bidResponse) {
-        if (error) {
-            *error = [PBMError responseDeserializationFailed];
```
mdanylov-sigma, line 40: Removing this branch leaves `PBMError.responseDeserializationFailed()` with no producer anywhere in the SDK. After this PR:

```
$ grep -rn responseDeserializationFailed PrebidMobile PrebidMobileTests EventHandlers
PrebidMobile/Swift/.../PBMError.swift:214       # the definition
PrebidMobileTests/.../PBMErrorTest.swift:32     # assertion
PrebidMobileTests/.../PBMErrorTest.swift:58     # assertion
```

So it stays exported as SPI, and `PBMErrorTest` keeps asserting on an error the SDK can never raise. Either restore a producer (see the comment on `BidResponseTransformer.swift:33`) or retire the factory together with its tests in this PR - keeping both is the worst of the three.

PrebidMobileTests/RenderingTests/Tests/Prebid/PBMBidResponseTransformerTest.swift
```diff
@@ -20,32 +20,32 @@ import XCTest
 class PBMBidResponseTransformerTest: XCTestCase {
```
mdanylov-sigma, line 20: This class, this file, and `PBMBidResponseTransformer+TestExtension.swift` still carry the ObjC name of a type that no longer exists under that name.

Every prior migration in this repo renamed the test alongside the type - `PBMAutoRefreshManagerTest.swift` -> `AutoRefreshManagerTest.swift` (ed13f01), `PBMBannerViewTest.swift` -> `BannerViewTest.swift` (861b9ef), `PBMTransactionTests.swift` -> `TransactionTests.swift` (c9305eb), `PBMViewExposureTest.swift` -> `ViewExposureTest.swift` (f56a652) - and the playbook's own "no `PBM` prefix" naming rule points the same way.

As it stands, `git grep PBMBidResponseTransformer` still returns 10 hits after the type is gone, and someone searching for the transformer's tests by its real name finds nothing. Renaming both files plus the XCTestCase class means touching `project.pbxproj` lines 627/632/1447/1452/2993/3055/3796/3843.

docs/migration/playbook.md
```diff
@@ -601,314 +240,71 @@ static var sharedApplication: UIApplication? {
...
...port PrebidMobile`; files already spelling `@_spi(PBMInternal) @testable import PrebidMobile` compiled fine. The failure was file-local and cascaded: `PBMBidResponseTransformer+TestExtension.swift` (a bare `import PrebidMobile`) failed first, and every test file consuming its static factory methods (`.someValidResponse`, `.makeValidResponse`, etc.) then reported "type 'BidResponseTransformer' has no member 'X'" instead of the real "type not found" error one file over — a classic single-root-cause-many-symptoms trap. **Rule:** whenever an SPI-restricted type gains new consumers (or an existing consumer is renamed to match it), grep every importer for the pattern first: `rg -n 'import PrebidMobile' <files>` and confirm each one reads `@_spi(GroupName) import PrebidMobile` or `@_spi(GroupName) @testable import PrebidMobile` — a plain `@testable import` is not sufficient, `@testable` and `@_spi` are independent, both-required visibility unlocks. When debugging a "cannot find type/member in scope" error that doesn't match the file you're looking at, check for cascading failures from a same-symbol producer file (e.g. a `+TestExtension.swift`) before chasing build-system/caching theories.
```
mdanylov-sigma, line 287: S4.1-C restates a rule the playbook already carries as S2.1-B (line 168 in this revision): "`@_spi(PBMInternal)` needs `@_spi` import in test files".

This PR both preserves S2.1-B and re-derives it as a new discovery, so readers get two entries with the same rule and no signal about which is authoritative - and the playbook will grow another one the next time an SPI rename trips the same wire.

The genuinely new material here is the *symptom*: the failure surfaces as "type has no member X" in files one hop away from the file with the bad import. Suggest folding that into S2.1-B as a debugging note and dropping S4.1-C.

mdanylov-sigma (commented):
Review of the `PBMBidResponseTransformer` -> `BidResponseTransformer` port.

The bridge mechanics look right: the `@objc(PBMBidResponseTransformer)` + `@objc(transformResponse:error:)` pair reproduces the exact selector `PBMBidRequester.m:128` calls, `@_spi(PBMInternal)` does not hide it from ObjC (same shape as `PBMError`, which already appears in the generated header), the pbxproj edits are complete and land the new file in the right target's Sources phase, and `git grep PBMBidResponseTransformer` on the branch is clean apart from the test files noted inline.

The one I would push on is the S4.1-A reasoning at `BidResponseTransformer.swift:33` - I do not think that branch was dead in the way the new playbook rule generalizes.

One PR-scope note that does not fit inline: the description and `pr-phase-4-1.md` both present the playbook change as "three new gaps", but the diff rewrites about two thirds of the file (1050 -> 402 lines, -782/+134) and drops 20 of its 50 code blocks. I diffed the gap-ID sets and nothing is actually lost - only S4.1-A/B/C are new, and the 39-row orphan-header table is intact - so this is a condensation rather than a deletion. But a reviewer cannot establish that from the description, and a 358-line source change ends up buried under ~900 lines of unannounced doc churn. Worth either splitting it into its own PR or stating its scope explicitly in the "Playbook updates" section.

YuriyVelichkoPI:
Thanks for the thorough review. I went through all 7 comments against the current branch (diffed against `master` for the deleted ObjC to confirm what's actually new vs. carried over). Findings:

**Confirmed as genuine regression introduced by this port:**
- `BidResponseTransformer.swift:20` (lost `NS_UNAVAILABLE` enforcement) — the old header blocked ObjC callers at compile time; the new `private override init()` only restricts Swift callers, so `[[BidResponseTransformer alloc] init]` now silently compiles from ObjC. This is an actual migration-fidelity gap and I'll fix it (`@available(*, unavailable)` on the init) in this PR.

**Confirmed but pre-existing (verified via `git diff master...HEAD` / `git show master:...`), not introduced by this PR:**
- The "hollow `BidResponse`" nil-check removal (`:33`) — `BidResponse.swift` is untouched by this PR; it was already a non-failable Swift twin before this port, so the ObjC check was already dead code pre-PR. Real bug, but not a regression from this migration step.
- The `"Invalid request"` substring match (`:27`) and the `imp[0]`-hardcoded/dead-clause logic (`:45`) — both identical byte-for-byte in the deleted ObjC (`master:PrebidMobile/Objc/.../PBMBidResponseTransformer.m`). Ported verbatim, not changed by this PR.
- Out-param not cleared on success (`:25`) — a real, if currently latent, divergence from the `throws` translation. Worth a caller-side fix, but not exploitable with the current single call site.

**Doc/naming only, not functional:**
- Test files still carrying the `PBM` prefix, and S4.1-C duplicating S2.1-B.

Given the above, I'd like to set the scope for this migration series going forward:

**This migration should stay strictly 1:1.** The goal of each step is to move ObjC → Swift with identical behavior, not to fix bugs discovered along the way — even real ones. Where a comment identifies a genuine functional/logical issue that exists independently of the port (i.e., reproducible against `master` before this PR), please file it as a separate issue instead of a blocking review comment here — otherwise every migration PR turns into a bug bash for code that isn't changing. I've filed follow-ups for the pre-existing items above (hollow-response signal, `"Invalid request"` substring match, `imp[0]` hardcoding).

The one exception is the `NS_UNAVAILABLE` loss — that's a fidelity gap in the port itself (the ObjC guarantee existed and the Swift port dropped it), so it belongs in this PR, not a follow-up.

**Ask for future reviews on migration PRs:** please leave only (a) comments that require a change to make this specific port behavior-preserving, or (b) comments with a link to a follow-up issue for anything pre-existing/out-of-scope. That keeps the review focused on "did this PR change behavior" rather than "is the pre-migration behavior correct," and makes it easy to track what's blocking vs. deferred.

**On the playbook/doc notes** (S4.1-A/B/C, the duplicate-rule note, etc.): these are working notes for the migration effort itself, not permanent SDK documentation. Once the full migration lands and there's no more ObjC/Swift twin to reconcile, `playbook.md` and the `pr-phase-*.md` files go away entirely — there's no repo-wide guidance left to maintain at that point. So I'd rather not spend review cycles converging on "authoritative" vs. "duplicate" rule wording; overlapping entries can stay as-is or be marked `(info)` rather than treated as defects to resolve now.

mdanylov-sigma (commented):
Re-review of the current head (`6a5e651e`) under the strict 1:1 rule. This supersedes my earlier review — several of those comments were out of scope under that rule, and one was wrong.

**First, a correction that costs you a commit.** My comment on `BidResponseTransformer.swift:20` claimed `private override init()` dropped the header's `NS_UNAVAILABLE`. It does not, and `6a5e651e` was not needed. Swift emits an ObjC guard for either spelling:

```objc
// private override init()
- (nonnull instancetype)init SWIFT_UNAVAILABLE;
+ (nonnull instancetype)new SWIFT_UNAVAILABLE_MSG("-init is unavailable");

// @available(*, unavailable) override init()   <- current head
- (nonnull instancetype)init OBJC_DESIGNATED_INITIALIZER SWIFT_UNAVAILABLE;
```

I compiled `[[X alloc] init]` and `[X new]` against both, and against the original `NS_UNAVAILABLE` header: all six combinations are rejected at compile time (clang infers `+new`'s unavailability from `-init`). Keep the current spelling — no revert needed — but `Functions.swift:20` uses the `private` form and does **not** need the same change. Sorry for the detour. `pr-phase-4-1.md` still describes this line as `private override init()`, which is the only thing now stale about it.

## Blocking — 1

One inline comment, on the `throws` bridge at `BidResponseTransformer.swift:26`. It is the only finding whose provenance is the port itself.

## Deferred — 2, filed

Both are verbatim-ported logic or behavior reproducible on `master`, so they stay out of this PR:

- #1343 — `BidResponse` absorbs a failed ORTB deserialization, so `PBMError.responseDeserializationFailed()` has no reachable producer and a structurally invalid 200 arrives as a no-bid. **This withdraws my earlier comment on line 34**: asking you to add a `rawResponse != nil` guard was asking for a behavior change, not a port fix. Gap S4.1-A's conclusion is right for this file — `BidResponse.init(jsonDictionary:)` is non-failable on `master`, so the ObjC `if (!bidResponse)` was already unreachable there and dropping it is behavior-preserving.
- #1344 — `"Invalid request"` matched against the whole body before any structural check, plus the hardcoded `imp[0]` and its dead first clause in `classifyRequestError`. Byte-for-byte copies of the ObjC.

My earlier note that the test class is still named `PBMBidResponseTransformerTest` is cosmetic, not a defect — take it or leave it.

## Checked, no action needed

Recording these so you do not have to re-derive them:

- **`String.contains` vs `NSString containsString:`** — no divergence for these needles. Ran all six against haystacks carrying a trailing combining mark, an NBSP and an interior ZWJ; Swift and `NSString` agree on every case, because the needles are pure ASCII and `rangeOfString:` is composed-character-sequence aware. The hazard from #1336 does not apply here.
- **Nonnull parameter** — `PrebidServerResponseCallback` is `(PrebidServerResponse) -> Void`, already non-optional on `master`, so the non-optional Swift parameter cannot trap where the ObjC degraded. #1336's other hazard does not apply either.
- **Nil / non-UTF-8 `rawData`** — ObjC produced a nil `responseBody` and `containsString:` on nil returned `NO`; Swift's `?? ""` reaches the same branch. Equivalent.
- **Selector and nullability** — `@objc(transformResponse:error:)` carries the error label, and the emitted method matches the deleted header exactly: `+ (BidResponse * _Nullable)transformResponse:(PrebidServerResponse * _Nonnull)response error:(NSError * _Nullable * _Nullable)error`.
- **Mechanics** — pbxproj edits complete; no orphaned `#import`s; `classifyRequestError` had no callers outside the file; every test file touching the type carries `@_spi(PBMInternal)` on its import.
- **Playbook** — diffed the gap-ID sets across the change: 1050 -> 402 lines, no ID lost, only S4.1-A/B/C added. Worth stating that scope in the description, but not a defect.

## Verification on this head

The test-plan checkboxes were ticked at `599384c7`, one commit before the current head, so I re-ran them:

- `./scripts/buildPrebidMobilePackage.sh` — succeeded
- `./scripts/testPrebidMobile.sh --latest --quick` — **812 tests, 0 failures**

PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCore/BidResponseTransformer.swift
```diff
@@ -0,0 +1,51 @@
...
+
+@objc(PBMBidResponseTransformer) @_spi(PBMInternal) public class BidResponseTransformer: NSObject {
+
+    @available(*, unavailable)
+    override init() {
+        super.init()
+    }
+
+    @objc(transformResponse:error:)
+    public static func transform(_ response: PrebidServerResponse) throws -> BidResponse {
```
mdanylov-sigma, line 26: The ObjC cleared the out-parameter on the success path:

```objc
if (error) { *error = nil; }
return bidResponse;
```

The compiler-generated `throws` thunk does not — it writes `*error` only when the call throws. Built this exact signature and called it from ObjC:

```
success path -> result=ok, err=STALE      <- caller's pre-existing error survives
throw   path -> result=(nil), err=thrown
```

That is a guarantee the ObjC method gave and the port cannot express, so it is in scope under the 1:1 rule even though nothing is broken today. `PBMBidRequester.m:127` declares `NSError *trasformationError = nil;` immediately before the call, so the `bidResponse && !trasformationError` test on line 130 is still safe — but that conjunct now rests on caller discipline rather than on this method. A reused error variable breaks it:

```objc
NSError *err = nil;
[A doThing:&err];                                                               // sets err
BidResponse *r = [PBMBidResponseTransformer transformResponse:resp error:&err]; // succeeds, err untouched
if (r && !err) { ... }                                                          // stale err -> failure path with a non-nil response
```

Two ways to close it, both small: drop the `!trasformationError` conjunct at `PBMBidRequester.m:130` and branch on `bidResponse` alone (the Cocoa contract anyway), or record the divergence in `pr-phase-4-1.md` next to S4.1-A so it is deliberate and on the record.

mdanylov-sigma (approved):
LGTM!

PR #1341, opened 2026-09-08
Multiformat Banner Ad Unit in the Rendering and Mediation APIs

### LINKED ISSUE #1340: Multiformat Banner Ad Unit in the Rendering and Mediation APIs
# Background

`BannerView` and `MediationBannerAdUnit` accept a single `adFormat`. A banner placement therefore competes for either display or outstream video demand, never both.

All other ad units already support multiformat requests via `adFormats`. The banner renderer already picks the creative from the winning bid's `ext.prebid.type`, so multiformat banners need no rendering or server changes.

# Proposed changes

- Add `adFormats: Set<AdFormat>` to `BannerView` and `MediationBannerAdUnit`; default stays `[.banner]`.
- Deprecate `adFormat` as a shim over `adFormats`.
- With `[.banner, .video]`, the impression carries both `imp.banner` and `imp.video`, as interstitials do today.

OlenaPostindustria:
Closes #1340


PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/BannerView.swift
```diff
@@ -64,7 +64,21 @@ public class BannerView:
         set { adUnitConfig.additionalSizes = newValue }
     }
     
+    /// The set of ad formats requested for this ad unit.
+    ///
+    /// Defaults to `[.banner]`. Set `[.video]` for an outstream video banner, or
+    /// `[.banner, .video]` for a multiformat request.
+    ///
+    /// Only `.banner` and `.video` can be rendered by `BannerView`.
```
mdanylov-sigma, line 72: The doc says only `.banner` and `.video` render, but nothing enforces it — `adFormats = [.banner, .native]` is accepted.

`BasicParameterBuilder.appendFormatSpecificParameters` then adds `imp.native`, `includeformat` goes out, and PBS can legitimately return a native bid. `Bid.adFormat` resolves it to `AdFormat.native`, and then `PBMCreativeFactoryJob.m:126-131`:

```objc
if (adType == AdFormat.banner || self.creativeModel.isCompanionAd) {
    [self attemptAUIDCreative];
} else if (adType == AdFormat.video) {
    [self attemptVASTCreative];
} else if (adType == nil) {
    PBMLogError(@"The winning bid ad format is nil.")
}
```

`.native` matches no branch — the completion is never invoked and no error is logged. The ad only fails later via the generic factory timeout ("Failed to complete in specified time interval"), which gives no clue about the real cause.

The old single-value `adFormat` was equally permissive, but a `Set` invites mixing formats in a way a scalar never did. Worth filtering unsupported formats in the setter (with a `Log.warn`), or at least adding the `.native` case to the factory's dispatch. Same applies to `MediationBannerAdUnit.adFormats`.

PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/BannerView.swift
```diff
@@ -64,7 +64,21 @@ public class BannerView:
...
+    public var adFormats: Set<AdFormat> {
+        get { adUnitConfig.adFormats }
+        set { adUnitConfig.adFormats = newValue }
```
mdanylov-sigma: Nothing rejects an empty set. `bannerView.adFormats = []` passes straight through to `adConfiguration.adFormats`, `appendFormatSpecificParameters` adds nothing, and `PBMPrebidParameterBuilder`'s per-format loop iterates zero times — so the request ships a bare `imp` with no `banner`, `video` or `native` object and PBS rejects it.

The deprecated getter still reports `.banner` (`first ?? .banner`), so the misconfiguration is invisible from the API surface.

Unreachable with the old scalar property. Suggest clamping an empty set to `[.banner]` with a warning, same as the setter-side filtering for unsupported formats.

PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/BannerView.swift
```diff
@@ -64,7 +64,21 @@ public class BannerView:
         set { adUnitConfig.additionalSizes = newValue }
     }
     
+    /// The set of ad formats requested for this ad unit.
+    ///
+    /// Defaults to `[.banner]`. Set `[.video]` for an outstream video banner, or
```
mdanylov-sigma, line 69: A publisher who follows this line and nothing else sends `imp.video` with no `placement` and no `plcmt`.

Compare the interstitial units, which default both in `init`:

```swift
// BaseInterstitialAdUnit.swift:115-116, MediationBaseInterstitialAdUnit.swift:83-84
videoParameters.placement = .Interstitial
videoParameters.plcmnt = .Interstitial
```

`BannerView.init` and `MediationBannerAdUnit.init` set only `bannerParameters.api`. Bidders that require the outstream signal will drop the video imp, so the unit quietly gets display-only demand and multiformat looks like it does nothing.

Either default to `.InBanner` / `.AccompanyingContent` in the banner initializers the way the interstitials do, or say explicitly in this doc comment that `videoParameters.placement` must be set.

PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/BannerView.swift
```diff
@@ -64,7 +64,21 @@ public class BannerView:
...
+        get { adUnitConfig.adFormats }
+        set { adUnitConfig.adFormats = newValue }
+    }
+    
     /// The ad format (e.g., banner, video).
+    ///
+    /// - Note: Deprecated. Use `adFormats` instead, which supports multiformat requests.
+    @available(*, deprecated, message: "Use `adFormats` instead.")
     public var adFormat: AdFormat {
         get { adUnitConfig.adFormats.first ?? .banner }
```
mdanylov-sigma, line 91: `Set.first` is not stable here. `AdFormat` is an `NSObject` subclass with no `hash`/`isEqual:` override, so `Set<AdFormat>` orders by pointer identity combined with Swift's per-process random hash seed.

With `adFormats = [.banner, .video]`, `bannerView.adFormat` returns `.banner` on one launch and `.video` on the next. Existing publisher code doing `if bannerView.adFormat == .video { ... }` becomes a coin flip the moment the unit is moved to multiformat — which is exactly the migration this shim exists to make safe.

Suggest making it deterministic, e.g. `adUnitConfig.adFormats.contains(.video) ? .video : .banner`.

Related: `BannerViewTest.swift:120` currently asserts only `XCTAssertTrue(bannerView.adFormats.contains(bannerView.adFormat))`, which documents the nondeterminism rather than pinning it. Same for `MediationBannerAdUnitTest.swift:108`.

PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/MediationAPI/MediationBannerAdUnit.swift
```diff
@@ -46,7 +46,21 @@ public class MediationBannerAdUnit : NSObject {
...
     
+    /// The set of ad formats requested for this ad unit.
+    ///
+    /// Defaults to `[.banner]`. Set `[.video]` for an outstream video banner, or
+    /// `[.banner, .video]` for a multiformat request.
+    ///
+    /// Only `.banner` and `.video` can be rendered by the banner mediation adapters.
+    public var adFormats: Set<AdFormat> {
+        get { adUnitConfig.adFormats }
+        set { adUnitConfig.adFormats = newValue }
```
mdanylov-sigma: Now that `Set<AdFormat>` is the headline API on the banner classes, `AdFormat`'s equality semantics become a footgun worth guarding against.

`AdFormat` is `NSObject, OptionSet` with no `isEqual:`/`hash` override, so both `Set` membership and `PBMPrebidParameterBuilder.m:181/209/292` (`adFormat == AdFormat.banner`) compare by *pointer*. Only the three static singletons work.

Two natural ways for a publisher to get this wrong:

```swift
adUnit.adFormats = [AdFormat.banner.union(.video)]   // OptionSet.union builds a fresh AdFormat(rawValue: 3)
adUnit.adFormats = [AdFormat(rawValue: 1), AdFormat(rawValue: 2)]  // required init(rawValue:) is public
```

In both cases no element matches `AdFormat.banner` or `AdFormat.video`, so the imp ships with neither `banner` nor `video` — while `adFormats.count >= 2` still sets `includeformat`. Silent failure, no SDK-side diagnostic.

Overriding `isEqual:`/`hash` on `AdFormat` in terms of `rawValue` would make all of this behave the way the type's `OptionSet` conformance already advertises. Pre-existing, but this PR is what puts `Set<AdFormat>` in front of publishers.

InternalTestApp/PrebidMobileDemoRendering/ViewControllers/Adapters/Prebid/OpenX/PrebidBannerController.swift
```diff
@@ -84,10 +84,11 @@ class PrebidBannerController: NSObject, AdaptedController, PrebidConfigurableBan
         if adSizes.count > 1 {
             adBannerView?.additionalSizes = Array(adSizes.suffix(from: 1))
         }
-        if let adFormat = adFormat {
-            adBannerView?.adFormat = adFormat
+        
+        if let adFormats {
+            adBannerView?.adFormats = adFormats
```
mdanylov-sigma, line 89: Ordering bug: this silently discards the `refreshInterval` set five lines above.

`AdUnitConfig.adFormats` has a `didSet` that calls `updateAdFormat()`:

```swift
private func updateAdFormat() {
    if adConfiguration.adFormats == adFormats { return }
    self.adConfiguration.adFormats = adFormats
    self.refreshInterval = (adConfiguration.winningBidAdFormat == .video) ? 0 : refreshIntervalDefault
}
```

`winningBidAdFormat` is still `nil` at configuration time, so any format change resets `refreshInterval` to 60. Line 80 assigns the tester's configured interval, line 89 throws it away — with only a debug log. Because the early-return compares sets, this fires for every `.video` and `[.banner, .video]` case and never for a plain `[.banner]` one.

Same order in `PrebidGAMBannerController.swift:82`, `PrebidAdMobBannerViewController.swift:120` and `PrebidMAXBannerController.swift:105`, so the configured refresh interval is ignored on every video/multiformat test case.

Swapping the two blocks fixes the test app, but the underlying SDK behaviour is the real problem: `adFormats` is now public API, so `bannerView.refreshInterval = 30; bannerView.adFormats = [.banner, .video]` silently gives publishers 60. Either `updateAdFormat()` should stop touching `refreshInterval` unless it was never explicitly set, or the new `adFormats` doc comment should state the ordering requirement.

InternalTestApp/PrebidMobileDemoRendering/Model/TestCasesManager.swift
```diff
@@ -2010,13 +2010,33 @@ struct TestCaseManager {
...
+                     tags: [.banner, .video, .inapp, .server],
+                     exampleVCStoryboardID: "AdapterViewController",
+                     configurationClosure: { vc in
+                guard let adapterVC = vc as? AdapterViewController else {
+                    return
+                }
+                         
+                let bannerController = PrebidBannerController(rootController: adapterVC)
+                let configIds = ["prebid-ita-banner-300-250", "prebid-ita-video-outstream"]
+                bannerController.prebidConfigId = configIds.randomElement()!
```
mdanylov-sigma, line 2032: Picking randomly between a display-only and a video-only stored request means no run ever exercises a stored impression that can actually return *either* format — which is the case that tests the `includeformat` / `ext.prebid.type` branch this feature depends on.

It also makes failures hard to reproduce: each launch pins the outcome, so a tester who sees a bug has to relaunch until the dice land the same way, with nothing on screen saying which config was used.

Suggest either a genuinely multiformat stored request (one config id that can serve both), or two explicitly named deterministic cases. Same pattern at lines 2183 (GAM), 2951 (AdMob) and 3754 (MAX).

PrebidMobileTests/RenderingTests/Tests/ParameterBuilderTests/PrebidParameterBuilderTest.swift
```diff
@@ -810,6 +810,71 @@ class PrebidParameterBuilderTest: XCTestCase {
...
     }
 
+    func testMultiformatBannerView() {
+        let size = CGSize(width: 300, height: 250)
+        let bannerView = BannerView(frame: CGRect(origin: .zero, size: size), configID: "configID", adSize: size)
+        bannerView.adFormats = [.banner, .video]
+        
+        let bidRequest = buildBidRequest(with: bannerView.adUnitConfig)
+        
+        // PBS must echo hb_format so the SDK knows which creative type won
```
mdanylov-sigma: This comment is inaccurate and could send someone the wrong way later. The SDK never reads `hb_format` — `Bid.adFormat` (`Bid.swift:78`) resolves the creative type from `bid.ext.prebid.type`:

```swift
public var adFormat: AdFormat? {
    AdFormat.allCases.filter { $0.stringEquivalent == bid.ext?.prebid?.type }.first
}
```

Grepping the SDK and all three adapter targets, `hb_format` appears only in this comment and in the `includeformat` comment in `PBMPrebidParameterBuilder.m`. `includeformat` exists so the *ad server* line items can key on `hb_format` — it's not part of the rendering decision.

Suggest: `// includeformat makes PBS return hb_format so ad server line items can target the winning format`.

InternalTestApp/PrebidMobileDemoRendering/ViewControllers/Adapters/Prebid/OpenX/CustomRenderer/CustomRendererBannerController.swift
```diff
@@ -89,7 +89,7 @@ class CustomRendererBannerController:
             adBannerView?.additionalSizes = Array(adSizes.suffix(from: 1))
         }
         if let adFormat = adFormat {
-            adBannerView?.adFormat = adFormat
+            adBannerView?.adFormats = [adFormat]
```
mdanylov-sigma: This is the one banner controller left on the singular `adFormat: AdFormat?` (line 30) — the other four (`PrebidBannerController`, `PrebidGAMBannerController`, `PrebidAdMobBannerViewController`, `PrebidMAXBannerController`) all moved to `adFormats: Set<AdFormat>?` in this PR.

Consequence: the plugin-renderer path can't be pointed at a multiformat banner, so custom renderers are the one integration the new feature can't be manually tested against. Anyone adding such a case has to migrate this controller first.

Worth converting here for consistency: `var adFormats: Set<AdFormat>?` with `adFormats.contains(.video)` for the placement branch, matching `PrebidBannerController`.

Example/PrebidDemo/PrebidDemoSwift/Examples/In-App/InAppVideoBannerViewController.swift
```diff
@@ -35,7 +35,7 @@ class InAppVideoBannerViewController: BannerBaseViewController, BannerViewDelega
         
         // 2. Configure the BannerView
         prebidBannerView.delegate = self
-        prebidBannerView.adFormat = .video
+        prebidBannerView.adFormats = [.video]
```
mdanylov-sigma, line 38: The demo app got only the mechanical `adFormat` -> `adFormats` renames — no multiformat banner example was added, in either the Swift or the ObjC demo.

Worth adding one, for two reasons:

1. The Original API already ships the equivalent — `IntegrationCaseManager.m:96-104` lists "GAM (Original API) Multiformat Banner 300x250" with `AdFormatMultiformat`. The rendering/mediation API is now the odd one out.
2. This file is the only place in the repo that shows the `adFormats` + `videoParameters.placement` pairing that a multiformat banner actually needs (see my comment on `BannerView.swift`). The demo app is what publishers copy from, so it's where that pairing should be demonstrated for the multiformat case too.

Four InternalTestApp cases and zero demo cases leaves the documented integration path unexercised by the demo build.

mdanylov-sigma (commented):
Review of the multiformat banner change. The request-building side is solid and the new parameter-builder tests confirm it, but a few things on the **rendering** side aren't as ready as #1340 assumes, plus some missing validation on the new public setter. Inline comments below; two items land in files this PR doesn't touch, so they're here:

---

### 1. `isBuiltInVideo` latches to `true` and is never cleared (blocker)

`BannerView.adUnitConfig` is a *single* instance: it goes to `AdLoadFlowController.savedAdUnitConfig` and from there to every `DisplayView` via `PluginRendererFactory.createBannerView` — no copy anywhere on that path.

`DisplayView.display(transaction:)` does:

```swift
if adConfiguration.adConfiguration.winningBidAdFormat == .video {
    adConfiguration.adConfiguration.isBuiltInVideo = true
}
```

...and nothing ever sets it back to `false`.

So on a `[.banner, .video]` unit with the default 60s auto-refresh: auction 1 returns video -> `isBuiltInVideo = true`. Auction 2 returns an HTML banner -> `loadAd()` updates `winningBidAdFormat` to `.banner`, but `isBuiltInVideo` is still `true`. `PBMHTMLCreative` fires `creativeDidComplete` after its display timer, and `PBMAdViewManager.m:187/201` then skips both `removeFromSuperview` and `adDidComplete` because the flag says built-in video.

Before this PR a banner unit was pinned to one format for its whole life, so the flag could never go stale. Multiformat is the first thing that makes it reachable. Suggest resetting it in `DisplayView.loadAd()` alongside `winningBidAdFormat`.

---

### 2. Auto-refresh isn't suppressed when a video creative wins (blocker)

The only guard is in `AdUnitConfig.refreshInterval`'s setter:

```swift
if adConfiguration.winningBidAdFormat == .video {
    Log.warn("'refreshInterval' property is not assignable for Outstream Video ads")
    _refreshInterval = 0
    return
}
```

But that only runs on a *write*, and the last write is `updateAdFormat()` at configuration time, when `winningBidAdFormat` is still `nil`. `DisplayView.loadAd()` sets it to `.video` afterwards and nobody writes `refreshInterval` again — so `_refreshInterval` stays at 60.

`BannerView.mayRefreshNow` doesn't help: it only blocks on `isCreativeOpened`, which `PBMAdViewManager.isCreativeOpened` defines as clickthrough-browser-visible or MRAID-expanded — not video playback.

Net effect: a `[.banner, .video]` banner that wins a 30s outstream video gets torn down and reloaded mid-playback at 60s. This was technically reachable before for a deliberately video-only unit, but multiformat makes it the default path for any display banner that opts into video demand, so it's worth fixing here.

---

Everything else is inline. Nice catch on the dead `randomId` in the MAX multiformat interstitial case.

PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/BannerView.swift
OlenaPostindustria: `placement/plcmt` are optional in ORTB. The banner units never defaulted them, so this is pre-existing behavior rather than something this PR changes.

PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/BannerView.swift
OlenaPostindustria: The SDK never reads the deprecated `adFormat` getter. All internal reads go through `AdUnitConfig.adFormats`, so this can't affect SDK behaviour.

PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/BannerView.swift
mdanylov-sigma: yeap, sorry, too strange AI review, missed it 😭

InternalTestApp/PrebidMobileDemoRendering/Model/TestCasesManager.swift
OlenaPostindustria: This is intentional. There is no stored request on the test server that can return either format, so the random pick between a display-only and a video-only stored request is how the test app simulates the server choosing the winning format.

Example/PrebidDemo/PrebidDemoSwift/Examples/In-App/InAppVideoBannerViewController.swift
OlenaPostindustria: I would not add these cases to the public demo apps. Prebid Server can't return a random format from a single stored impression, so the only way to demo both winning formats is a client-side random pick between two stored requests. That is a test harness trick, not an integration pattern publishers should copy.

PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/BannerView.swift
```diff
@@ -469,6 +496,12 @@ extension BannerView : AdLoadFlowControllerDelegate, BannerAdLoaderDelegate {
         loadedAdView adView: UIView,
         adSize: CGSize
     ) {
+        // The refresh timer is armed before the primary ad server answers, so the winning
+        // format can only be honoured here. A video creative must not be torn down mid-playback.
+        if lastBidResponse?.winningBid?.adFormat == .video {
```
mdanylov-sigma: This cancels auto-refresh even when Prebid lost the ad server auction and a GAM creative is what's actually on screen.

`lastBidResponse` is `adLoadFlowController?.bidResponse` (line 46), assigned once per cycle in `handleBidResponse` (`AdLoadFlowController.swift:178`) and never cleared when the ad server wins. And this callback is the single success path for *both* outcomes:

```swift
// AdLoadFlowController.swift:236
private func deployPendingViewAndSendSuccessReport() {
    flowState = .idle
    guard let adObject = primaryAdObject ?? prebidAdObject else { return }
    adLoader?.reportSuccess(with: adObject, adSize: adSize)
}
```

`BannerAdLoader.reportSuccess` forwards whichever object it got into `bannerAdLoader(_:loadedAdView:)`.

So on a GAM integration:

1. Prebid auction returns a video bid, `winningBid.adFormat == .video`.
2. GAM serves its **own** creative, because no `hb_pb` line item matched or the app event timed out. Both land on `loadingDelegate?.adServerDidWin(...)` (`GAMBannerEventHandler.swift:214` and `:251`).
3. That routes to `adLoader(_:loadedPrimaryAd:adSize:)`, sets `primaryAdObject`, deploys.
4. This check sees `.video` on the *losing* Prebid bid and cancels the timer.

A GAM HTML banner is on screen with auto-refresh silently dead. The app-event-timeout path makes this routine rather than exotic.

Worth noting this isn't limited to multiformat: a video-only banner (`adFormats = [.video]`, supported today) hits it too, and that banner kept refreshing before this PR. So as written this is a behaviour regression for existing video banner publishers on GAM. That reading is from the call graph rather than an observed run, so worth confirming against a real GAM test case before you act on it.

Gating on what is actually rendering avoids the whole class of problem. `adView is DisplayView` is the direct signal, since the Prebid renderer is the only thing that produces one. See my note on the next line for the shape I'd suggest overall.

PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/GAM/BannerView.swift
```diff
@@ -469,6 +496,12 @@ extension BannerView : AdLoadFlowControllerDelegate, BannerAdLoaderDelegate {
...
+            autoRefreshManager?.cancelRefreshTimer()
```
mdanylov-sigma: Second issue on the same block: this cancel is permanent, so one video win turns the slot static for good.

`cancelRefreshTimer` cancels the pending work item and nils `delayedBlock`. The only thing that re-arms it is:

```swift
// BannerView.swift:482
public func adLoadFlowControllerWillRequestPrimaryAd(_ adLoadFlowController: AdLoadFlowController) {
    autoRefreshManager?.setupRefreshTimer()
    ...
}
```

Which is reached only from `requestPrimaryAdServer`, only from `moveToNextLoadingStep` in `.demandReceived`, only on a new load cycle. New cycles start from `adLoadFlowController.refresh()`, called by the refresh timer's own `refreshBlock` or by the public `loadAd()`. Cancel the timer and nothing schedules the next cycle, so nothing re-arms the timer. The loop is broken for the life of the view unless the publisher calls `loadAd()` by hand.

For a video-only unit that may be acceptable. For the `[.banner, .video]` unit this PR exists to enable, one video win kills refresh even though the next twenty auctions would return HTML.

### Suggested shape (fixes this and the ad-server issue above)

`AutoRefreshManager` already has a "not now, ask me again later" path:

```swift
// AutoRefreshManager.refresh()
if !isBackground, mayRefreshNowBlock() {
    refreshBlock()
} else {
    Log.info("Creative is invisible or opened. Skipping refresh.")
    setupRefreshTimer()   // re-arms itself
}
```

And `BannerView.mayRefreshNow` is already where "don't reload right now" lives, next to `isAdOpened`, `!isVisible()` and `isCreativeOpened`. A video creative in flight belongs in that list.

Moving the condition there fixes both problems as a consequence of being evaluated at the right time against the right state:

- it reads what is currently deployed rather than the bid response, so an ad server win never triggers it;
- it self-recovers, since nothing is cancelled and the next tick just proceeds.

That last point matters: re-arming from `videoPlaybackDidComplete` would be the obvious alternative, but if the video errors out or the user scrolls away that callback may never arrive and refresh would stay dead. A condition doesn't have that failure mode.

PrebidMobile/Swift/PrebidMobileRendering/Prebid/Integrations/MediationAPI/MediationBannerAdUnit.swift
```diff
@@ -46,10 +46,33 @@ public class MediationBannerAdUnit : NSObject {
         adUnitConfig.configId
     }
     
+    /// The set of ad formats requested for this ad unit.
+    ///
+    /// Defaults to `[.banner]`. Set `[.video]` for an outstream video banner, or
+    /// `[.banner, .video]` for a multiformat request.
+    ///
+    /// Only `.banner` and `.video` can be rendered by the banner mediation adapters.
+    public var adFormats: Set<AdFormat> {
```
mdanylov-sigma, line 55: This ad unit gains multiformat here, but not the video auto-refresh protection that `BannerView` got, so AdMob and MAX multiformat banners will interrupt a video creative mid-playback.

`completeWithResult` arms the timer unconditionally:

```swift
// MediationBannerAdUnit.swift:305
autoRefreshManager?.setupRefreshTimer()
```

No reference to `response.winningBid?.adFormat`, and `mayRefreshNowBlock` only checks visibility and error state. `refreshDelayBlock` returns `adUnitConfig.refreshInterval`, which defaults to 60. So a `[.banner, .video]` unit that wins a 30s outstream video re-runs `fetchDemand` at 60s, the publisher's completion handler reloads the mediation view, and the creative is swapped mid-playback.

To be clear about what changed and what didn't: this is **not** a regression from removing the guard in `AdUnitConfig.refreshInterval`. That guard could never have fired here anyway, for two independent reasons. Both adapters build a fresh `AdUnitConfig` at render time (`PrebidAdMobBannerAdapter.swift:83`, `PrebidMAXMediationAdapter+MAAdViewAdapter.swift:61`), so `DisplayView` sets `winningBidAdFormat` on *that* object and never on this ad unit's own config, which is the one driving the timer. And even if it did, the guard only ran on a write to `refreshInterval`, which nothing does after a win. Removing it was correct.

The point is just that the replacement landed on only one of the two ad units this PR makes multiformat. The winning bid is already in hand in `handlePrebidResponse` as `response.winningBid`, so the check is cheap to add.

Not a blocker on correctness grounds since nothing regresses, but shipping the protection for `BannerView` only means the mediation half lands with a known creative-interruption bug. A follow-up issue would be fine if you'd rather keep this PR tight.

mdanylov-sigma (commented):
Follow-up on the fixes pushed since my last review. The core corrections all look right to me: `isBuiltInVideo` derived per load, `AdFormat` given real value equality with good test coverage, `updateAdFormat` removed entirely (which also fixes `InterstitialRenderingAdUnit.adFormats` reporting `[.banner]` before any set), and the validation helper is cleanly factored.

The auto-refresh fix is the one I'd like to revisit. Two issues inline, and they share a single root cause: the suppression is expressed as a one-shot cancel at load time, when it wants to be a condition evaluated at refresh time. Details on the suggested shape are in the comment on `cancelRefreshTimer`.

---

### One more gap: `RewardedAdUnit.adFormats` was missed by the validation pass

`RewardedAdUnit.swift:33` still assigns straight through:

```swift
public var adFormats: Set<AdFormat> {
    get { adUnitConfig.adFormats }
    set { adUnitConfig.adFormats = newValue }
}
```

The other four public setters now route through `AdFormat.validated(_:supported:)`: `BannerView:76`, `MediationBannerAdUnit:60`, `InterstitialRenderingAdUnit:39`, `MediationInterstitialAdUnit:35`. So both failures the validation was added to close are still reachable on this one: `[]` produces an impression with no format object, and `[.native]` reaches the unhandled branch in `PBMCreativeFactoryJob.m:126-131` where no case matches and the load only fails on the factory timeout.

(`MediationRewardedAdUnit` is fine, it sets `[.video]` internally rather than exposing a setter.)

Low priority and arguably outside the scope of #1340. Neither failure is a regression, both were reachable before this PR too. Equally happy to see it as a follow-up, just noting it so the "every `adFormats` setter validates" invariant is either true or knowingly not.

mdanylov-sigma (dismissed):
LGTM!

mdanylov-sigma (approved):
LGTM

Issue #1324, opened 2026-08-26
Questions regarding multiformat MediationBannerAdUnit, TFUA, ORTB config, and GPID


jaysongkkim:
## Background

We are currently integrating the Prebid iOS SDK with our mediation setup and have several questions regarding the recommended implementation. We are using **PrebidMobile 3.3.1**, installed via CocoaPods.

The following observations are based on the CocoaPods source. The root path is `PrebidMobile/Swift/`. The directory structure may differ in the GitHub repository, but the relevant symbols and implementations should be the same.

We would appreciate your guidance on the following points.

### 1. Can `MediationBannerAdUnit` expose multiformat (`adFormats`) configuration?
Our banner inventory supports multiple formats, such as banner and video. Currently, `MediationBannerAdUnit` exposes only a single `adFormat`:

```swift
// PrebidMobileRendering/Prebid/Integrations/MediationAPI/MediationBannerAdUnit.swift:50

public var adFormat: AdFormat {
    get { adUnitConfig.adFormats.first ?? .banner }
    set { adUnitConfig.adFormats = [newValue] }
}
```

Setting adFormat therefore replaces the existing formats with a single format. However, the underlying AdUnitConfig already supports multiple formats:

```swift
// PrebidMobileRendering/Prebid/PBMCore/AdUnitConfig.swift:32

public var adFormats: Set<AdFormat>
}
```
Also, MediationInterstitialAdUnit already exposes the full set:
```swift
// PrebidMobileRendering/Prebid/Integrations/MediationAPI/MediationInterstitialAdUnit.swift:27

public var adFormats: Set<AdFormat> {
    get { adUnitConfig.adFormats }
    set { adUnitConfig.adFormats = newValue }
```
There are also existing examples where multiple formats are configured internally:
```swift
// PrebidMobileRendering/Prebid/Integrations/GAM/BaseInterstitialAdUnit.swift:113

adUnitConfig.adConfiguration.adFormats = [.banner, .video]

}
```
In addition, MediationBannerAdUnit already contains both banner and video parameters:
```swift
// MediationBannerAdUnit.swift:62
public var bannerParameters: BannerParameters
// MediationBannerAdUnit.swift:67
public var videoParameters: VideoParameters

```

Based on this, it appears that the underlying SDK architecture already supports multiple formats, while the MediationBannerAdUnit API currently exposes only a single format.

[...]

PR #1342, opened 2026-09-08
fix: map BidInfo imp event to ext.prebid.events.imp

mdanylov-sigma:
## Problem

`BidInfo.create` reads `events.win` for **both** event keys:

```swift
if let winURL = bidResponse.winningBid?.events?.win {
    bidInfo.addEvent(key: BidInfo.EVENT_WIN, value: winURL)
}

if let impURL = bidResponse.winningBid?.events?.win {   // <- should be .imp
    bidInfo.addEvent(key: BidInfo.EVENT_IMP, value: impURL)
}
```

`ORTBExtPrebidEvents` parses both `win` and `imp` correctly, so the data is present on the bid. `BidInfo` simply reads the wrong field. Any publisher reading `bidInfo.events[BidInfo.EVENT_IMP]` from a `fetchDemand` completion gets the win notice URL and fires the wrong tracker.

The slip dates to the original feature commit `62a196cd` ("feat: surface win & imp events via SDK", 2023-10-11) and has shipped in every release since 2.1.6. Android is unaffected: its `BidInfo` assigns `winningBid.getEvents()` as a whole map rather than field by field.

## Why the existing test missed it

`testBidInfoCompletion` asserted only `XCTAssertNotNil` on both keys, which passes with the bug. Its fixture also used the same URL for `win` and `imp`:

```json
"events": { "win": "https://prebid.org/", "imp": "https://prebid.org/" }
```

so even an equality assertion would not have caught it.

## Changes

- `BidInfo.create` now reads `events.imp` for `BidInfo.EVENT_IMP`.
- `sample_ortb_native_with_win_event.json` gets distinct win and imp URLs.
- `testBidInfoCompletion` asserts the exact URL per key instead of non-nil.

## Test plan

- [x] `testBidInfoCompletion` **fails** on the old mapping once the fixture carries distinct URLs, confirming the new assertion is a real regression guard
- [x] `testBidInfoCompletion` passes with the fix
- [x] Full `AdUnitTests` class green (16/16)
- [x] PR subset green: `./scripts/testPrebidMobile.sh --latest --quick`, 0 failures

The fixture's other consumer, `testFetchDemandBidsAutoRefreshWithSimilarGlobalTimeout`, does not assert on event URLs, so the fixture change is safe.

Found while researching #1324.


YuriyVelichkoPI (approved):
LGTM

Issue #1344, opened 2026-09-09
Prebid Server error classification matches substrings against the whole response body: false positives on valid bids, and imp[0] hardcoded


mdanylov-sigma:
Split out of the review of #1337 as **pre-existing** — the Swift port in that PR is a verbatim copy of `PBMBidResponseTransformer.m` on `master`. The Swift-migration PRs are strictly 1:1, so this does not belong in one of them.

### 1. `"Invalid request"` is matched against the whole decoded body before any structural check

```objc
NSString * const responseBody = [[NSString alloc] initWithData:response.rawData encoding:NSUTF8StringEncoding];
if ([responseBody containsString:@"Invalid request"]) {
    if (error) { *error = [self classifyRequestError:responseBody]; }
    return nil;
}
```

A well-formed 200 bid response whose creative markup or bidder-ext strings happen to contain the literal — say `"adm": "<html><div>Invalid request format detected</div></html>"` — parses fine as JSON but is classified as a server error. `PBMBidRequester` then delivers `(nil, PBMError.serverError(<entire body>))`: the whole auction is dropped, and the response body is echoed into the error's `userInfo`.

A real Prebid Server `Invalid request:` reply is a non-JSON plaintext body, so gating the check on `response.jsonDict == nil` first removes the false positive without changing behavior for any genuine error response. Worth noting the existing suite would not catch a change here either way: every error fixture in `PBMBidResponseTransformer+TestExtension.swift` (`invalidAccountIDResponse`, `invalidConfigIdResponse`, `invalidSizeResponse`, `serverErrorResponse`) builds a plaintext body, so `jsonDict` is nil on all of them.

### 2. `classifyRequestError:` hardcodes `imp[0]`, and its first clause is dead

```objc
if ([responseBody containsString:@"Invalid request: Request imp[0].banner.format"]
    || [responseBody containsString:@"Request imp[0].banner.format"]
    || [responseBody containsString:@"Unable to set interstitial size list"])
```

- Any string containing the first needle also contains the second, so the first clause can never be the deciding one.
- A multi-impression request rejected with `Invalid request: Request imp[1].banner.format should define ...` matches neither and is reported as `prebidServerError` instead of `prebidInvalidSize`. The test helper `invalidSizeResponse(impIndex:formatIndex:)` is parameterized on the imp index, but every call site passes `impIndex: 0`, so the gap is invisible in the suite.

[...]

PR #1346, opened 2026-09-11
fix: report ORTB deserialization failure instead of a hollow no-bid (#1343)

### LINKED ISSUE #1343: BidResponse silently absorbs a failed ORTB deserialization; PBMError.responseDeserializationFailed() has no reachable producer
Split out of the review of #1337 as **pre-existing** — reproducible on `master` before that PR. The Swift-migration PRs are strictly 1:1, so this does not belong in one of them.

### What happens

`BidResponse.init(jsonDictionary:)` is a **non-failable** `convenience init`:

```swift
public convenience init(jsonDictionary: [String : Any]) {
    let rawResponse = RawBidResponse(jsonDictionary: jsonDictionary, ...)
    self.init(rawBidResponse: rawResponse)
}
```
`PrebidMobile/Swift/PrebidMobileRendering/Prebid/PBMCore/BidResponse.swift:40-54`

`RawBidResponse` is `ORTBBidResponse<…>`, whose JSON initializer *is* failable and returns nil whenever the body has no top-level `"id"`:

```swift
convenience required init?(jsonDictionary: [String : Any]) {
    let json = JSONObject<KeySet>(jsonDictionary)
    guard let requestID: String = json[.id] else { return nil }
    ...
}
```
`ORTBBidResponse.swift:74-80`

`BidResponse.init(rawBidResponse:)` guards on that nil and returns early, leaving `rawResponse`, `allBids`, `winningBid`, `tmaxrequest` and `ext` all nil (`BidResponse.swift:56-62`). The deserialization failure is absorbed, not reported.

### Two consequences

**1. `PBMError.responseDeserializationFailed()` has no reachable producer.** `PBMBidResponseTransformer.m` nil-checked the constructed `BidResponse`, but a non-failable Swift initializer bridges to ObjC as `- (nonnull instancetype)initWithJsonDictionary:` and cannot return nil, so that branch was already dead. The factory stays exported and `PBMErrorTest` (lines 32, 58) keeps asserting on an error the SDK cannot raise.

**2. A structurally invalid 200 is delivered as a no-bid.** For a body that is well-formed JSON but not a valid ORTB bid response — `"id"` missing or renamed, e.g. `{"seatbid":[…],"cur":"USD"}` — the transformer returns a hollow `BidResponse` with no error. With the default `requireServerSideBidCache == false`, `PBMBidRequester` takes the success branch and the publisher sees a no-bid instead of a deserialization failure. Nothing in production checks `rawResponse` for nil (`git grep rawResponse` outside `BidResponse.swift` returns only tests).

### Suggested direction

Either restore the signal — have the transformer reject a `BidResponse` whose `rawResponse` is nil and throw `PBMError.responseDeserializationFailed()` — or, if swallowing is intentional, retire the factory together with its assertions. Keeping both is the state we are in now.

Context: #1337 correctly drops the already-dead ObjC branch under the migration's 1:1 rule (playbook Gap S4.1-A); this issue tracks the underlying behavior, which predates it.

mdanylov-sigma:
## Problem

`BidResponse.init(jsonDictionary:)` is non-failable. When the body has no top-level `"id"`, the `ORTBBidResponse` initializer returns nil and `BidResponse` keeps an empty shell: `rawResponse`, `allBids`, `winningBid`, `tmaxrequest` and `ext` are all nil. `BidResponseTransformer` returned that shell as a success, which had two consequences:

1. **`PBMError.responseDeserializationFailed()` had no producer.** Its ObjC nil-check became dead once `BidResponse` got a non-failable Swift initializer, and #1337 dropped it under the 1:1 rule. `PBMErrorTest` kept asserting on an error the SDK could not raise.
2. **A structurally invalid 200 was delivered as a no-bid.** A body that is valid JSON but not an ORTB bid response went down the success branch of `PBMBidRequester`. With the default `requireServerSideBidCache == false`, the publisher saw "no demand" instead of a response-structure error, which is exactly what a misconfigured host answering with some unrelated JSON looks like.

## Change

`BidResponseTransformer.transform(_:)` now throws `PBMError.responseDeserializationFailed()` when the constructed `BidResponse` has a nil `rawResponse`. That error already maps to `ResultCode.prebidInvalidResponseStructure`, the same result code as `jsonDictNotFound()`.

The check lives in the transformer rather than in `BidResponse`: `init(adUnitId:targetingInfo:)` builds from `[:]` on purpose, and `init(jsonDictionary:)` is public, so making it failable would be a breaking change.

## Scope of the behavior change

Only bodies without a top-level string `id` take the new path. Prebid Server always echoes the request id, including on no-bid responses (`bidResponse.ID = bidRequest.ID` in PBS-Go `exchange/exchange.go`, `.id(bidRequest.getId())` in PBS-Java `BidResponseCreator`), so a real PBS response is unaffected.

Android parity note: Android's `BidResponse` reads `id` with `optString` and only sets `hasParseError` on a `JSONException`, so Android still reports such a body as a no-bid. Happy to open a matching Android issue if we want the platforms aligned.

## Tests

- `PBMBidResponseTransformerTest.testMissingResponseId`: a JSON body carrying a bid but no `id` throws `responseDeserializationFailed()`.
- `PBMBidRequesterTest.testBanner_missingResponseId_returnsDeserializationError`: the requester completes with `(nil, responseDeserializationFailed())` instead of a hollow response.
- `PBMErrorTest` is unchanged; its assertions on the factory now cover a reachable error.

## Test plan

- [x] Both new tests **fail** against master's `BidResponseTransformer.swift` (3 assertion failures across the 2 tests, no crash), confirming they guard the fix; both pass with it
- [x] Targeted classes green: `PBMBidResponseTransformerTest` (17), `PBMBidRequesterTest` (16), `PBMErrorTest` (6)
- [x] Full plan green: `PrebidMobileTests` test plan, 1258 tests executed, 0 unexpected failures. `AdViewUtilsTests.testFindPrebidCacheIDSuccess` timed out loading its WKWebView on the first iteration and passed on retry; it does not touch the transformer, and the PR plan skips that class.
- [x] Merges cleanly with the #1344 branch in either order (both edit `transform(_:)`).

Closes #1343


YuriyVelichkoPI (approved):
LGTM

PR #462, opened 2021-12-02
Introduce Rendering API

YuriyVelichkoPI:
Closes #410

PR #1347, opened 2026-09-11
chore: deprecate unused Targeting.contentUrl (#1345)

### LINKED ISSUE #1345: Deprecate or wire up unused `Targeting.contentUrl`
# Description 

`Targeting.shared.contentUrl` is a public iOS-only property that nothing in the SDK reads. It never reaches the bid request and has no Android counterpart, yet the iOS targeting guide on docs.prebid.org still lists it.

## Options

**A. Deprecate and remove**
- Deprecate the property, pointing to `Targeting.shared.setGlobalORTBConfig(_:)` for setting `app.content.url`.
- Remove it from the iOS targeting guide and document the global ORTB config approach for both platforms.

**B. Wire it up**
- Map it to `app.content.url` in the app parameter builder and add a test.
- Add a matching property on Android to keep the APIs aligned.

mdanylov-sigma:
## Problem

`Targeting.shared.contentUrl` is public, but nothing in the SDK reads it. It arrived with the rendering API in #462 as a plain stored property and was never wired to the bid request or to Open Measurement (every OM web view session is created with a `nil` or `""` content URL). Android has no counterpart, yet the iOS targeting guide lists it as a "recommended" property.

## Option A: deprecate now, remove in 4.0

Going with option A from the issue:

- **The replacement already exists on both platforms.** The global ORTB config sets `app.content.url`, and `app` is not a protected field on iOS (`ArbitraryGlobalORTBHelper`) or Android (`OpenRtbMerger`).
- **Wiring it up (option B) has costs.** It would add an Android API for a field publishers can already set. It would also start sending whatever value apps assigned years ago, when the property had no effect.
- **Deprecating now and removing in 4.0** keeps 3.x non-breaking.

## Change

- `Targeting.contentUrl` gets `@available(*, deprecated, message:)` saying it will be removed in 4.0 and pointing at `setGlobalORTBConfig(_:)`, plus a matching `Warning` in its doc comment.
- `ParameterBuilderServiceTest.testGlobalORTBConfigAppContentUrlInParamsDict` checks that `app.content.url` from the global config reaches the built request and merges into the SDK's `app` object instead of replacing it. Nothing previously covered this path end to end through `Targeting`.

Nothing in the repo references `contentUrl`, so the deprecation adds no warnings to the SDK, the demo apps or the tests.

## Docs

[prebid/prebid.github.io#6738](https://github.com/prebid/prebid.github.io/pull/6738) keeps `contentUrl` in the iOS targeting guide, marked deprecated with removal in PrebidMobile 4.0 and a pointer to the global ORTB config. It also adds an `app.content.url` example to the Arbitrary OpenRTB section of both the iOS and Android guides. The example already works on current SDK versions, so the docs change does not need to wait for this release.

## Test plan

- [x] Build succeeds with no `contentUrl` diagnostics in the log
- [x] Targeted classes green: `ParameterBuilderServiceTest`, `TargetingTests`, `ArbitraryORTBServiceTests` (90 tests), including the new `testGlobalORTBConfigAppContentUrlInParamsDict`
- [x] PR subset green: `PrebidMobilePRTests`, 859 tests, 0 failures
- [x] The follow-up commit only changes the deprecation message and doc comment text; `swiftc -parse` on `Targeting.swift` is clean, and CI builds the framework

The new test documents the existing replacement path, so it also passes on master. It guards the claim in the deprecation message rather than a behavior change.

Closes #1345


OlenaPostindustria (approved):
LGTM!

PR #1348, opened 2026-09-11
fix: classify Prebid Server errors only for non-JSON bodies, match any imp index (#1344)

### LINKED ISSUE #1344: Prebid Server error classification matches substrings against the whole response body: false positives on valid bids, and imp[0] hardcoded

mdanylov-sigma:
## Problem

Two pre-existing defects in the Prebid Server error classification in `BidResponseTransformer` (the Swift port in #1337 is a verbatim copy of the ObjC).

### 1. "Invalid request" was matched against the whole body before any structural check

A valid 200 bid response whose creative markup or bidder ext happens to contain the phrase, for example `"adm": "<html><div>Invalid request format detected</div></html>"`, parsed fine as JSON but was classified as a server error. `PBMBidRequester` then completed with `(nil, PBMError.serverError(<entire body>))`: the auction was dropped and the whole response body was copied into the error's `userInfo`.

### 2. The invalid-size check hardcoded `imp[0]`

`Invalid request: Request imp[1].banner.format[0] should define ...` matched neither needle and was reported as `prebidServerError` instead of `prebidInvalidSize`. The first needle, `"Invalid request: Request imp[0].banner.format"`, was also subsumed by the second, so it could never decide the outcome.

## Change

- The "Invalid request" check now runs only when `response.jsonDict` is nil.
- The two `imp[0]` needles are replaced by one regular expression, `imp\[\d+\]\.banner\.format`.

## Why gating on `jsonDict` cannot change a genuine error

- PBS-Go writes `Invalid request: %s` as plaintext and does not set a JSON content type on that path (`writeError` in `endpoints/openrtb2/auction.go`). PBS-Java writes `Invalid request format: <msg>` as plaintext (`AuctionHandler`).
- `PrebidServerConnection` sets `jsonDict` only when the content type is JSON **and** the body parses as a dictionary. If a server labeled a plaintext error as JSON, parsing fails, `serverResponse.error` is set, and `PBMBidRequester` returns before the transformer runs.

So a response that reaches the transformer with a non-nil `jsonDict` was never a plaintext rejection. For comparison, Android's `FetchDemandResult.parseErrorMessage` anchors its patterns at `^Invalid request:` and already uses `imp\[\d\]`.

## Side effect worth a look

PBS-Java prefixes imp paths with a lowercase `request.` (`request.imp[0].banner.format[0] should define ...`, from `ImpValidator`). The old case-sensitive `Request imp[0]` needle never matched it, so PBS-Java size rejections were reported as `prebidServerError`. The regex matches both servers, so those now surface as `prebidInvalidSize`. `testInvalidSizeFromPrebidServerJava` pins that; happy to drop it if it is not wanted.

## Tests

- `testValidBidWithInvalidRequestInMarkup`: a valid bid whose `adm` contains the phrase is returned as a winning bid.
- `testInvalidSizeForNonFirstImp`: `imp[1].banner.format[2]` maps to `prebidInvalidSize`.
- `testInvalidSizeFromPrebidServerJava`: the PBS-Java wording maps to `prebidInvalidSize`.

The existing error fixtures (`invalidAccountIDResponse`, `invalidConfigIdResponse`, `invalidSizeResponse`, `serverErrorResponse`) are all plaintext, so they keep exercising the classifier unchanged.

## Test plan

- [x] All three new tests **fail** against master's `BidResponseTransformer.swift`. The two size tests get `prebidServerError` (code -201) instead of `prebidInvalidSize` (-103), and the markup test catches the thrown server error. All three pass with the fix.
- [x] `PBMBidResponseTransformerTest` green (19 tests, including the 3 new ones)
- [x] PR subset green: `PrebidMobilePRTests`, 861 tests, 0 failures
- [x] `PBMBidRequesterTest`: 14 of 15 green when run as a class; see the note below

**Note on `PBMBidRequesterTest.testBanner_requestInProgress`.** It failed in both class-level runs of `PBMBidRequesterTest` on this branch, and in 1 of 10 isolated iterations. Each time `exp_ok` timed out and the test took 30 to 70 s instead of about 1 s, meaning the main thread stalled before the mocked 1 s callback fired. It passed in both PR-subset runs.

It does not come from this change:
- In an A/B run alternating class-level runs from two separate builds, the build with master's transformer failed the same way on its first run. I confirmed that build contains the old `imp[0]` needle by grepping its binary.
- The test's response is JSON, where this PR's `transform(_:)` does strictly less work, and the transformer only runs after the stall.

Happy to open a separate issue for the flake.

This PR and #1346 (for #1343) both edit `BidResponseTransformer.transform(_:)`; they merge cleanly in either order.

Closes #1344


YuriyVelichkoPI (approved):
In general I'd like to avoid using regexp. The certain strings to search looks more clear and verbose. 

But in this particular case it looks reasonable. Any error related to the `format` object points to the problems with impression size. So such generalization should improve errors handling 

LGTM

PR #1350, opened 2026-09-13
feat: support OpenRTB 2.6 EIDs in user.eids

mdanylov-sigma:
iOS counterpart of prebid/prebid-mobile-android#935 (Android PR: prebid/prebid-mobile-android#954).

## What

- `ExternalUserId` gets the OpenRTB 2.6 provenance fields `inserter`, `matcher` and `mm`. They are optional settable properties, so `init(source:uids:ext:)` and its ObjC selector stay unchanged.
- New `Prebid.shared.eidsPlacement` (`EidsPlacement`: `.openRTB26`, `.openRTB25`, `.compatible`) controls whether EIDs go to `user.eids`, `user.ext.eids` or both. Default: `.compatible`.
- `ORTBUser` gets an `eids` field, so `user.eids` survives decoding and encoding.

## How placement works

The SDK builds its EID list (`Targeting.userExt["eids"]`, `setExternalUserIds`, SharedID) exactly as before. `ParameterBuilderService.placeEids` then runs after the arbitrary ORTB merge: it combines `user.eids` and `user.ext.eids` (entries already in `user.eids` are not repeated) and writes that one list to each location the placement enables.

It runs after the merge because Prebid Server drops `user.ext.eids` whenever `user.eids` is present (PBS-Go `moveEIDFrom25To26`, PBS-Java `BidRequestOrtb25To26Converter`). Writing only the SDK's EIDs to `user.eids` would silently drop EIDs a publisher sends in `user.ext.eids` through `Targeting.userExt` or an ORTB config.

## Review comments on the Android PR and how this PR handles them

| Android #954 comment | This PR |
|---|---|
| Shared `Ext` instance keeps a stale `eids` key after switching placement | `user.ext` is built fresh for every request, and placement works on a copy. `testEidsPlacementChangeLeavesNoStaleEids` switches 2.5, 2.6, 2.5 and checks each request. |
| Test cleanup resets to a non-default placement | `Prebid.reset()` restores `.compatible`, and `ParameterBuilderServiceTest.tearDown` calls it. `resetTargeting` now also clears external user IDs and the global ORTB config, which used to leak between tests. `testEidsPlacementDefaultsToCompatible` pins the default. |
| Default flip needs confirming; non-default placements untested | Default is `.compatible` (see notes). All three placements and the no-EID case are tested end to end through `buildParamsDict`. |
| Both representations built even when one is discarded | One list is built; placement only assigns it. |
| `CollectionUtils.isEmpty` on lists that are never null | Android only. |
| Two serialization paths can drift | No separate `Eid`/`Uid` model classes. `ExternalUserId.toJSONDictionary()` is the only serializer, and both locations get the same array. |

## Notes for reviewers

- **Default.** `.compatible` changes the payload (EIDs appear twice) but not the auction on Prebid Server, which reads `user.eids` and now gets the full list. Servers that only read `user.ext.eids` keep working. `.openRTB26`, which #935 proposed, is cleaner but can silently lose EIDs on servers that don't read `user.eids`. The default should match what the Android PR ships; flipping it is one line plus test updates. Release notes should call it out either way.
- **4.0.** Candidate: default to `.openRTB26` and deprecate `.openRTB25` and `.compatible`.
- **Pre-existing, out of scope.** An `ExternalUserId` with an empty source serializes to `{}` and is still sent, and empty uid ids are not filtered. PBS removes both with a warning; Android skips them. Worth a follow-up issue.
- **Docs.** `prebid-mobile/pbm-api/ios/pbm-targeting-ios.md` says SharedID goes to `user.ext.eids`. It needs an update and an `eidsPlacement` entry.

## Test plan

- [x] `./scripts/testPrebidMobile.sh --latest --quick`: 873 tests. The only failure, `PrebidServerStatusRequesterTests.testRequestStatus_Success` (calls the live status endpoint with a 3s timeout), passed on retry.
- [x] Affected classes: `ExternalUserIdTests`, `UserUniqueIDTests`, `ParameterBuilderServiceTest`, `SharedIdTests`, `PrebidParameterBuilderTest` (77 tests, 0 failures)
- [x] `ORTBAbstractTest`: `user.eids` encode and decode
- [ ] SwiftLint (not installed locally)


YuriyVelichkoPI (dismissed):
LGTM

YuriyVelichkoPI (approved):
LGTM

PR #1351, opened 2026-09-13
Fix #1015: write the log file when the SDK is built without DEBUG

### LINKED ISSUE #1015: Problem with printing Logging and build dynamic Framework from source code
**1. Logging wont write to file when not in debug mode, and should not using print for log**
The logging function wont work when scheme not in debug mode, both from write log to file and print the message to console. 
I understand that should not print log to console when not in debug mode, but instead of using `debugPrint`, in `Log` file the code was like this
```
func print(_ object: Any) {
    // Only allowing in DEBUG mode
    #if DEBUG
    Swift.print(object)
    #endif
}
```
Also the function for check if logging enable `isLoggingEnabled`, always return false when it is not in Debug. 
```
private class func isLoggingEnabled(for currentLevel: LogLevel) -> Bool {
        #if !(DEBUG)
        return false
        #endif
        
        if currentLevel.rawValue < Log.logLevel.rawValue {
            return false
        }
        
        return true
    }
```
It will block write log to file, it should not have check for `DEBUG`, if using `debugPrint` function instead of print. 

**2. The xcframework build command guide from readme file is not really working**
The current command for archive simulator framework only work on simulator on macbook with intel chip, it not work on macbook which run m1 chip. 
```
xcodebuild archive \
	only_active_arch=NO \
	defines_module=YES \
	SKIP_INSTALL=NO \
	-workspace PrebidMobile.xcworkspace \
	-scheme "${schemes[$n]}" \
	-configuration Release \
	-arch x86_64 \
	-sdk "iphonesimulator" \
	-derivedDataPath $XCODE_BUILD_DIR \
	-archivePath "$XCODE_ARCHIVE_DIR/${schemes[$n]}$POSTFIX_SIMULATOR.xcarchive" \
```
Because the arch is `x86_64`, now m1 macbook simulator already run with arm architecture chip. 
It should be update like this 
```
xcodebuild archive \
	only_active_arch=NO \
	defines_module=YES \
	SKIP_INSTALL=NO \
	-workspace PrebidMobile.xcworkspace \
	-scheme "${schemes[$n]}" \
	-configuration Release \
	-destination 'generic/platform=iOS Simulator' \
	-derivedDataPath $XCODE_BUILD_DIR \
	-archivePath "$XCODE_ARCHIVE_DIR/${schemes[$n]}$POSTFIX_SIMULATOR.xcarchive" \
```
So it can run on both Macbook with intel chip and Apple chip simulators.

***May i suggest a PR for this ? ***

mdanylov-sigma:
## Summary

Fixes #1015.

`SDKConsoleLogger.isLoggingEnabled(for:)` returned `false` whenever the SDK was compiled without the `DEBUG` flag:

```swift
#if !(DEBUG)
return false
#endif
```

The default logger dropped every message before it reached `Log.serialWriteToLog`, so `Prebid.shared.debugLogFileEnabled` (`Log.logToFile`) did nothing outside a Debug build. That covers:

- CocoaPods and SPM apps built with a Release configuration (TestFlight or ad hoc QA builds), which is exactly where a log file is most useful.
- The prebuilt XCFrameworks. `buildPrebidMobile.sh` archives with `-configuration Release`, which defines no `SWIFT_ACTIVE_COMPILATION_CONDITIONS`, so publishers integrating the XCFrameworks never got a log file, even from their own Debug scheme.

A custom `PrebidLogger` installed with `Log.setCustomLogger` was not affected.

## Change

The log level is still checked first. After that, a message is formatted and dispatched when it has somewhere to go: the console (DEBUG builds) or the log file (`Log.logToFile`):

```swift
if currentLevel.rawValue < Log.logLevel.rawValue {
    return false
}

return isDebugBuild || Log.logToFile
```

This also removes the `code after 'return' will never be executed` warning that Release builds reported for this function.

What stays the same:

- Console output is still DEBUG only. The module-level `print` wrapper is untouched, which the issue author said is the expected behavior.
- A Release build with `logToFile` off (the default) still returns before building the message, so production apps pay no new `DateFormatter` or dispatch cost per log call.
- No public API change. `isDebugBuild` is an internal stored property initialized from `#if DEBUG`, so tests can exercise the non-DEBUG path. The implicit `init()` is unchanged.

## Part 2 of the issue (simulator archive architecture)

Already resolved on master by #1191, which added `-arch arm64` next to `-arch x86_64` for the simulator archive in `scripts/buildPrebidMobile.sh`. Resolving the `Lib-PrebidMobile` Release simulator build settings with those flags gives `ARCHS = x86_64 arm64` and no `EXCLUDED_ARCHS`, so the XCFramework simulator slice runs on Apple Silicon. No change is needed there.

## Test plan

- [x] `LogTest` (including the new `testLogToFileInReleaseBuild` and `testLogLevelInReleaseBuild`) and `LogObjCTest` pass locally: 17 tests, 0 failures (Xcode 26.5, iOS 26.5 simulator)
- [x] Release build of `Lib-PrebidMobile` for the simulator with `BUILD_LIBRARY_FOR_DISTRIBUTION=YES` succeeds, with no diagnostics in `SDKConsoleLogger.swift`
- [x] End-to-end check against that Release framework: a command-line harness run with `simctl spawn` sets `Log.logToFile = true`, calls `Log.info`, and reads the message back from `Log.getLogFileAsString()`:
  `prebid-mobile-sdk: 2026-09-14 12:18:03592 [ℹ️][main.swift]:7 harness -> issue-1015 release check`
- [x] Negative control: the same harness run against master's SDK, built the same way (the loaded framework was confirmed with `DYLD_PRINT_LIBRARIES`), prints an empty log file, which reproduces the issue
- [ ] CI quick tests

CI builds the tests in Debug, so the new unit tests set `isDebugBuild = false` to reach the non-DEBUG branch. They cannot catch a reintroduced compile-time `#if !(DEBUG)` guard; the end-to-end Release check above covers that case.


YuriyVelichkoPI (approved):
LGTM

PR #1260, opened 2026-04-25
Fix converting of decimal numbers in setGlobalORTBConfig

YuriyVelichkoPI:
Fixes #1259

PR #1352, opened 2026-09-13
Fix #1318: dedupe merged ORTB number arrays by value

### LINKED ISSUE #1318: setOpenRTB method merging NSNumber and NSDecimalNumber
**Describe the bug**

What’s happening:

`setImpORTBConfig:' doesn’t replace the delivery array, it appends to whatever the SDK already put there. Since the SDK 
apparently already emits "delivery": [3] (Download) by default for this ad unit, your override gets concatenated onto it - not merged/deduped as a set. That’s why:
```

•	[1] → [3, 1] (default 3 + your 1)
•	[1, 3] → [3, 1, 3] (default 3 + your 1 and 3, duplicate 3)
```

The `setOpenRTB` method works a bit in different way than expected and duplicates values.

The two arrays being merged come from different JSON-number pipelines:
- The hardcoded default `@(3)` is a plain `NSNumber`.
- The override string is parsed via `Functions.dictionaryPreservingDecimals` (`Utilities/Functions.swift`), which decodes all JSON numbers as `NSDecimalNumber` specifically (to preserve decimal precision for things like bid floors).

`NSDecimalNumber`'s `isEqual:` doesn't reliably match a plain `NSNumber` of the same value. So when `removingDuplicates()` runs its `AnyHashable` equality check, 

`NSNumber(3)` and `NSDecimalNumber(3)` aren't recognized as duplicates, and the second `3` survives.

**To Reproduce**

```

let videoParameters = VideoParameters(mimes: ["video/mp4"])
videoParameters.protocols = [Signals.Protocols.VAST_2_0]
videoParameters.playbackMethod = [Signals.PlaybackMethod.AutoPlaySoundOn]

let adUnit = BannerAdUnit(configId: CONFIG_ID, size: AD_SIZE)
adUnit.adFormats = [.video]
adUnit.videoParameters = videoParameters

// Override delivery (1 = Streaming, 2 = Progressive, 3 = Download)
adUnit.setImpORTBConfig("""
{
  "video": {
    "delivery": [1, 3]
  }
}
""")

```

**Current**
```
"delivery": [3, 1, 3]
```

**Expected**

```
"delivery": [3, 1]
```

mdanylov-sigma:
Closes #1318

## Problem

`setImpORTBConfig`, `setGlobalORTBConfig` and the ad unit global ORTB config are deep-merged into the SDK-built request by `Dictionary.deepMerging(with:)`. Arrays under the same key are concatenated and then deduplicated through `AnyHashable`.

Since #1260 (3.3.1), publisher JSON is parsed with `Functions.dictionaryPreservingDecimals`, which boxes every number as `NSDecimalNumber`. The SDK-built request still holds plain `NSNumber`. `AnyHashable` unwraps the first to `Decimal` and the second to a native integer or floating type, and never treats those as equal, so values present on both sides survive dedup:

```swift
// SDK: imp.video.delivery = @[@(3)]
adUnit.setImpORTBConfig(#"{"video": {"delivery": [1, 3]}}"#)
```

| | `imp[0].video.delivery` |
|---|---|
| Before | `[3, 1, 3]` |
| After | `[3, 1]` |

The same applies to every numeric array the SDK already fills in (`video.protocols`, `api`, ...), not only `delivery`.

## Fix

`deepMerging` now dedupes through `removingDuplicateJSONValues()`, which compares non-boolean numbers by `decimalValue`:

- `NSNumber(3)`, `NSDecimalNumber(3)` and `3.0` compare equal. `Decimal` keeps large integers exact, where a `Double` key would merge distinct values above 2^53.
- Booleans are excluded with the same `CFBooleanGetTypeID` check `JSONParsing.swift` already uses, so `true` stays distinct from `1`.
- The first occurrence wins, so SDK values keep their position and boxing.
- Arrays that aren't hashable (arrays of objects such as `banner.format`) are still concatenated without dedup, as before.

The generic `removingDuplicates()` is unchanged (still used for `battr`).

## Out of scope

Merging still appends rather than replaces, so an ORTB config can't remove the SDK default `delivery: [3]`. Making `delivery` configurable is tracked in #1314.

## Test plan

- [x] `CollectionExtensionTest.testDeepMergeArraysWithDuplicates_numbersWithDifferentBoxing`
- [x] `CollectionExtensionTest.testDeepMergeArraysWithDuplicates_booleansAreNotNumbers`
- [x] `ArbitraryORTBServiceTests.testMergeImpORTB_deduplicatesNumbersAlreadySetBySDK_issue1318` (issue repro)
- [x] All three new tests fail without the fix (`delivery` comes out `[3,1,3]`, as reported)
- [x] Related classes pass: `CollectionExtensionTest` (16), `ArbitraryORTBServiceTests` (41), `ArbitraryORTBHelperTests` (6), `FunctionsTests` (14)
- [x] `./scripts/testPrebidMobile.sh --latest --quick` (861 tests, 0 failures)


YuriyVelichkoPI (approved):
LGTM

Issue #1338, opened 2026-09-08
[Performance] High SDK request preparation latency (~450ms) reduces bidder response window


jaysongkkim:
## Description
During timeout testing on the iOS SDK, we noticed significant latency before the network request is actually dispatched. Internal SDK request preparation alone takes approximately **450ms**.

## Impact
This internal overhead severely shrinks the remaining time window for demand partners/bidders to process and respond within the configured timeout limit. As a result, bid participation rates and win rates are negatively affected.

## Steps to Reproduce / Observations
1. Configure timeout settings and trigger an ad request on iOS SDK.
2. Trace the execution timeline from `fetchDemand` (or ad request call) up to the actual HTTP request dispatch.
3. Observe ~450ms delay consumed internally by request preparation (e.g., serialization, configuration assembly, targeting buildup).

## Suggested Action
Investigate and optimize the internal processing pipeline during request preparation to minimize delay before network execution.

mdanylov-sigma:
Dug into this. The ~450 ms is real, but it is **not** spread across serialization/targeting buildup — steady-state request assembly is only a few milliseconds. The latency is concentrated in three specific places.

I instrumented `ParameterBuilderService.buildParamsDict` (i.e. exactly what `PBMBidRequester.getRTBRequest` calls), 20 iterations, iPhone 16 Pro simulator:

```
FULL buildParamsDict (getRTBRequest)       median=   6.509 ms
  CTTelephonyNetworkInfo() alloc           median=   2.237 ms
  DeviceInfoParameterBuilder.build         median=   3.313 ms
  NetworkParameterBuilder.build            median=   0.039 ms
  BasicParameterBuilder.build              median=   0.004 ms
  UserConsentParameterBuilder.build        median=   0.004 ms
  AppInfoParameterBuilder.build            median=   0.017 ms
  SKAdNetworksParameterBuilder.build       median=   0.015 ms
  GeoLocationParameterBuilder.build        median=   0.003 ms
  ORTBBidRequest.jsonDictionary            median=   0.019 ms
  ORTBParameterBuilder.buildOpenRTB (JSON) median=   0.022 ms
  ArbitraryORTBService.merge               median=   0.000 ms
```

So JSON serialization and targeting assembly are ~0.1 ms combined and are not the problem. Everything expensive is a **synchronous call into a system daemon**, repeated on every request.

## 1. Cold user-agent fetch blocks the request (dominant)

[...]

PR #1354, opened 2026-09-18
fix: reuse a single URLSession so connections are pooled (#1339)

### LINKED ISSUE #1339: [Performance] Lack of HTTP connection / request reuse adds ~270ms overhead per request
## Description
While profiling network performance during ad requests, we observed an additional delay of approximately **270ms** per request. This appears to be caused by requests not reusing existing HTTP connections or `URLSession` configurations, leading to repeated connection setups/handshakes.

## Impact
An unnecessary ~270ms latency penalty is added to every request cycle, compounding with the request preparation time and further eating into the bidder timeout limit.

## Observations
- Each request incurs standard connection setup costs that could be avoided through proper connection pooling or persistent `URLSession` management.
- Re-creating sessions or connection objects per request adds noticeable overhead (~270ms).

## Suggested Action
- Review `URLSession` instance management to ensure persistent connections / HTTP keep-alive are properly utilized.
- Reuse network session configurations across consecutive ad requests wherever possible.

mdanylov-sigma:
Fixes #1339.

## Problem

`PrebidServerConnection.createSession(_:)` built a brand-new `URLSession` for every call — `post`, `get`, `head`, `download` and `fireAndForget` each did `let session = createSession(timeout)` and then `task.resume()`.

`URLSession` keeps its connection pool *per session*, so a new session can never reuse a connection opened by a previous one. Every request paid a full DNS + TCP + TLS handshake.

To be clear about what was *not* the cause: `URLSessionConfiguration.ephemeral` is fine, and pooling works normally inside an ephemeral session. Session *creation* is also cheap (~0.01 ms). The cost was purely the lost pool.

This was not limited to the auction. Each of these opened its own session and its own connection: the auction `POST` (`PBMBidRequester`), `burl`/impression pings (`PrebidServerEventTracker`, `AdModelEventTracker`), win notifications (`WinNotifierImpl`), click tracking (`TrackingURLVisitors`) and Prebid JS downloads (`PrebidJSLibraryManager`). Several target the same host as the auction, so they now ride its existing h2 connection.

Separately, no session was ever invalidated anywhere in the SDK, so the `#if DEBUG` branch — which passes `delegate: self` plus a fresh `OperationQueue` — leaked a session and a queue per request.

## Measurement

A/B against `https://prebid-server-test-j.prebid.org/openrtb2/auction` using `URLSessionTaskMetrics`, with a configuration identical to `createSession`, 5 sequential POSTs each:

```
=== A) NEW ephemeral URLSession per request  (before) ===
  reused=no  proto=h2 dns=  55.0 tcp= 548.0 tls= 397.0 ttfb= 166.3 TOTAL=  822.4 ms
  reused=no  proto=h2 dns=   1.0 tcp= 440.0 tls= 294.0 ttfb= 153.0 TOTAL=  599.1 ms
  reused=no  proto=h2 dns=   1.0 tcp= 572.0 tls= 318.0 ttfb= 166.3 TOTAL=  743.2 ms
  reused=no  proto=h2 dns=   1.0 tcp= 448.0 tls= 308.0 ttfb= 144.0 TOTAL=  598.1 ms
  reused=no  proto=h2 dns=   1.0 tcp= 397.0 tls= 268.0 ttfb= 137.0 TOTAL=  541.9 ms

=== B) ONE long-lived URLSession reused       (after) ===
  reused=no  proto=h2 dns=   1.0 tcp= 476.0 tls= 322.0 ttfb= 154.1 TOTAL=  638.6 ms
  reused=YES proto=h2 dns=  -1.0 tcp=  -1.0 tls=  -1.0 ttfb= 189.9 TOTAL=  191.5 ms
  reused=YES proto=h2 dns=  -1.0 tcp=  -1.0 tls=  -1.0 ttfb= 154.2 TOTAL=  156.0 ms
  reused=YES proto=h2 dns=  -1.0 tcp=  -1.0 tls=  -1.0 ttfb= 160.7 TOTAL=  162.0 ms
  reused=YES proto=h2 dns=  -1.0 tcp=  -1.0 tls=  -1.0 ttfb= 261.2 TOTAL=  265.9 ms
```

`-1.0` means the phase did not happen at all. After the first request the handshake disappears and the request collapses to round-trip time: **~600 ms → ~160 ms**. The absolute saving scales with RTT to the PBS host, consistent with the ~270 ms reported in the issue.

## Change

Cache one `URLSession` per `PrebidServerConnection` and move the timeout onto the request.

Three details that keep this behaviour-preserving:

1. **Per-request timeout still wins.** Verified against a non-responding local server: `config=60s` + `request.timeoutInterval=2s` fires `NSURLErrorTimedOut` at 2.02 s, matching the previous `config=2s` at 2.01 s. The auction timeout (`Prebid.shared.timeoutMillis` / `timeoutMillisDynamic`) is unchanged.
2. **`fireAndForget` and `download` now set the timeout explicitly.** They previously relied on the configuration alone (`FIRE_AND_FORGET_TIMEOUT`, 3 s) and would otherwise have silently inherited the session-wide value.
3. **`get(_:timeout:)` defaults to `0`** and `TrackingURLVisitors` passes `0`. `setTimeoutInterval(_:)` applies the value only when it is greater than zero, leaving the `URLRequest` default in place rather than depending on how URLSession reads a literal zero.

`protocolClasses` gained a `didSet` that rebuilds the session, because `URLSessionConfiguration` is copied when the session is created. Without it the 7 test call sites that append `MockServerURLProtocol` *after* constructing the connection would have no effect.

## Connection concurrency, and why the per-host cap moves to 12

Sharing one session also means sharing its connection pool, which `URLSession` caps at `httpMaximumConnectionsPerHost` — **6** by default. Before this PR every request had its own session and therefore its own pool, so per-host parallelism was effectively unbounded. Inheriting 6 silently would be a real behavioural change, so the third commit sets it explicitly.

### Why it matters

Requests past the cap queue. That queuing is not free, but *which* clock it spends is subtle:

- **`URLRequest.timeoutInterval` does not start until a request gets a connection.** Verified: 12 parallel requests against an HTTP/1.1 endpoint, each with a 2000 ms request timeout, all completed successfully even though the queued half took 3034 ms. So queuing cannot cause a spurious network timeout.
- **`AdUnit.baseFetchDemand` is different.** It arms its demand timeout in wall-clock time from the `fetchDemand` call, before the request is even built:

  ```swift
  DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(timeout), execute: timeoutHandler)
  ```

  A request still waiting for a connection is already spending that budget.

### Measurement

HTTP/1.1 endpoint, 900 ms latency, default 2000 ms demand timeout:

```
adUnits  cap   onTime  demandTimedOut   slowest
     8    6        8            0       1824 ms
     8   12        8            0       1018 ms

    12    6       12            0       1837 ms
    12   12       12            0        970 ms

    16    6       12            4       2746 ms   <-- 4 bids lost
    16   12       16            0       1841 ms
```

Two waves already burns 92% of the demand budget. Three waves loses bids outright.

### Why 12

- Keeps a realistic number of simultaneous ad units in a single wave — even feed-heavy apps rarely run more than 12 concurrent auctions.
- Double the platform default, so meaningfully safer than 6 without swapping one extreme (unbounded) for another.
- **Costs essentially nothing in practice.** HTTP/2 multiplexes, so the cap never engages there — 12 parallel requests to the real PBS host used exactly **1** TCP connection. This only spends anything against an HTTP/1.1 origin, where it is buying back correctness.
- Higher values give diminishing returns while adding real handshakes and radio time on mobile.

It is a single named constant with the reasoning in a comment, and a test pins the value, so changing it later has to be deliberate. Happy to go 8 or 16 if you read the trade-off differently.

## Test plan

- [x] `ServerConnectionTest` — 14/14 pass (10 existing + 4 new).
- [x] New regression tests: session identity is stable across requests; mutating `protocolClasses` rebuilds the session; `fireAndForget` and `download` carry their own timeout; the per-host cap is 12.
- [x] Mutation-tested the new tests rather than trusting a green run — defeating the cache and dropping the timeout assignments makes both of the relevant new tests fail.
- [x] Concurrency checks: invalidating mid-flight completes all in-flight requests; create → invalidate → resume still fires its callback; 200 concurrent calls to the session getter create exactly one session.
- [x] Full `PrebidMobileTests` plan: 1353 tests. The one remaining failure is `AdViewUtilsTests.testFindPrebidCacheIDSuccess`, which also fails on pristine `master` (confirmed by building the unmodified tree into a separate `derivedDataPath`) and is unrelated to this change.
- [x] CI green.

## Notes for the reviewer

- **The second commit fixes a test that is broken on `master`, not by this PR.** `BannerViewTest.testBannerViewReportsExpirationAndKeepsDeployedViewWhenRefreshIsDisabled` is a semantic merge conflict: #1341 removed the branch forcing `refreshInterval` to 0 for outstream video, and #1267 — merged after but branched before — added a test relying on it. `master` CI has not run since either landed, so this PR was the first run to surface it, and it would fail the same way on every open PR. See [the comment below](https://github.com/prebid/prebid-mobile-ios/pull/1354#issuecomment-5735270247) for the detail. **Happy to split that commit into its own PR** if you'd rather keep this one single-purpose.
- `session` is `internal` rather than `private` so the tests can assert the identity is stable. Happy to swap it for a dedicated test hook if you'd prefer to keep it private.
- With one long-lived session the DEBUG delegate retain drops from O(requests) to O(connections), but it isn't fully gone — `delegate: self` still means the session retains the connection until invalidated. Fully breaking it would mean moving `checkCertificateChallenge` onto a separate lightweight delegate object. I left that out of scope; happy to file a follow-up.
- SwiftLint isn't installed on my machine so I could not run it. Added lines are well under the 250 column limit, but note this file's class body was likely already over `type_body_length: 100` before this change.
- #1338 (the ~450 ms of request preparation) is a separate set of causes — cold `WKWebView` user-agent fetch blocking the auction, per-request `CTTelephonyNetworkInfo` allocation, and prep running on the caller's thread. Not addressed here.


mdanylov-sigma:
Heads up: the first CI run failed on a test that is **broken on `master` already**, unrelated to this PR. I've pushed a second commit fixing it, but it deserves its own explanation since it isn't really my change to make.

### `BannerViewTest.testBannerViewReportsExpirationAndKeepsDeployedViewWhenRefreshIsDisabled`

This is a semantic merge conflict between two PRs that landed the same day:

- **#1341** (`7e561aa4`) removed the branch in `AdUnitConfig.refreshInterval` that forced the interval to 0 for outstream video, and documented the property as "Pure configuration: independent of `adFormats` and of the winning bid's format":

  ```diff
  -            if adConfiguration.winningBidAdFormat == .video {
  -                Log.warn("'refreshInterval' property is not assignable for Outstream Video ads")
  -                _refreshInterval = 0
  ```

- **#1267** (`4f064373`), merged after it but branched before it, added this test relying on exactly that removed behaviour:

  ```swift
  // Outstream video is not refreshable: the interval is forced to 0.
  bannerView.adUnitConfig.adConfiguration.winningBidAdFormat = .video
  bannerView.refreshInterval = 0
  XCTAssertEqual(bannerView.refreshInterval, 0)   // 15.0 is not equal to 0.0
  ```

Each was green on its own base; the combination is not. `master` CI has not run since either landed, so this PR is the first run to surface it — and it will fail the same way on **every** open PR until it's fixed.

Assigning `0` now clamps up to `refreshIntervalMin` (15); only a negative value yields 0. Since `reportAdExpired` gates on `refreshInterval > 0`, a negative assignment is what actually expresses "refresh is disabled", so that's what the second commit does, and it drops the now-irrelevant `winningBidAdFormat` line. The test's intent is unchanged.

I'm happy to pull that commit out into its own PR against `master` if you'd rather keep this one to the `URLSession` change — say the word and I'll split it.

### `DownloadSizeHelperTest.testSize`

Also failed on the first run, then passed on retry. That is the known flake, not related to this change.

mdanylov-sigma:
Added a third commit setting `httpMaximumConnectionsPerHost` to 12, and expanded the PR description with the reasoning. Summarising here since it's a deliberate value choice worth disagreeing with.

## Why the change was needed

Sharing one session also means sharing its connection pool, which `URLSession` caps at **6 per host** by default. Before this PR each request had its own session and therefore its own pool, so per-host parallelism was effectively **unbounded**. Keeping one session without touching this would silently reduce that to 6 — a real behavioural change that should be made on purpose rather than inherited.

Requests past the cap queue, and the subtle part is *which* clock that queuing spends:

- **`URLRequest.timeoutInterval` does not start until a request gets a connection.** I verified this: 12 parallel requests against an HTTP/1.1 endpoint, each with a 2000 ms request timeout, all succeeded even though the queued half took 3034 ms. So queuing on its own cannot cause a spurious network timeout.
- **`AdUnit.baseFetchDemand` is different.** It arms the demand timeout in wall-clock time from the `fetchDemand` call, before the request is even built:

  ```swift
  DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(timeout), execute: timeoutHandler)
  ```

  So a request still sitting in the connection queue is already spending its demand budget. That is the actual risk, and it's what decided the value.

## The data

HTTP/1.1 endpoint, 900 ms latency, default 2000 ms demand timeout:

```
adUnits  cap   onTime  demandTimedOut   slowest
     8    6        8            0       1824 ms
     8   12        8            0       1018 ms

    12    6       12            0       1837 ms
    12   12       12            0        970 ms

    16    6       12            4       2746 ms   <-- 4 bids lost
    16   12       16            0       1841 ms
```

Two waves already burns 92% of the demand budget; three waves loses bids outright.

## Why 12

- Keeps a realistic number of simultaneous ad units in a single wave — even feed-heavy apps rarely run more than 12 concurrent auctions.
- Double the platform default: meaningfully safer than 6, without swapping one extreme (unbounded) for another.
- **Near-zero cost in practice.** HTTP/2 multiplexes, so the cap never engages there — 12 parallel requests to the real PBS host used exactly **1** TCP connection. It only spends anything against an HTTP/1.1 origin, where it's buying back correctness.
- Higher values give diminishing returns while adding real handshakes and radio time on mobile.

It's a single named constant with the rationale in a comment, and a test pins the value so a future change has to be deliberate. If you read the trade-off differently — 8 for conservatism, or 16 for denser apps — it's a one-line change and I have no strong attachment to 12 over its neighbours.

## Also verified while I was in here

- Invalidating the session mid-flight completes all in-flight requests (5/5, no errors).
- `create task → invalidate → resume` still fires its callback and completes — so the `protocolClasses` `didSet` can't drop a request.
- 200 concurrent calls to the session getter create exactly one session.

CI is green on all four checks.

